PR #730, opened 2015-04-17
Httpd optinal cgi client serializing

AntonKozlov:
This changeset allows to control, weither httpd allows it's cgi childs run in parallel.
When on, child started and then gathered when it's done, sometime. This mode intentend to long standing connections, e.g. status
When off, httpd waits for child as soon as it started. This mode is used when resources are limited and more preceice proccess control required


abusalimov:
@AntonKozlov do you have time to fix the comments now? Otherwise, I could fix it myself.

AntonKozlov:
Some comments are outdated, so I didn't respond to them

abusalimov:
OK, so what's about rebasing the commits to add `httpd: ` prefix (to most them)?

AntonKozlov:
Is it really necessary?
20 апр. 2015 г. 22:31 пользователь "Eldar Abusalimov" <

abusalimov:
I believe, it's worth doing. Let's try to keep the history consistent.

PR #733, opened 2015-04-20
Fix obscure initialization bug

AntonKozlov:
There was NoRuntime dependency on thread_local, which could use
sysmalloc. There is no need to say, that allocator was not initialized.


abusalimov:
:shipit:

PR #739, opened 2015-04-27
webui: Reduce too long led's name (maxlen is 9)

AntonKozlov:
- Use monospace font-family as default for .block-caption
 - Add led.short_name field


anton-bondarev:
LGTM
But may be use both led number and led name at the same time.

Issue #721, opened 2015-04-14
New README and contributing guides


abusalimov:
The current README file is completely outdated.
- [x] Provide more actual project description
- [ ] Add [contributing guide-lines](https://help.github.com/articles/setting-guidelines-for-repository-contributors/) (CONTRIBUTING.md file)
- [x] Quick start

sikmir:
[Quick start] What about advising to use qemu-linaro for ARM emulation?

AntonKozlov:
qemu-linaro have beagleboard emulation, which is our only available arm/qemu platform

sikmir:
@AntonKozlov, I just suggest to mention about it in Quick Start guide.

abusalimov:
@sikmir Could you suggest a PR for that, please? The current setup works for me just fine, likely because I have an environment similar to that on Travis, so that I'm afraid I won't be able to reproduce the issue and properly fix it.

sikmir:
@abusalimov It's not an issue in fact, just a note for archlinux users.
Case 1:
```fish
~ sudo pacman -S qemu
~ qemu-system-arm --machine help | grep overo
(nothing)
```
Case 2:
```fish
~ yaourt qemu-linaro
~ qemu-system-arm --machine help | grep overo
overo                Gumstix Overo board (OMAP3530)
```
I don't know why follow one works for travis:
```fish
sudo apt-get install -qq qemu qemu-system
```

AntonKozlov:
travis seems to use some ubuntu-based distro (may be ubuntu itself). It's qemu built with linaro patches.

https://launchpad.net/qemu-linaro

`Want native binaries? The qemu-system package in Ubuntu Maverick and later already includes some Linaro changes ...`

Deryugin:
Debian has the same problem with qemu from default repos.
Also, build fails if */bin/sh* points to */bin/dash* (*"Bad substitution"*), and it's default for some debian and ubuntu versions.
Issue could be fixed by linking */bin/sh* to */bin/bash*.

PR #741, opened 2015-04-29
README: Add Getting Started guide

abusalimov:
Mostly moved from the legacy Wiki: http://code.google.com/p/embox/w/edit/GettingStarted

Cc: @anton-bondarev


abusalimov:
More discussion here: #721

PR #744, opened 2015-04-30
VFS

Deryugin:


src/compat/posix/fs/dvfs/dirent.c
```diff
@@ -0,0 +1,101 @@
...
+ * @date 5 Apr 2015
+ * @author Denis Deryugin
+ */
+
+#include <errno.h>
+#include <posix_errno.h>
+#include <string.h>
+
+#include <dirent.h>
+#include <defines/null.h>
```
anton-bondarev: use standard header where is posible. NULL is defined in stddef.h

src/compat/posix/fs/dvfs/dirent.c
```diff
@@ -0,0 +1,101 @@
...
+#include <dirent_dvfs_impl.h>
+#include <framework/mod/options.h>
+#include <fs/dvfs.h>
+#include <mem/objalloc.h>
```
anton-bondarev: Are you sure that you can use objalloc may be anought pool?

src/compat/posix/fs/dvfs/fcntl.c
```diff
@@ -0,0 +1,26 @@
...
+ * @date 3 Apr 2015
+ * @author Denis Deryugin
+ */
+
+#include <fcntl.h>
+#include <kernel/task/resource/file_table.h>
+
+int fcntl(int fd, int cmd, ...) {
+	va_list args;
+	union {
```
anton-bondarev: why do you use union here?

src/compat/posix/fs/dvfs/dirent.c
```diff
@@ -0,0 +1,101 @@
...
+	struct dentry *prev;
+
+	if (!dir) {
+		SET_ERRNO(EBADF);
+		return NULL;
+	}
+
+	prev = dir->prv_dentry;
+
+	l = (struct lookup) {
```
anton-bondarev: May be it's better to use 
l.parrent = xxx;
l.item = xxx;

src/compat/posix/fs/dvfs/fsop.c
```diff
@@ -0,0 +1,52 @@
...
+
+int creat(const char *pathname, mode_t mode) {
+	int rc;
+	rc = open(pathname, O_CREAT | O_WRONLY | O_TRUNC | mode);
+	DPRINTF(("creat(%s, %d ...) = %d\n", pathname, mode, rc));
+	return rc;
+}
+
+int mkdir(const char *pathname, mode_t mode) {
+	return creat(pathname, O_DIRECTORY);
```
anton-bondarev: Did you forgot mode parameter?
O_DIRECTORY | mode?

src/compat/posix/fs/dvfs/fcntl.c
Deryugin: It was originally copy-pasted from old fcntl() and then adopted to new VFS.
I agree, it's bad. Will be fixed

src/compat/posix/fs/dvfs/open.c
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @author  Denis Deryugin
+ * @date    29 Mar 2015
+ */
+
+#include <fs/dvfs.h>
+#include <kernel/task.h>
+#include <kernel/task/resource/file_table.h>
+
+int open(const char *path, int __oflag, ...) {
```
anton-bondarev: you should include header where routine is defined.

src/compat/posix/fs/dvfs/dirent.c
Deryugin: I prefer this style because this way you shouldn't write variable name twice.
In this particular case "l" is a short name, but if it would be something like "data->lookup", or if there are many fields to be filled, code would be shorter.
Also, this way all omitted fields will be automatically set to zero, so it's more safe in some way.

src/drivers/block_dev/block_dev.c
```diff
@@ -29,18 +25,10 @@
 
 #include <embox/block_dev.h>
 
-#define MAX_DEV_QUANTITY OPTION_GET(NUMBER,dev_quantity)
+#define MAX_DEV_QUANTITY 16 /* XXX */
```
anton-bondarev: why did you switch option_get to constant?

src/compat/posix/fs/dvfs/fsop.c
Deryugin: You are right.

src/drivers/block_dev/block_dev.c
Deryugin: "max_dev_quantity" should be declared in "block_common" Mybuild module, as pools are defined there.
At the same, "max_dev_quantity" are required in "block" module, which depends on "block_common" and which is VFS-dependent.
I have no idea how to use MAX_DEV_QUANTITY with "option_get" in this situation.

src/compat/posix/fs/dvfs/dirent.c
Deryugin: You should be right. I will replace it.

src/compat/posix/fs/dvfs/dirent.c
Deryugin: Will fix it.

src/drivers/block_dev/block_dev.c
Deryugin: Oh, sorry. I see that this constant is not even used in block_dev.c, so it could be removed and replaced with option_get. Will fix it.

src/compat/posix/fs/dvfs/dirent.c
anton-bondarev: ok

src/compat/posix/fs/dvfs/getcwd.c
```diff
@@ -0,0 +1,39 @@
...
+		dir = "/";
+	}
+	assert(dir != NULL); /* $PWD must be set */
+
+	if (strlen(dir) + 1 >= size) {
+		SET_ERRNO(ERANGE);
+		return NULL;
+	}
+
+	strcpy(buff, dir);
```
anton-bondarev: strncpy(buff, dir, size)?

src/compat/posix/fs/dvfs/libgen.c
```diff
@@ -0,0 +1,79 @@
...
+
+	if (path_len > 0) {
+		return c;
+	}
+
+	return NULL;
+}
+
+char * basename(char *path) {
+	static char bname_buff[2];
```
anton-bondarev: is this variable used  only for return "."? Why don't use it for "/" or better remove one and use return constant string 
return "."

src/compat/posix/fs/dvfs/open.c
```diff
@@ -0,0 +1,36 @@
...
+	res = dvfs_open(path, file, __oflag);
+
+	ft = task_fs();
+	for (res = 0; res < FILE_TABLE_SZ; res++)
+		if (ft->file[res] == NULL) {
+			ft->file[res] = file;
+			break;
+		}
+
+	if (res == FILE_TABLE_SZ || (file->f_inode == NULL && file->f_dentry == NULL)) {
```
anton-bondarev: it's too big condition. If I understand it's posible to check file field before the loop?

src/compat/posix/fs/dvfs/read.c
```diff
@@ -0,0 +1,24 @@
...
+
+ssize_t read(int fd, void *buf, size_t nbyte) {
+	struct file *file;
+	struct file_table *ft;
+
+	ft = task_resource_file_table(task_self());
+	if (fd < FILE_TABLE_SZ && ft->file[fd] != NULL) {
+		file = ft->file[fd];
+		return dvfs_read(file, buf, nbyte);
+	} else {
```
anton-bondarev: may be 'else' redundantly?

src/compat/posix/fs/dvfs/xattr.c
```diff
@@ -0,0 +1,35 @@
+/**
+ * @file
+ * @brief
+ *
+ * @author  Denis Deryugin
+ * @date    3 Apr 2015
+ */
+
+#include <defines/size_t.h>
```
anton-bondarev: stddef.h?

src/fs/driver/dfs/dfs.c
```diff
@@ -49,10 +50,19 @@ static inline int _erase(unsigned int block) {
...
 
+/* @brief Write non-aligned raw data to \b erased NAND flash
+ * @param offset Start position on disk
+ * @param buff   Source of the data
+ * @param len    Length of the data in bytes
+ *
+ * @returns Bytes written or negative error code
+ */
 static inline int _write(unsigned long offset, const void *buff, size_t len) {
+	assert(buff);
```
anton-bondarev: you should use asserts after 'variables definition' block

src/fs/driver/dfs/dfs.c
```diff
@@ -224,43 +258,46 @@ struct super_block_operations dfs_sbops = {
...
-		return NULL;
+	if (sbi->inode_count > sbi->max_inode_count)
+		return -ENOMEM;
 
-	dirent.pos_start = sbi.free_space;
+	dirent.pos_start = sbi->free_space;
 	dirent.len = 136; /* XXX */
-	strcpy(dirent.name, d_new->name);
-	dfs_write_dirent(sbi.inode_count, &dirent);
+	strcpy(dirent.name, i_new->i_dentry->name);
```
anton-bondarev: use strncpy everywhere as posible

src/fs/driver/dfs/dfs.c
```diff
@@ -293,42 +333,66 @@ static struct inode *dfs_ilookup(char const *path, struct dentry const *dir) {
...
+static int dfs_iterate(struct inode *next, struct inode *parent, struct dir_ctx *ctx) {
+	struct dfs_dir_entry dirent;
+	assert(ctx);
+
+	if (ctx->pos >= ((struct dfs_sb_info *) dfs_sb()->sb_data)->inode_count)
+		return -1;
+
+	dfs_read_dirent(ctx->pos, &dirent);
+
+	*next = (struct inode) {
```
anton-bondarev: where will this structure be allocated?

src/fs/driver/dfs/dfs.c
```diff
@@ -360,36 +427,41 @@ struct file_operations dfs_fops = {
...
 
 	return 0;
+}
+
+static struct dumb_fs_driver dfs_dumb_driver = {
+	.name = "DumbFS",
+	.fill_sb = &dfs_fill_sb,
 };
 
+ARRAY_SPREAD_DECLARE(struct dumb_fs_driver *, dumb_drv_tab);
```
anton-bondarev: declare and add in single file?

src/fs/driver/fat/fat.c
```diff
@@ -1,6 +1,6 @@
 /**
- * @file
- * @brief
+ * @file  fat.c
```
anton-bondarev: don't use file name for doxigen

src/fs/driver/fat/fat.c
```diff
@@ -13,65 +13,24 @@
...
-#include <embox/block_dev.h>
-
+#include <fs/file_system.h>
+#include <fs/fs_driver.h>
+#include <fs/node.h>
+#include <fs/vfs.h>
 #include <framework/mod/options.h>
 
 #define FAT_MAX_SECTOR_SIZE OPTION_GET(NUMBER, fat_max_sector_size)
+uint8_t fat_sector_buff[FAT_MAX_SECTOR_SIZE];
```
anton-bondarev: This is a global variable? It's better use getter for receiving one.

src/fs/driver/fat/fat_common.c
```diff
@@ -0,0 +1,1608 @@
...
+#include <string.h>
+
+#include <embox/block_dev.h>
+#include <framework/mod/options.h>
+#include <fs/fat.h>
+#include <mem/misc/pool.h>
+#include <util/math.h>
+
+#define LABEL "EMBOX_DISK"
+#define SYSTEM "FAT12"
```
anton-bondarev: only FAT12?

src/fs/driver/fat/fat_common.c
```diff
@@ -0,0 +1,1608 @@
...
+
+extern uint8_t fat_sector_buff[512]; /* XXX */
```
anton-bondarev: Early I commented this variable, except that comment here is a constant 512 although first was 
uint8_t fat_sector_buff[FAT_MAX_SECTOR_SIZE];

src/fs/driver/fat/fat_common.c
```diff
@@ -0,0 +1,1608 @@
...
+
+uint32_t fat_get_next(struct fat_fs_info *fsi,
+		struct dirinfo * dirinfo, struct dirent * dirent);
+
+char bootcode[130] =
```
anton-bondarev: also global?

src/fs/driver/fat/fat_common.c
```diff
@@ -0,0 +1,1608 @@
...
+
+	if (0 > block_dev_write(fsi->bdev, (char *) buffer, sec_size, sector * sec_size / dev_blk_size)) {
+		return DFS_ERRMISC;
+	} else {
+		return DFS_OK;
+	}
+}
+
+int fat_create_partition(void *bdev) {
+	uint16_t bytepersec = 512;
```
anton-bondarev: one more constant 512?

src/fs/driver/fat/fat_common.c
```diff
@@ -0,0 +1,1608 @@
...
+			}
+			fat_set_fat_(fsi, di->p_scratch, &i, di->currentcluster, tempclus);
+		}
+	} while (!tempclus);
+
+	/* We shouldn't get here */
+	return DFS_ERRMISC;
+}
+
+#define MSDOS_DOT     ".          "
```
anton-bondarev: extract to header?

src/fs/driver/dfs/dfs.c
Deryugin: It's allocated in VFS iterate handler (dvfs_iterate)

src/fs/driver/dfs/dfs.c
Deryugin: It's used in FAT driver also.

src/fs/driver/fat/fat_common.c
Deryugin: Yes, current FAT driver is only able to format drive to FAT12.
However, it can mount and use FAT12/16/32 if it was formatted before.
Implementing FAT16/32 formatting is added to Trello, but it's not very important right now.

src/fs/driver/fat/fat_common.c
Deryugin: The problem is that sector size should be passed as option to the abstract FAT class, but I got problems getting this option in child modules, so I left it constant.

PR #745, opened 2015-04-30
BananaPi port: working SGI

dimart:
As discussed, this change provides basic support for SGI on BananaPi, successfully passing all IRQ tests.

### DIFF
--- a/src/drivers/interrupt/bpi_intc.c
+++ b/src/drivers/interrupt/bpi_intc.c
@@ -1,0 +1,153 @@
+/**
+ * @file
+ * @brief Banana Pi interrupt controller driver
+ *
+ * @date 15.04.15
+ * @author Dmitrii Petukhov
+ */
+
+#include <assert.h>
+
+#include <kernel/critical.h>
+#include <hal/reg.h>
+#include <hal/ipl.h>
+#include <drivers/irqctrl.h>
+
+#include <kernel/irq.h>
+#include <embox/unit.h>
+#include <kernel/printk.h>
+
+EMBOX_UNIT_INIT(bpi_intc_init);
+
+/** On BananaPi SGI source numbers lie between 0 and 15 */
+#define SGI_MAX_SRC         15
+
+#define GIC_BASE            0x01C80000
+
+/** 0x1000-0x1FFF Distributor */
+#define GICD_BASE          (GIC_BASE + 0x1000)
+
+/** 0x2000-0x3FFF CPU interfaces */
+#define GICC_BASE          (GIC_BASE + 0x2000)
+
+#define GICD_CTRL          (GICD_BASE + 0x000)
+#define GICD_ISENABLER(n)  (GICD_BASE + 0x100 + (n) * 0x4)
+#define GICD_ICENABLER(n)  (GICD_BASE + 0x180 + (n) * 0x4)
+#define GICD_ISACTIVER(n)  (GICD_BASE + 0x300 + (n) * 0x4)
+#define GICD_ICACTIVER(n)  (GICD_BASE + 0x380 + (n) * 0x4)
+#define GICD_IPRIORITYR(n) (GICD_BASE + 0x400 + (n) * 0x4)
+#define GICD_SGIR          (GICD_BASE + 0xF00)
+#define GICD_CPENDSGIR(n)  (GICD_BASE + 0xF10 + (n) * 0x4)
+
+#define GICC_CTRL          (GICC_BASE + 0x00)
+#define GICC_PMR           (GICC_BASE + 0x04)
+#define GICC_INTACK        (GICC_BASE + 0x0c)
+#define GICC_EOI           (GICC_BASE + 0x10)
+
+/** CPUTargetList[0] corresponds to CPU interface 0 */
+#define CPU_TARGET_LIST_0 (1 << 16) 
+
+void software_init_hook(void) {
+	int i;
+
+	REG_STORE(GICD_CTRL, 0);
+	REG_STORE(GICD_CTRL, 1);
+
+	REG_STORE(GICD_ICENABLER(0), 0xffff0000);
+	REG_STORE(GICD_ISENABLER(0), 0x0000ffff);
+
+	/**
+	 * Set priority on PPI and SGI interrupts
+	 */
+	for (i = 0; i < 32; i += 4) {
+		REG_STORE(GICD_IPRIORITYR(i >> 2), 0xa0a0a0a0);
+	}
+
+	/**
+	 * The GICC_PMR for a CPU interface defines a priority threshold for the target processor.
+	 * The GIC only signals pending interrupts with a higher priority than this threshold value 
+	 * to the target processor.
+	 */
+	REG_STORE(GICC_PMR, 0xf0);
+
+	REG_STORE(GICC_CTRL, 0x1);
+}
+
+static int bpi_intc_init(void) {
+	return 0;
+}
+
+void irqctrl_enable(unsigned int interrupt_nr) {
+	assert(interrupt_nr <= SGI_MAX_SRC);
+	/**
+	 * Writes to bits corresponding to the SGIs are ignored.
+	 * SGIs are always enabled.
+	 */
+}
+
+void irqctrl_disable(unsigned int interrupt_nr) {
+	assert(interrupt_nr <= SGI_MAX_SRC);
+	/** 
+	 * Writing 1 to a Clear-enable bit disables forwarding of 
+	 * the corresponding interrupt from the Distributor to the CPU interfaces
+	 * Writes to bits corresponding to the SGIs are ignored.
+	 * SGIs are always enabled.
+	 *
+	 * REG_STORE(GICD_ICENABLER(), ABC)
+	 */
+
+	/**
+	 * Writing to a Clear-active bit Deactivates the corresponding interrupt
+	 *
+	 * REG_STORE(GICD_ICACTIVER(interrupt_nr >> 5), 1 << (interrupt_nr & 0x1f));
+	 */
+}
+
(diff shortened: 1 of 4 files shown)


src/drivers/interrupt/bpi_intc.c
```diff
@@ -0,0 +1,150 @@
...
+
+#define GICC_CTRL          (GICC_BASE + 0x00)
+#define GICC_PMR           (GICC_BASE + 0x04)
+#define GICC_INTACK        (GICC_BASE + 0x0c)
+#define GICC_EOI           (GICC_BASE + 0x10)
+
+/** CPUTargetList[0] corresponds to CPU interface 0 */
+#define CPU_TARGET_LIST_0 (1 << 16) 
+
+void software_init_hook(void) {
```
anton-bondarev: Please use UNIT_INIT macro instead of software_init_hook

src/drivers/interrupt/bpi_intc.c
```diff
@@ -0,0 +1,150 @@
...
+	 * The GICC_PMR for a CPU interface defines a priority threshold for the target processor.
+	 * The GIC only signals pending interrupts with a higher priority than this threshold value 
+	 * to the target processor.
+	 */
+	REG_STORE(GICC_PMR, 0xf0);
+
+	REG_STORE(GICC_CTRL, 0x1);
+}
+
+static int bpi_intc_init(void) {
```
anton-bondarev: Why do this routing need?

src/drivers/interrupt/bpi_intc.c
```diff
@@ -0,0 +1,150 @@
...
+	/** SGI is made pending by writing to the SGIR */
+	REG_STORE(GICD_SGIR, CPU_TARGET_LIST_0 | (interrupt_nr & 0xf));
+
+	/** 
+	 * Writing to a Set-active bit Activates the corresponding interrupt
+	 * REG_STORE(GICD_ISACTIVER(interrupt_nr >> 5), 1 << (interrupt_nr & 0x1f));
+	 */
+}
+
+void interrupt_handle(void) {
```
anton-bondarev: I think you should use common interrupt_handler routine instead of this code and implement irqctrl_clear function to finish irq treatment using REG_STORE(GICC_EOI, irqstat); .

src/drivers/interrupt/bpi_intc.c
```diff
@@ -0,0 +1,150 @@
...
+
+		ipl_disable();
+
+	}
+
+	critical_leave(CRITICAL_IRQ_HANDLER);
+	critical_dispatch_pending();
+}
+
+void swi_handle(void) {
```
anton-bondarev: This is interrupt handler? If so please extract it to separate file (module)

anton-bondarev:
Thanks for contributing, please fix my notices if you're agree with them

abusalimov:
Thanks for the efforts! I have few comments on the code too, but before that please rebase your commits against `master` and squash them into a single one.

AntonKozlov:
These files directly derived from omap3.
All comments are reasonable, but applied to omap3 as well.
Many of them is about general interrupt handler, but we have no such one.

I consider extracting general intterupt handler from this, omap, and other interrupt controllers as important, but different kind of work, which requires deep understanding of interrupt model and it's cornercases (for all architectures and different arm variants)

IMO, dusscussion should continue until author agree on some comments or argumentedly regret others

AntonKozlov:
I think this is good, old 5-star PR, and there 2 ways for it: merge as is, or reject. I think, it should be in master 
@abusalimov @anton-bondarev

abusalimov:
Basically I don't care. But why are most of commits repeated two times?

abusalimov:
Plus there are some your commits from other branches. If you guarantee that merging all of these won't cause any problems in future for us, then merge.

FWIW, since we've already messed up and just rebase doesn't work, to clean up everything:
```
git checkout feature-bpi-gic-driver
git fetch upstream
git reset --hard upstream/master
git cherry-pick 88620cf c023802 --no-commit
git commit -m'bananapi: Add support for SGI'
```

anton-bondarev:
I'm agree, it should be in master.
But I think the branch should be rebase before merge

AntonKozlov:
Agree, branch have 3 of branch commits have copies. Also, seems like `Add missing option.conf to bpi template` could be deleted (there are no options.conf in current master).
@dimart I'd suggest `git rebase -i upstream/master` and manually adjust commit list.

abusalimov:
@AntonKozlov @dimart I tried to do a simple rebase manually with no luck. See my last message for the working commands to pick the necessary commits, without the unnecessary "Add missing option.conf to bpi template" one.

AntonKozlov:
Hmm, how simple rebase could not work? =)
P.S. I've tried sugested before suggest it, works as expected

abusalimov:
> @dimart I'd suggest git rebase -i upstream/master and manually adjust commit list.
Ah, I missed the `-i`. OK, this would work too, of course. =)

abusalimov:
To summarize,

Option A:
```
git checkout feature-bpi-gic-driver
git fetch upstream
git reset --hard upstream/master
git cherry-pick 88620cf c023802 --no-commit
git commit -m'bananapi: Add support for SGI'
```

Option B (equivalent):
```
git checkout feature-bpi-gic-driver
git fetch upstream
git rebase -i upstream/master

    # Remove everything from an opened todo-file,
    # enter the following and save:
    pick    88620cf  BananaPi port: working SGI
    fixup   c023802  Fix code style
```

Check that everything's fine, there must be a single commit:
```console
$ git log --oneline upstream/master..HEAD
1243401 bananapi: Add support for SGI
```

dimart:
@AntonKozlov @abusalimov thanks for your help! 
That looks alright now.

src/drivers/interrupt/bpi_intc.c
abusalimov: I guess the content of `software_init_hook` should have been moved into this one, shouldn't it?
@anton-bondarev what do you think?

src/drivers/interrupt/bpi_intc.h
```diff
@@ -0,0 +1,14 @@
...
+ * @brief Banana Pi interrupt controller HAL definitions.
+ *
+ * @date 15.04.15
+ * @author Dmitrii Petukhov
+ */
+
+#ifndef IRQCTRL_BPI_INTC_IMPL_H_
+#define IRQCTRL_BPI_INTC_IMPL_H_
+
+#define __IRQCTRL_IRQS_TOTAL 123
```
abusalimov: Are there really 123 interrupt lines? If so, why do `irqctrl_xxx` methods contain `assert(interrupt_nr < 16);`?

src/drivers/interrupt/bpi_intc.c
```diff
@@ -0,0 +1,150 @@
...
+	 * REG_STORE(GICD_ICACTIVER(interrupt_nr >> 5), 1 << (interrupt_nr & 0x1f));
+	 */
+}
+
+void irqctrl_clear(unsigned int interrupt_nr) {
+	assert(interrupt_nr < 16);
+}
+
+void irqctrl_force(unsigned int interrupt_nr) {
+	assert(interrupt_nr < 16);
```
abusalimov: Shoudn't there be `IRQCTRL_IRQS_TOTAL` used instead of 16?

abusalimov:
@dimart Look great now!

I have few remarks to the code so far, could fix or provide comments on them?

src/drivers/interrupt/bpi_intc.c
dimart: As pull request title says, at this moment driver handles only software generated interrupts (SGI). GIC-400, general interrupt controller of BananaPi, supports only 16 SGI, that's why the assertion's here.

src/drivers/interrupt/bpi_intc.h
dimart: Yes, there are. :)
Check this out: http://lib.chipdip.ru/974/DOC000974439.pdf (page 147)
I've answered the second question above.

src/drivers/interrupt/bpi_intc.c
abusalimov: Ah, OK, now I get it, thanks!

It would be more clear if all occurrences of magic `16` are replaced with some `#define`, like `GIC_SGI_TOTAL` (probably you could suggest a better name than me).

dimart: Definitely, you're right!
Should I just commit & push to "feature-bpi-gic-driver" branch?

abusalimov: Yes, sure. Just name your commit as the previous one with a `fixup! ` prefix, that is `fixup! bananapi: Add support for SGI`. More about such commits [here](https://github.com/embox/embox/blob/721-contributing-guide/CONTRIBUTING.md#commit---fixup-followed-by-rebase---autosquash).

dimart: Didn't know, thanks!

abusalimov: You're welcome :)

abusalimov:
@AntonKozlov @anton-bondarev What's about moving the code from `software_init_hook` into `bpi_intc_init` (see [above](#discussion-diff-29494869))? Would it work as expected in this case?

Apart from this, LGTM from me. :+1:

anton-bondarev:
I suggested this earlier, but @AntonKozlov said that it has potential problems during initialization sequence. But if it works on a real board I think it looks good

AntonKozlov:
Yeah, it's general workaround about loader, that could use interrupts (i.e. uboot). This hook used to initalize tables before `ipl_init` issued before this `bpi_intc_init` called by initializer call.
I think, it's done. @dimart, thank you for your work!

PR #746, opened 2015-05-01
Lthread minor

vloginova:
- Fix buildbot failing because of lthread
 - Make resume and yield functions inline.


anton-bondarev:
LGTM

PR #751, opened 2015-05-15
Ledbox abort reset

AntonKozlov:


abusalimov:
LGTM aside from a minor comment above.

PR #753, opened 2015-05-19
Ledbox settings quickfix

AntonKozlov:
Quick resolution of not working dfs-based flash settings


AntonKozlov:
outdated, see #751

PR #758, opened 2015-05-28
File system clean

anton-bondarev:
Remove ffs &etfs file sytem drivers
Remove wrong asserts from index_operation.c
Clean headers a little


Deryugin:
Template x86/test/packetdrill is now broken beacuse of missing dependency

src/drivers/net/tun.c: In function ‘tun_dev_read’:
src/drivers/net/tun.c:171:9: error: implicit declaration of function ‘thread_self’ [-Werror=implicit-function-declaration]
  struct waitq_link *wql = &thread_self()->schedee.waitq_link;
         ^
src/drivers/net/tun.c:171:41: error: invalid type argument of ‘->’ (have ‘int’)
  struct waitq_link *wql = &thread_self()->schedee.waitq_link;

PR #762, opened 2015-05-29
e1000: Fix e1000 big packet receive

AntonKozlov:


src/drivers/net/e1000.c
```diff
@@ -197,16 +201,18 @@ static void e1000_rx(struct net_device *dev) {
...
+			rx_skbs[cur] = skb_alloc(E1000_MAX_RX_LEN);
 			if (rx_skbs[cur] == NULL)
 				panic("Can't allocate rx_skbs %d", cur);
 			rx_descs[cur].buffer_address = (uint32_t) rx_skbs[cur]->mac.raw;
 
-
+			len = rx_descs[cur].length - E1000_RX_CHECKSUM_LEN;
+			skb = skb_realloc(len, skb);
+			if (!skb) {
+				panic("Can't reallocate rx_skb");
```
abusalimov: That's weird anyway. I.e. instead of dropping the packet, we just drop the system? :)

AntonKozlov: You are right, changed again

src/drivers/net/e1000.c
```diff
@@ -197,16 +201,17 @@ static void e1000_rx(struct net_device *dev) {
...
 
-
-			skb->dev = dev;
-			netif_rx(skb);
-
-
+			len = rx_descs[cur].length - E1000_RX_CHECKSUM_LEN;
+			skb = skb_realloc(len, skb);
+			if (skb == NULL) {
+				skb->dev = dev;
```
abusalimov: I guess, you already hate me, but this is a NULL pointer dereference. :fearful: 

How about this?
```c
if (!skb)
    goto drop_pack;
```

AntonKozlov:
No, we are having good conversation here, if you ask me =)
I'll adopt your snippet

abusalimov:
:beers:

AntonKozlov:
Please close PR without merge, so I'll be able to rebase it

abusalimov:
I think rebasing the branch itself and updating this PR in-place is OK since I commented of diffs, not on commits: diff comments shouldn't disappear.

abusalimov:
Discussing on one PR but then merging another one doesn't feel somewhat right, after all... Anyway, this decision is up to you, do what is more convenient for you in this case. :8ball:

src/drivers/net/e1000.c
```diff
@@ -197,16 +201,19 @@ static void e1000_rx(struct net_device *dev) {
 			skb = rx_skbs[cur];
 			assert (skb);
 
-			rx_skbs[cur]= skb_alloc(ETH_FRAME_LEN);
+			rx_skbs[cur] = skb_alloc(E1000_MAX_RX_LEN);
 			if (rx_skbs[cur] == NULL)
 				panic("Can't allocate rx_skbs %d", cur);
```
abusalimov: I believe, this should be handled through dropping a packet too.

abusalimov:
I'm confused about [this](https://github.com/embox/embox/blob/2f1c976bb874922256dc6a54edcec956d8360b2a/src/drivers/net/e1000.c#L201-L210) snippet from `e1000_rx`:
```c
			skb = rx_skbs[cur];
			assert (skb);

			rx_skbs[cur] = skb_alloc(E1000_MAX_RX_LEN);
			if (rx_skbs[cur] == NULL)
				panic("Can't allocate rx_skbs %d", cur);
			rx_descs[cur].buffer_address = (uint32_t) rx_skbs[cur]->mac.raw;

			len = rx_descs[cur].length - E1000_RX_CHECKSUM_LEN;
			skb = skb_realloc(len, skb);
```

Does it mean that the driver manages an intermediate pool of pre-allocated skbs? But what is the motivation of doing that? DMA?

src/drivers/net/e1000.c
```diff
@@ -284,7 +291,7 @@ static int e1000_open(struct net_device *dev) {
 
 	for (size_t i = 0; i < E1000_RXDESC_NR; i ++) {
 	        struct sk_buff *skb;
-		skb = skb_alloc(ETH_FRAME_LEN);
+		skb = skb_alloc(E1000_MAX_RX_LEN);
 		if (skb == NULL)
 			panic("Can't allocate skb %d", i);
```
abusalimov: And this one should probably refuse to open the device, i.e. throw ENOMEM maybe?

AntonKozlov: yeah, this is right

src/drivers/net/e1000.c
```diff
@@ -283,11 +288,11 @@ static int e1000_open(struct net_device *dev) {
...
-		if (skb == NULL)
-			panic("Can't allocate skb %d", i);
-		rx_skbs[i]= skb;
+	        struct sk_buff *skb = skb_alloc(E1000_MAX_RX_LEN);
+		if (skb == NULL) {
+			return -ENOMEM;
+		}
+		rx_skbs[i] = skb;
 		rx_descs[i].buffer_address = (uint32_t) skb->mac.raw;
 	}
```
abusalimov: I've just realized that this shouldn't be done in `e1000_open` at all, since it initializes some *global* structures and doesn't use `dev` being opened. This should be in `e1000_init`, am I right?
Anyway, just breaking the loop and returning is not graceful enough, and one should clean up already allocated data, i.e. rollback the loop and `skb_free` each item from (i .. 0].

src/drivers/net/e1000.c
```diff
@@ -282,18 +297,18 @@ static int e1000_open(struct net_device *dev) {
...
-		rx_descs[i].buffer_address = (uint32_t) skb->mac.raw;
+	for (int i = 0; i < E1000_RXDESC_NR; i ++) {
+	        struct sk_buff *skb = skb_alloc(E1000_MAX_RX_LEN);
+		if (skb == NULL) {
+			e1000_stop(dev);
+			return -ENOMEM;
+		}
+		nic_priv->rx_skbs[i] = skb;
+		nic_priv->rx_descs[i].buffer_address = (uint32_t) skb->mac.raw;
 	}
```
abusalimov: Just for the memories: it was suggested to move the buffer allocation logic to the beginning of the `e1000_open`, before setting up any hardware registers.
It is also a bit confusing that in case of an allocation error, `e1000_open` calls `e1000_stop`, which in turn also touches some registers. `E1000_REG_CTRL_RST` may be harmless though, I'm not sure about that.

abusalimov:
It would be nice if you amend the last commit, but even without fixing that remark on `e1000_open` <--> `e1000_stop` IMO the PR is ready to merge, @AntonKozlov please take the final decision.
Except for that, LGTM :shipit:

src/drivers/net/e1000.c
AntonKozlov: Card should be stopped before skb's could be freed.
What you suggest should only result in e1000_open change, e1000_close may be not called from e1000_open, but it will stay same

abusalimov: Sounds reasonable, now I get it.

PR #763, opened 2015-05-31
usbnet driver

alexkalmuk:
Purpose: Obtain ping between Embox and Linux through usbnet driver.

What 's done:
- Added usbnet (Ethernet over USB) driver
- Added USB CDC class

### DIFF
--- a/src/cmds/hardware/lsusb/lsusb.c
+++ b/src/cmds/hardware/lsusb/lsusb.c
@@ -1,0 +1,54 @@
+/**
+ * @file
+ *
+ * @brief Show list of USB devices
+ *
+ * @date 14.05.2015
+ * @author Anton Bondarev
+ */
+#include <stdio.h>
+#include <unistd.h>
+
+#include <drivers/usb/usb.h>
+
+static void print_usage(void) {
+	printf("Usage: lsusb [-h]\n");
+	printf("\t[-h]      - print this help\n");
+}
+
+static void print_error(void) {
+	printf("Wrong parameters\n");
+	print_usage();
+}
+
+static void show_usb_dev(struct usb_dev *usb_dev) {
+	printf("Bus %03d Device %03d ID %04x:%04x\n",
+			usb_dev->bus_idx,
+			usb_dev->idx,
+			usb_dev->dev_desc.id_vendor,
+			usb_dev->dev_desc.id_product);
+
+}
+
+int main(int argc, char **argv) {
+	struct usb_dev *usb_dev = NULL;
+	int opt;
+
+	while (-1 != (opt = getopt(argc, argv, "h"))) {
+		switch (opt) {
+		case '?':
+		case 'h':
+			print_usage();
+			return 0;
+		default:
+			print_error();
+			return 0;
+		}
+	}
+
+	while ((usb_dev = usb_dev_iterate(usb_dev))) {
+		show_usb_dev(usb_dev);
+	}
+
+	return 0;
+}
--- a/src/drivers/net/usbnet.c
+++ b/src/drivers/net/usbnet.c
@@ -80,12 +80,13 @@ static void usbnet_rcv_notify(struct usb_request *req, void *arg) {
 		size_t len;
 
 		len = nic_priv->pdata - nic_priv->data;
-		if ((skb = skb_alloc(len))) {
+		skb = skb_alloc(len);
+		if (!skb) {
+			printk("usbnet: skbuff allocation failed\n");
+		} else {
 			memcpy(skb->mac.raw, nic_priv->data, len);
 			skb->dev = nic;
 			netif_rx(skb);
-		} else {
-			printk("usbnet: skbuff allocation failed\n");
 		}
 
 		/* Prepare for copying new packet */
@@ -126,35 +127,38 @@ static int usbnet_probe(struct usb_driver *drv, struct usb_dev *dev,
 
 	nic = (struct net_device *) etherdev_alloc(sizeof *nic_priv);
 	if (!nic) {
-		return -ENOMEM;
+		res = -ENOMEM;
+		goto out_ret;
 	}
 	nic->drv_ops = &usbnet_drv_ops;
 	nic_priv = netdev_priv(nic, struct usbnet_priv);
 	nic_priv->usbdev = dev;
 	nic_priv->data = nic_priv->pdata = sysmalloc(ETH_FRAME_LEN);
 	if (!nic_priv->data) {
-		etherdev_free(nic);
-		return -ENOMEM;
+		res = -ENOMEM;
+		goto out_free_nic;
 	}
 
 	cdc->privdata = (void *) nic;
 
 	res = inetdev_register_dev(nic);
 	if (res < 0) {
-		goto err_out;
+		goto out_free_data;
 	}
 
-    timer_init_msec(&nic_priv->timer, TIMER_ONESHOT,
-            USBNET_TIMER_FREQ, usbnet_timer_handler, nic_priv);
+	res = timer_init_msec(&nic_priv->timer, TIMER_ONESHOT,
+		USBNET_TIMER_FREQ, usbnet_timer_handler, nic_priv);
 	if (res < 0) {
-		goto err_out;
+		goto out_free_data;
 	}
 
-    return 0;
+	return 0;
 
-err_out:
+out_free_data:
 	sysfree(nic_priv->data);
+out_free_nic:
 	etherdev_free(nic);
+out_ret:
 	return res;
 }
 
@@ -172,16 +176,16 @@ static void usbnet_send_notify_hnd(struct usb_request *req, void *arg) {
 static void usb_net_bulk_send(struct usb_dev *dev, struct sk_buff *skb) {
 	struct usb_endp *endp;
 	size_t i = 0;
-	int len;
+	size_t len;
 
 	endp = dev->endpoints[3];
 
-	for (i = skb->len, len = min(skb->len, endp->max_packet_size);
-			i != 0; i -= len, len = min(i, endp->max_packet_size)) {
+	for (i = skb->len; i != 0 && (len = min(i, endp->max_packet_size)); i -= len) {
 		usb_endp_bulk(endp, usbnet_send_notify_hnd,
 				skb->mac.raw + (skb->len - i), len);
 	}
 
+	/* Send zero length packet if skb->len % max_packet_size == 0 */
 	if (len == endp->max_packet_size) {
 		usb_endp_bulk(endp, usbnet_send_notify_hnd, skb->mac.raw + i, 0);
 	}
--- a/src/drivers/net/usbnet.c
+++ b/src/drivers/net/usbnet.c
@@ -8,11 +8,9 @@
 #include <assert.h>
 #include <errno.h>
 #include <string.h>
-
 #include <util/math.h>
 
 #include <net/l2/ethernet.h>
-#include <net/l3/arp.h>
 #include <net/netdevice.h>
 #include <net/inetdevice.h>
 #include <net/skbuff.h>
@@ -45,8 +43,11 @@ static int usbnet_xmit(struct net_device *dev, struct sk_buff *skb);
 
 struct usbnet_priv {
 	struct usb_dev *usbdev;
+	/* Storage for incoming packet */
 	char *data;
+	/* Current position in packet */
 	char *pdata;
+	/* Timer for polling */
 	struct sys_timer timer;
 };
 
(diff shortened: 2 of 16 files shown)


src/cmds/hardware/lsusb/lsusb.c
```diff
@@ -0,0 +1,54 @@
...
+		case 'h':
+			print_usage();
+			return 0;
+		default:
+			print_error();
+			return 0;
+		}
+	}
+
+	for (usb_dev = usb_dev_iterate(NULL); usb_dev; usb_dev = usb_dev_iterate(usb_dev)) {
```
abusalimov: Just to clarify and check myself, is it the same as doing the following?
```c
struct usb_dev *usb_dev = NULL;
...

while ((usb_dev = usb_dev_iterate(usb_dev))) {
    ...
}
```

src/drivers/net/usbnet.c
```diff
@@ -0,0 +1,212 @@
...
+		return;
+	}
+
+	/* End of packet */
+	if (req->len < in_endp->max_packet_size) {
+		struct sk_buff *skb;
+		size_t len;
+
+		len = nic_priv->pdata - nic_priv->data;
+		if ((skb = skb_alloc(len))) {
```
abusalimov: IMO, it's better to stick to the common pattern:
```c
skb = skb_alloc(len);
if (!skb) {
    printk("usbnet: skbuff allocation failed\n");
} else {
    ...
}
```

I.e. deal with possible errors just after the allocation and forget about them right away.

src/drivers/net/usbnet.c
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	cdc->privdata = (void *) nic;
+
+	res = inetdev_register_dev(nic);
+	if (res < 0) {
+		goto err_out;
+	}
+
+    timer_init_msec(&nic_priv->timer, TIMER_ONESHOT,
```
abusalimov: bad indentation

src/drivers/net/usbnet.c
```diff
@@ -0,0 +1,212 @@
...
+            USBNET_TIMER_FREQ, usbnet_timer_handler, nic_priv);
+	if (res < 0) {
```
abusalimov: Copy paste of the above check? Or missing `res = timer_init_msec(...)`?

src/drivers/net/usbnet.c
```diff
@@ -0,0 +1,212 @@
...
+	nic = (struct net_device *) etherdev_alloc(sizeof *nic_priv);
+	if (!nic) {
+		return -ENOMEM;
+	}
+	nic->drv_ops = &usbnet_drv_ops;
+	nic_priv = netdev_priv(nic, struct usbnet_priv);
+	nic_priv->usbdev = dev;
+	nic_priv->data = nic_priv->pdata = sysmalloc(ETH_FRAME_LEN);
+	if (!nic_priv->data) {
+		etherdev_free(nic);
```
abusalimov: Since you already use `goto` below, maybe it's worth using single return point everywhere, including this one?

src/drivers/net/usbnet.c
```diff
@@ -0,0 +1,212 @@
...
+	{ },
+};
+
+static void usbnet_send_notify_hnd(struct usb_request *req, void *arg) {
+}
+
+static void usb_net_bulk_send(struct usb_dev *dev, struct sk_buff *skb) {
+	struct usb_endp *endp;
+	size_t i = 0;
+	int len;
```
abusalimov: `i` is `size_t`, but `len` is `int`? This looks weird.

src/drivers/net/usbnet.c
```diff
@@ -0,0 +1,212 @@
...
+
+	endp = dev->endpoints[3];
+
+	for (i = skb->len, len = min(skb->len, endp->max_packet_size);
+			i != 0; i -= len, len = min(i, endp->max_packet_size)) {
```
abusalimov:
```c
i = skb->len;
while (i) {
   len = min(i, endp->max_packet_size);
   ...
   i -= len;
}
```

or
```c
for (i = skb->len; len = min(i, endp->max_packet_size), i != 0; i -= len)
    ...
```

src/drivers/net/usbnet.c
```diff
@@ -0,0 +1,212 @@
...
+		usb_endp_bulk(endp, usbnet_send_notify_hnd,
+				skb->mac.raw + (skb->len - i), len);
+	}
+
+	if (len == endp->max_packet_size) {
```
abusalimov: Is this possible?

src/drivers/usb/class/usb_class_cdc.c
```diff
@@ -12,100 +14,244 @@
...
+		if (type == CS_INTERFACE) {
+			len = *cur;
+			cur += len;
+		} else {
+			break;
+		}
+	}
+
+	return (cur - start);
 }
```
abusalimov: How about this?
```c
static size_t cdc_skip_type(uint8_t *start, uint8_t skip_type) {
	uint8_t *cur = start;

	while (cur[1] == skip_type) {
		uint8_t len = *cur;
		cur += len;
	}

	return (cur - start);
}
```

src/drivers/usb/class/usb_class_cdc.c
```diff
@@ -12,100 +14,244 @@
...
-	printk("line_coding - %d\n", line_coding);
+struct usb_desc_interface *cdc_get_interface(
+		struct usb_desc_configuration *conf, size_t index) {
+	size_t cur = 0;
+	char *curp = (char *) conf + INTERFACE_DESC_OFFSET;
+
+	while (cur != index) {
+		curp += sizeof (struct usb_desc_interface);
+		curp += cdc_skip_func_descs(curp);
+		curp += cdc_skip_endpoints(curp);
```
abusalimov:
```c
curp += cdc_skip_type(curp, CS_INTERFACE);
curp += cdc_skip_type(curp, USB_DESC_TYPE_ENDPOINT);
```
BTW, is it always true, that func descs go _before_ endpoints?

src/drivers/usb/class/usb_class_cdc.c
```diff
@@ -12,100 +14,244 @@
...
+	 * conf->b_num_interfaces + alternated settings */
+	for (i = 0; i < conf->b_num_interfaces + 1; i++) {
+		iface = cdc_get_interface((struct usb_desc_configuration *) cdc->getconf, i);
+		endpoints = (char *) iface + sizeof (struct usb_desc_interface);
+		endpoints += cdc_skip_func_descs(endpoints);
 
-static void usb_class_cdc_get_conf_hnd(struct usb_request *req, void *arg) {
+		for (j = 0; j < iface->b_num_endpoints; j++) {
+			ep = (struct usb_desc_endpoint *) (endpoints +
+					 j * sizeof (struct usb_desc_endpoint));
```
abusalimov: Pointer arithmetic:
```c
struct usb_desc_endpoint *endpoint = (struct usb_desc_endpoint *) endpoints + j;
```

src/drivers/usb/class/usb_class_cdc.c
```diff
@@ -12,100 +14,244 @@
...
+
+static void cdc_get_conf_content_hnd(struct usb_request *req, void *arg) {
+	struct usb_dev *dev = req->endp->dev;
+	struct usb_class_cdc *cdc = usb2cdcdata(dev);
+	struct usb_desc_configuration *c;
+
+	c = (struct usb_desc_configuration*) cdc_conf;
+
+	cdc->getconf = sysmalloc(c->w_total_length);
+	assert(cdc->getconf);
```
abusalimov: Are you _really_ sure that `sysmalloc` _always_ succeeds? :)

src/drivers/usb/usb_pnp.c
```diff
@@ -77,15 +85,28 @@ void usb_dev_addr_settled(struct usb_dev *dev) {
 	usb_dev_configure(dev);
 }
 
+static int usb_add_newdev_action(struct lthread *self) {
+	struct usb_dev *dev;
+
+	dlist_foreach_entry(dev, &usb_newdev_list, dev_link) {
+		dlist_del(&dev->dev_link);
+		dlist_head_init(&dev->dev_link);
```
abusalimov: There's `dlist_del_init` for that, if I'm not mistaken.

src/drivers/net/usbnet.c
alexkalmuk: Thanks, there is missing <code>res = timer_init_msec(...)</code>

src/drivers/usb/class/usb_class_cdc.c
alexkalmuk: Thank you, it looks better.

src/drivers/usb/class/usb_class_cdc.c
alexkalmuk: Yes. Endpoints always go after interface, while func descs are part of interface.

src/drivers/usb/class/usb_class_cdc.c
alexkalmuk: Have no idea why assert is placed here :) I use <code>if (!data)</code> in other places (e.g., in usbnet.c)

src/drivers/net/usbnet.c
alexkalmuk: You are right, it is impossible. There should be placed check if <code>skb->len</code> is multiple of <code>endp->max_packet_size</code>. In that case we must send zero length packet.

abusalimov: I didn't quite get that.
Do you mean that an empty packet should be sent in case when `skb` is bigger than the maximum allowed size? That is, if the loop above runs more than one iteration, right?

alexkalmuk: No, I mean that  an empty packet should be sent in case when <code>skb->len == K * max_packet_size</code>

abusalimov: Ah, I see now...

Then, I guess, this will work again as is after changing the above loop to
```c
for (i = skb->len; i != 0 && (len = min(i, endp->max_packet_size)); i -= len)
    ...
```
or to a `while` version, which I've suggested earlier.

src/cmds/hardware/lsusb/lsusb.c
alexkalmuk: Yes, you are right, it is the same.

src/drivers/usb/usb_pnp.c
alexkalmuk: Hm, I looked at src/util/dlist/dlist_debug.h: dlist_del_init and what I see is that dlist_del_init calls dlist_init after dlist_del, not dlist_head_init.

abusalimov: Well, honestly, I don't quite understand the difference, although I won't insist on this. :)

alexkalmuk: I wonder why dlist_init is there, it doesn't detach deleted head from the list. Okay, for safety reason I leave that piece of code as is.

alexkalmuk:
PING

src/drivers/net/usbnet.c
```diff
@@ -0,0 +1,216 @@
...
+	struct usb_class_cdc *cdc = usb2cdcdata(dev);
+	int res;
+
+	/* Enable in/out DATA interface */
+	cdc_set_interface(dev, 1, 1, usbnet_iface_hnd);
+
+	nic = (struct net_device *) etherdev_alloc(sizeof *nic_priv);
+	if (!nic) {
+		res = -ENOMEM;
+		goto err_out;
```
abusalimov: When jumping from here `nic_priv` is uninitialized, but used within clean up logic.

src/drivers/net/usbnet.c
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	return 0;
+
+err_out:
+	if (nic_priv->data)
+		sysfree(nic_priv->data);
+	if (nic)
+		etherdev_free(nic);
+	return res;
```
abusalimov: How about this:
```c
out_data:
	sysfree(nic_priv->data);
out_nic:
	etherdev_free(nic);
out_ret:
	return res;
```
And just using these labels above appropriately.

abusalimov:
OK, I think, everything's good now!
It would be nice if you organize your commits and squash follow-up changes before merging.
Anyway, LGTM.

src/drivers/net/usbnet.c
alexkalmuk: ok, fixed as proposed below

PR #765, opened 2015-06-01
dvfs: Fix dvfs open build

AntonKozlov:
dvfs open fails to build, at least at arm/stm32_modbus


Deryugin:
Well, we three header files with the same name:

https://github.com/embox/embox/blob/master/src/kernel/task/resource/idesc_table.h
https://github.com/embox/embox/blob/master/src/include/kernel/task/resource/idesc_table.h
https://github.com/embox/embox/blob/master/src/include/kernel/task/idesc_table.h

I guess some of this headers should be merged together to make it clear.

AntonKozlov:
Refer to #767 also

PR #766, opened 2015-06-01
mk: oldconf: Fix an undefined 'config_h' variable warning

anton-bondarev:
The warning was introduced in e51a50a ("mk: oldconf: Remove processing of option.conf from scripts / macros") merged through #750.


abusalimov:
:shipit:

PR #773, opened 2015-06-16
Ipl restore fix

vloginova:
Previously the function restored ipl only in case disabled IRQs in several architectures, that might cause problems from time to time. This request fixes the issue for such architectures, namely mips, x86, ppc and usermode.


anton-bondarev:
LGTM

PR #778, opened 2015-06-22
Add timer "low-level" interface: init, start

AntonKozlov:


abusalimov:
`timer_do_init`? Why not renaming the old `timer_init` into `timer_init_and_start` instead?

PR #780, opened 2015-06-29
Logger

vloginova:
Introducing new logging mechanism.


src/cmds/net/httpd/Httpd.my
```diff
@@ -17,6 +17,7 @@ package embox.cmd.net
 			Ilia Vaprol
 	''')
 module httpd {
+	option number log_level=0
```
AntonKozlov: Old default level was debug

src/drivers/clock/hpet.c
```diff
@@ -89,13 +90,11 @@ static int hpet_init(void) {
 	hpet_counter_device.cycle_hz = hpet_get_hz();
 	hpet_start_counter();
 
-#ifdef HPET_DEBUG
-	printk("Hz: %u\n", hpet_counter_device.cycle_hz);
+	log_debug("Hz: %u", hpet_counter_device.cycle_hz);
 	for (int i = 0; i < 10; ++i) {
 		ksleep(1000);
```
AntonKozlov: new code is not equalent old, ksleep called in debug and release

src/drivers/net/Mybuild
```diff
@@ -139,6 +140,8 @@ module usermode {
 }
 
 module lan9118 {
+	option number log_level=0
```
AntonKozlov: this default should be `info`, to keep messages about mac

src/util/log.h
```diff
@@ -0,0 +1,39 @@
...
+#include <framework/mod/self.h>
+#include <util/logging.h>
+
+struct logger {
+	const struct mod *mod;
+	struct logging logging;
+};
+
+#define LOG_DECLARE_LOGGER() \
+	extern struct logger mod_logger __attribute__ ((weak)); \
```
AntonKozlov: extra semicolon and backslash

src/util/log.h
```diff
@@ -0,0 +1,39 @@
...
+
+#define log_logp(level, fmt, ...) \
+	logging_raw(&mod_logger.logging, level, "%s: %s: " fmt "\n", \
```
AntonKozlov: in case if mod_logger is not defined (weak ref turns it to NULL), seems like its NULL reference

PR #784, opened 2015-07-02
mybuild: gen: Define alises for options inherited from supertypes

abusalimov:
That is, given two modules:

    module api {
        option foo
    }

    module impl extends api {
        option bar
    }

There are two options defined for instances of module `impl`:
its own option `bar` and an inherited `foo` option.

Before this change, one could only use `OPTION(api, foo)` to access a
value of `foo` from C-code at compile time. Now it is also aliased
through `OPTION(impl, foo)`, even though `impl` doesn't define `foo`
explicitly.


abusalimov:
Once applied, this should also fix `cleanup-mybuild-files` branch failing to build.

PR #785, opened 2015-07-02
Cleanup mybuild files

abusalimov:
Remove some rarely used features of Mybuild like imports or features. Also avoid clashes between module and package names that could only lead to ambiguity.

The only changes that affect the logic are in the last two commits. Though it'd be nice if you also check that I didn't miss anything when renaming modules (i.e. references to options in C code).


anton-bondarev:
LGTM

PR #787, opened 2015-07-02
Add Irqctrl pending() function to the 'irqctrl' API

abusalimov:
This allows us to throw away `<drivers/i8259.h>`.


anton-bondarev:
LGTM

PR #737, opened 2015-04-23
Add httpd with pthread

Nastasya-Vinogradik:

PR #791, opened 2015-07-07
737 merge

AntonKozlov:
This aims to merge #737. Although there is some changes around #737, functionality nearly equivalent.
Note, that cgi not supported at all. Threads allows keep socket in server and open pipes to child task, this way should be considered as slow but secure, probably it could be implemented.


anton-bondarev:
LGTM

AntonKozlov:
just for note, this branch started from 4ea8e6b9a0214d06f25d3464370f4f8131b50004

PR #793, opened 2015-07-13
705 raspberry port

vloginova:
Start porting Embox on Raspberry PI. These changes include essential hardware specific modules:
 - realization of the diag interface
 - interrupt controller
 - driver for the System Timer.


src/drivers/serial/raspi_uart.c
```diff
@@ -0,0 +1,99 @@
...
+#define UART0_FR     UART0_OFFSET + 0x18 /* Flag register */
+#define UART0_IBRD   UART0_OFFSET + 0x24 /* Integer Baud rate divisor */
+#define UART0_FBRD   UART0_OFFSET + 0x28 /* Fractional Baud rate divisor */
+#define UART0_LCRH   UART0_OFFSET + 0x2C /* Line Control register */
+#define UART0_CR     UART0_OFFSET + 0x30 /* Control register */
+#define UART0_IMSC   UART0_OFFSET + 0x38 /* Interrupt Mask Set Clear Register */
+#define UART0_ICR    UART0_OFFSET + 0x44 /* Interrupt Clear Register */
+
+
+static void delay(int32_t count) {
```
anton-bondarev: Assembler for delay routine in not very good idea. Can it be changed to more suit version?

src/drivers/clock/raspi_systick.c
```diff
@@ -0,0 +1,108 @@
...
+
+#include <hal/clock.h>
+#include <hal/reg.h>
+#include <hal/system.h>
+#include <kernel/irq.h>
+#include <kernel/time/clock_source.h>
+
+#include <embox/unit.h>
+
+#define BCM2835_SYSTEM_TIMER_BASE 0x20003000
```
anton-bondarev: May be extract it as module option?

src/arch/arm/raspi/cpu_reset.S
```diff
@@ -0,0 +1,20 @@
+/**
+ * @file
+ * @brief
+ *
+ * @author  Vita Loginova
+ * @date    09.07.2015
+ */
+
+.global hardware_init_hook
+hardware_init_hook:
```
anton-bondarev: It's a boot code?

anton-bondarev:
LGTM

src/drivers/clock/raspi_systick.c
vloginova: Why? It is a constant value for this chip. I don't think that there is a point to do that. At least for now.

src/arch/arm/raspi/cpu_reset.S
vloginova: Yes, see https://github.com/embox/embox/blob/master/src/arch/arm/armlib/head.S

src/drivers/serial/raspi_uart.c
vloginova: Yes, you're right, I'll fix it in a moment

src/drivers/clock/raspi_systick.c
anton-bondarev: Yes, it a chip constant. But not a board, another boards base on BCM2835 chip have the same base address. May be rename module as bcm2835_systick?

src/arch/arm/raspi/cpu_reset.S
anton-bondarev: Ok

src/drivers/clock/raspi_systick.c
vloginova: Then I should rename all these modules a soon as the all part of RPi peripherals. Should I do it? It seems like there is no other board with bcm285 (at least popular, I couldn't find it).

anton-bondarev: Shouldn't. May be in future.
Ok.

PR #796, opened 2015-07-20
Disable piggy build target by default

### LINKED ISSUE #789: Disable piggy build target by default
piggy target sometime fails the build on some arch (arm, arm-embedded compiler; msp). Why its built by default? Could be disabled?
@abusalimov @anton-bondarev

AntonKozlov:
Closes #789


anton-bondarev:
LGTM

PR #797, opened 2015-07-21
Framework clean net

anton-bondarev:
Remove heavy framework/net objects. It's also delete part of an old "__imple_x" problem


AntonKozlov:
:+1:

PR #805, opened 2015-08-10
Multiclet prepare

AntonKozlov:
Preparation to multiclet:
* compiler abstraction
* gcc specific features removed


anton-bondarev:
:shipit:

PR #807, opened 2015-08-11
moved headers from include/drivers/bluetooth into src/drivers

mrgaz:
moved headers from include/drivers/bluetooth into src/drivers


abusalimov:
Please, read commit message agreements [here](https://github.com/embox/embox/blob/721-contributing-guide/CONTRIBUTING.md#commit-message-agreements):
  - Start the subject with a scope / subsystem
  - Capitalize the sentence after the scope / labels
  - Use the imperative mood in the subject line. *Fix*, not *fixing/fixed*

PR #819, opened 2015-08-18
Restore test no autorun abillity

AntonKozlov:


AntonKozlov:
Only place where it's used is fs_testr/w. Another option is to use usual SUITE registration with autorun disabled via runlevel manipulation. #820

PR #823, opened 2015-08-24
Integrity check

AntonKozlov:


AntonKozlov:
Oops, dynamic text start isn't implemented yet

PR #835, opened 2015-09-29
Autoqemu comments handling fix

Deryugin:
There were some problems with commented options. For example, it you will replace

@Runlevel(2) include embox.driver.net.virtio

with

//@Runlevel(2) include embox.driver.net.virtio

(i.e. comment this string), autoqemu will try to use nic type "//virtio"

1) Original regex was not correct
2) There is no need to analyze commented strings

This branch fixes this issue


AntonKozlov:
:turtle: LGTM

PR #840, opened 2015-10-08
Rootfs automount

Deryugin:
Implement automatic mount of required FS on system initialization.
Corresponding folders are now being created even if root file system is read-only.


anton-bondarev:
LGTM

PR #847, opened 2015-10-14
fatfs file create error fix

Deryugin:
Fix error on file create. Error was related to ramdisk block size (default 4096, and fat was expecting 512).

### DIFF
--- a/src/drivers/block_dev/ramdisk/ramdisk.c
+++ b/src/drivers/block_dev/ramdisk/ramdisk.c
@@ -77,7 +77,6 @@ struct ramdisk *ramdisk_create(char *path, size_t size) {
 	}
 
 	ramdisk->blocks = ramdisk_size / RAMDISK_BLOCK_SIZE;
-	ramdisk->block_size = RAMDISK_BLOCK_SIZE;
 
 	ramdisk->p_start_addr = phymem_alloc(page_n);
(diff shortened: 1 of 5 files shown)


src/drivers/block_dev/ramdisk/ramdisk.c
```diff
@@ -167,7 +166,7 @@ static int read_sectors(block_dev_t *bdev,
 	char *read_addr;
 
 	ramdisk = (ramdisk_t *) bdev->privdata;
-	read_addr = ramdisk->p_start_addr + (blkno * ramdisk->block_size);
+	read_addr = ramdisk->p_start_addr + (blkno * bdev->block_size);
```
alexkalmuk: It seems something is wrong with ramdisk->block_size. If I understand right ramdisk->block_size and bdev->block_size mean the same thing. So may be remove the first one at all?

alexkalmuk:
LGTM

anton-bondarev:
LGTM

src/drivers/block_dev/ramdisk/ramdisk.c
Deryugin: It's already done
https://github.com/embox/embox/commit/6cc1c27f369e1d9609e6c69f816f6de54541206c

Issue #255, opened 2011-03-01
Remove non C-library headers from root include


abusalimov:
C standard library headers are listed here: http://en.wikipedia.org/wiki/Category:C_standard_library_headers

These headers should be moved or dropped at all:

header          | destination
--------------- | -----------
bitops.h        | util/
crc32.h         | util/crypt
err.h           | drop
getopt.h        | compat/posix/include/ or rewrite into cmd/util.h
md5.h           | util/crypt
netutils.h      | net/misc.h or net/util.h
open_prom.h     | ???
queue.h         | util/queue.h or util/list.h
shell_command.h | embox/cmd.h
shell_utils.h   | drop
termios.h       | compat/posix/include/
time.h          | compat/posix/include/
types.h         | drop or move into embox/types.h
unistd.h        | compat/posix/include/

abusalimov:
@anton-bondarev Is this finally done?

PR #855, opened 2015-10-20
Fix issue 255

Deryugin:
Now there are no non-standart libc headers in src/compat/libc/include.
https://github.com/embox/embox/issues/255 will be closed.


anton-bondarev:
LGTM

PR #856, opened 2015-10-21
mk: initfs: Remove explicit commands for creating parent directory

abusalimov:
The necessary behavior is handled through `$(common_prereqs)` that contains a secondarily-expanded order-only `| $(@D)/.` prerequisite.

This should fix weird and floating bug reproduced during parallel builds, when (presumably) an initfs target rule was itself applied for creating its parent directory, inheriting the target-specific `$(src_file)`:

    cp -r -T $(src_file) $@    # $@ is $(ROOTFS_DIR)/dst_file

Which depends on (order-only) `| $(ROOTFS_DIR)/`.

    cp -r -T $(src_file) $@    # $@ is now $(ROOTFS_DIR)/.

Resulting in the following error:

    cp: cannot overwrite directory .../rootfs//. with non-directory

An indirect reason of this behavior is a bogus double-slash ('//') as a separator that made Make choose the `$(ROOTFS_DIR)/%` rule instead of the proper `$(ROOTFS_DIR)/.` one. This is fixed in build-gen.

### DIFF
--- a/mk/image2.mk
+++ b/mk/image2.mk
@@ -31,19 +31,15 @@ include $(__include_initfs)
 include $(__include)
 
 initfs_cp_prerequisites = $(src_file) $(common_prereqs)
-$(ROOTFS_DIR)/%/. : | $(ROOTFS_DIR)/.
-	@mkdir -p $(@D)
-	@touch $@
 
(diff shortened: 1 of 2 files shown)


anton-bondarev:
LGTM

PR #859, opened 2015-10-23
Implement umount for new VFS

Deryugin:
Umount was not very helpful, as it was a stub. Now it works fine with all bdev types, therefore ramdisk-related issue was closed.

Some VFS issues were fixed, e.g. double destroy of dentries, which led to inconsistency of dentries.

A number of fatfs issues were fixed. There were a bunch of typos and misusing sectors/clusters numbers. Now fatfs fully supports subdirectories, clears garbage on system volume data on disk format (previously FAT and root directory were not being erased).

### DIFF
--- a/src/compat/posix/fs/dvfs/fsop.c
+++ b/src/compat/posix/fs/dvfs/fsop.c
@@ -29,16 +29,23 @@ int mkdir(const char *pathname, mode_t mode) {
 		parent[strlen(parent) - 1] = '\0';
 
 	t = strrchr(parent, '/');
-	memset(t, '\0', parent + DENTRY_NAME_LEN - t);
+	if (t) {
+		memset(t, '\0', parent + DENTRY_NAME_LEN - t);
 
-	dvfs_lookup(parent, &lu);
-	if (!lu.item)
-		return SET_ERRNO(ENOENT);
+		dvfs_lookup(parent, &lu);
+		if (!lu.item)
+			return SET_ERRNO(ENOENT);
 
-	lu.parent = lu.item;
-	lu.item = NULL;
+		lu.parent = lu.item;
+		lu.item = NULL;
+	} else {
+		parent[0] = '\0';
+		dvfs_lookup(pathname, &lu);
+	}
 
-	return dvfs_create_new(pathname + strlen(parent), &lu, S_IFDIR);
+	return dvfs_create_new(pathname + strlen(parent),
+	                       &lu,
+			       S_IFDIR | (mode & DVFS_DIR_VIRTUAL));
 }
 
 int remove(const char *pathname) {
--- a/src/compat/posix/fs/dvfs/fsop.c
+++ b/src/compat/posix/fs/dvfs/fsop.c
@@ -24,7 +24,8 @@ int mkdir(const char *pathname, mode_t mode) {
 	if (lu.item)
 		return SET_ERRNO(EEXIST);
 
-	strncpy(parent, pathname, sizeof(parent));
+	parent[0] = '\0';
+	strncat(parent, pathname, sizeof(parent) - 1);
 	if (parent[strlen(parent) - 1] == '/')
 		parent[strlen(parent) - 1] = '\0';
 
--- a/src/compat/posix/fs/dvfs/fsop.c
+++ b/src/compat/posix/fs/dvfs/fsop.c
@@ -15,35 +15,30 @@
 
 int mkdir(const char *pathname, mode_t mode) {
 	struct lookup lu;
-	const char *t;
-	size_t namelen;
-	const char *dirname;
-	int i;
+	char *t;
+
+	char parent[DENTRY_NAME_LEN];
 
 	dvfs_lookup(pathname, &lu);
-
-	/* TODO check lu->parent == parent that we want */
 
 	if (lu.item)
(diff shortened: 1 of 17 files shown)


src/compat/posix/fs/dvfs/fsop.c
```diff
@@ -15,35 +15,37 @@
...
-	/* TODO rewrite this piece of code. The problem is optional trailing slash */
-	namelen = strlen(pathname);
-	dirname = t = pathname;
-	for (i = 0; i < namelen; i++) {
-		if (pathname[i] == '/') {
-			dirname = t;
-			t = &pathname[i];
-		}
-	}
+	strncpy(parent, pathname, sizeof(parent));
```
anton-bondarev: You should use sizeof()-1 size and set to zero last char in the string

src/fs/driver/dfs/dfs.c
```diff
@@ -594,12 +601,6 @@ static int dfs_mount_end(struct super_block *sb) {
 	inode->length    = MIN_FILE_SZ;
 	inode->start_pos = _capacity(sizeof(struct dfs_sb_info)) +
 		sbi->max_inode_count * _capacity(sizeof(struct dfs_dir_entry));
-
-	memset(buf, 0xff, sizeof(buf));
```
anton-bondarev: May be define special constant FLACH_EMPTY_BYTE 0xFF?

src/fs/driver/fat/fat_common.c
```diff
@@ -252,18 +273,18 @@ uint32_t fat_get_volinfo(void *bdev, struct volinfo * volinfo, uint32_t startsec
 	volinfo->numsecs =  (uint16_t) lbr->bpb.sectors_s_l |
 		  (((uint16_t) lbr->bpb.sectors_s_h) << 8);
 
-	if (!volinfo->numsecs) {
```
anton-bondarev: Is there not necessery block selection?

src/fs/driver/fat/fat_common.c
```diff
@@ -374,7 +390,7 @@ uint32_t fat_get_fat_(struct fat_fs_info *fsi,
 	 * always to read two FAT sectors, but that luxury is (by design intent)
 	 * unavailable to DOSFS.
 	 */
-	offset = ldiv(offset, volinfo->bytepersec).rem;
```
anton-bondarev: Can overflow occur here?

src/fs/driver/fat/fat_common.c
```diff
@@ -397,11 +413,10 @@ uint32_t fat_get_fat_(struct fat_fs_info *fsi,
 			result = (uint32_t) p_scratch[offset] |
 			  ((uint32_t) p_scratch[offset+1]) << 8;
 		}
-		if (cluster & 1) {
+		if (cluster & 1)
```
anton-bondarev: What this expretion mean?

src/fs/driver/fat/fat_common.c
```diff
@@ -940,10 +956,10 @@ int fat_root_dir_record(void *bdev) {
 		return -1;
 	}
 
-	cluster = fat_get_free_fat_(&fsi, fat_sector_buff);
+	cluster = 0; //fat_get_free_fat_(&fsi, fat_sector_buff);
```
anton-bondarev: Is this correct?

anton-bondarev:
Are you sure that it's safely don't use ldiv() in FAT driver?

src/fs/driver/fat/fat_common.c
Deryugin: It's not really neccessary, m.b. it will just make code more readable. Still ok for me.

src/compat/posix/fs/dvfs/fsop.c
Deryugin: Fixed

src/fs/driver/dfs/dfs.c
Deryugin: Done.

src/fs/driver/fat/fat_common.c
Deryugin: I don't think so, all the values are in 32-bit range.

src/fs/driver/fat/fat_common.c
Deryugin: Probably, that's related to fact, that for FAT12 size of cluster entry (12 bits) is not multiple of byte size (8 bits), so three bytes store info about 2 clusters (12 bit for each cluster, 8 bit per byte), and align is required.

Deryugin:
Yes, I'm pretty sure. At least the rest of out system doesn't use it, so if there's a problem with 32-bit operations, FAT would not be weakest link.

src/fs/driver/fat/fat_common.c
Deryugin: Probably, that's related to fact, that for FAT12 size of cluster entry (12 bits) is not multiple of byte size (8 bits), so three bytes store info about 2 clusters (12 bit for each cluster, 8 bit per byte), and align is required.

src/fs/driver/fat/fat_common.c
Deryugin: Fixed.

PR #862, opened 2015-10-28
Block dev open in devfs

anton-bondarev:
Enable 'dd' command in dvfs by adding idesc operations to block devices.
Clean block devices


alexkalmuk:
LGTM

PR #863, opened 2015-10-29
Enable access mode in dvfs

anton-bondarev:
Enable access mode during open.
Clean access mode flags (remove FS_MAY_XXX)
Now inode flags don't compare with open() flags just setup idesc amode flags from input open() flags


src/compat/posix/fs/dvfs/open.c
```diff
@@ -47,16 +46,32 @@ int open(const char *path, int __oflag, ...) {
...
 	it = task_resource_idesc_table(task_self());
 	res = idesc_table_add(it, idesc, 0);
 	if (res < 0) {
+		idesc->idesc_ops->close(idesc);
 		assert(0);
 		//TODO free resources
 		return SET_ERRNO(-res);
 	}
+
+	lookup.item->usage_count++;
```
Deryugin: Using "dentry_ref_inc(lookup->item)" is prefered.

src/compat/posix/include/fcntl.h
```diff
@@ -49,6 +49,7 @@ extern int fcntl(int fd, int cmd, ...);
 #define F_DUPFD            12
 
 /* fcntl flags */
+#define O_ACCESS_MASK      0x3
```
Deryugin: It's not a big problem, but I recommend to define it as (O_RDONLY | O_WRONLY | O_RDWR) instead.

Deryugin:
LGTM

PR #864, opened 2015-10-30
uart: Enable irq for stm32 uart

alexkalmuk:
Add uart's interrupt support on stm32f4discovery.

Note: only dynamical case, without STATIC_IRQ (it's not very easy to do)


anton-bondarev:
LGTM

PR #881, opened 2015-11-20
Ramdisk name

Deryugin:
Handle some ramdisk problems
* Use given name, not hdr#
* Store index and free resources properly on delete


anton-bondarev:
LGTM

PR #882, opened 2015-11-20
Statvfs dvfs fat

anton-bondarev:
Change mount sequence
Change fill_sb() routines in initfs and devfs drivers for dvfs
Improove 'df' command and statvfs() syscall


Deryugin:
LGTM

PR #885, opened 2015-12-10
File format

Deryugin:
Implement simple file format detection (right now uses simple check for RIFF file format)

### DIFF
--- a/src/fs/file_format.c
+++ b/src/fs/file_format.c
@@ -1,0 +1,26 @@
+/**
+ * @file
+ * @brief Detect file format by header/etc
+ * @author Denis Deryugin <deryugin.denis@gmail.com>
+ * @version 0.1
+ * @date 2015-12-10
+ */
+
+#include <fs/file_format.h>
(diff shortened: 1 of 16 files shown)


src/fs/file_format.c
```diff
@@ -0,0 +1,34 @@
...
+#include <unistd.h>
+
+enum fformat raw_get_file_format(uint8_t *file) {
+	/* Is it a RIFF file? */
+	if (!memcmp(file, "RIFF", 4)) {
+		/* TODO check file consistency */
+		return RIFF_FILE;
+	}
+
+	return TEXT_FILE; /* Regular text file */
```
anton-bondarev: May be, REGULAR_FILE? 
as for me text file is associated with not binary format file

anton-bondarev:
LGTM

src/fs/file_format.c
Deryugin: This would be solved as more file formats would be supported. It's not that important after all.

Issue #442, opened 2012-07-25
fioctl retrieves wrong file address for ramfs


abusalimov:
##### What steps will reproduce the problem?
1. In conf/rootfs/ create files 'foo' and 'bar' with some data in both
2. Try `fioctl(fopen(file_name, "r"), 0, &file_addr));` for both 'foo' and 'bar' files
3. Check the results in file_addr variable

##### What is the expected output? What do you see instead?
I see random values (often zeros) that may vary in case of changing linkage options (RAM start, etc.)
Cc: @mrgaz

abusalimov:
@anton-bondarev @mrgaz @Deryugin does anyone know whether this is fixed?

Deryugin:
This problem is caused by poorly written initfs_ioctl(). This function should be rewritten.

Deryugin:
Well, this is embarassing, but there is no standart function "fioctl" (at least in glibc). I'm not sure this function should be placed into "libc" module.

Anyway, what kind of output do you expect? What "0" request code means for regular files?

abusalimov:
@Deryugin Yep, it's true. [Here's](http://stackoverflow.com/questions/14210276/how-to-decode-cmd-3222823425-in-ioctl-in-linux-2-6-29/14211113#14211113) how the `ioctl` command argument is encoded. The codes are unspecified, generally speaking, and are OS/FS/device-specific.

I guess, in this particular case, `0` means that the FS in question supports only a single command, and whoever implemented it, was kinda lazy to provide a proper `#define` for a request code, hence 0 as a placeholder.

Deryugin:
Fixed

PR #898, opened 2016-01-25
md5sum fix

Deryugin:
Now use proper address in RAM (fioctl was buggy) + print warning if fails.
Also fix issue #442


anton-bondarev:
LGTM

PR #905, opened 2016-02-03
Dvfs chmod

alexkalmuk:
What's done:
* syscall chmod() is added for both of old and new fs's
* chmod cmd is posix now

It's an open question how to sync an inode's data/metadata with a storage during execution of chmod(), but currently we leave that as is.

### DIFF
--- a/src/compat/posix/fs/dvfs/chmod.c
+++ b/src/compat/posix/fs/dvfs/chmod.c
@@ -1,0 +1,28 @@
+/**
+ * @file
+ * @brief
+ *
+ * @date 2.02.2016
+ * @author Alex Kalmuk
+ */
+
+#include <sys/stat.h>
+#include <errno.h>
+#include <fs/dvfs.h>
+
+int chmod(const char *path, mode_t mode) {
+	struct lookup l;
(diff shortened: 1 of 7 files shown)


anton-bondarev:
LGTM

src/compat/posix/fs/dvfs/chmod.c
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @date 2.02.2016
+ * @author Alex Kalmuk
+ */
+
+#include <sys/stat.h>
+#include <errno.h>
+#include <fs/dvfs.h>
+
+int chmod(const char *path, mode_t mode) {
```
Deryugin: I think it's better to check correctness of given "mode" arg.
Also I don't see where information is being updated on the media.

alexkalmuk: > Also I don't see where information is being updated on the media. 

This is explained in PR's description - currently there is no sync with external storage. Let's suppose for now to use tmpfs or some similar thing.

Deryugin:
LGTM

PR #913, opened 2016-03-02
ARM callstack

Deryugin:
Implement stack unwind for ARM Procedure Call Standart frame format (-mapcs-frame option for gcc).
Now backtrace could be obtained correctly.


anton-bondarev:
How I can test it?
It's anought to include 'whereai' module?

anton-bondarev:
LGTM

PR #915, opened 2016-03-04
mem: Add debug version of pool

alexkalmuk:
This version of pool is safe from double freeing (catches double freeing with asserts)

There is a bitmap of free blocks, so every time when somebody calls pool_free() it's checked whether the specified object was previously allocated.


anton-bondarev:
LGTM

PR #921, opened 2016-03-14
ti816x remove skb extra

Deryugin:
Rewrite ti816x net driver.

skb extra moved to separate pool, double-linked lists replaces with descriptor-based lists.

Now it's the way more simple, so some problems should be fixed.


anton-bondarev:
LGTM

PR #925, opened 2016-03-24
skb rework

Deryugin:
Minor skb rework (skb_extra removed; data, skb and extra are splitted).

ti816x driver rewritten for less bugs and improved performance


anton-bondarev:
LGTM

PR #929, opened 2016-04-19
Update README.md

AntonKozlov:
ethtool is not required


anton-bondarev:
Are you sure, that 'ethtool' isn't required now?
Earlier it used for creating 'tun' interface

AntonKozlov:
I haven't found any references to ethtool except README. Seems that it is
`ip` job now, scripts/qemu/start_script

20 апр. 2016 г. 10:45 пользователь "Anton Bondarev" <

anton-bondarev:
LGTM
It's work without ethtool

PR #932, opened 2016-04-26
readv/writev realization

denis-chusovitin:
Add readv/writev + tests

### DIFF
--- a/src/compat/posix/fs/readv.c
+++ b/src/compat/posix/fs/readv.c
@@ -1,0 +1,49 @@
+#include <unistd.h>
+#include <errno.h>
+
+#include <fs/index_descriptor.h>
+#include <fs/idesc.h>
+#include <kernel/task/resource/idesc_table.h>
+
+#include <netinet/in.h>
+#include <sys/socket.h>
+
+#include <util/macro.h>
+#include <kernel/task/resource/idesc_table.h>
+#include <net/sock.h>
+#include <net/socket/socket_desc.h>
+#include <net/socket/ksocket.h>
+
+#define  socket_idesc_check(sockfd, sk) \
+	if (!idesc_index_valid(sockfd) || !index_descriptor_get(sockfd)) {\
+		return SET_ERRNO(EBADF);            \
+	}                                \
+	if (NULL == (sk = idesc_sock_get(sockfd))) { \
+		return SET_ERRNO(ENOTSOCK);            \
+	} \
+
+ssize_t readv(int fd, const struct iovec *iov, int iovcnt) {
+	int ret;
+	struct msghdr msg;
+	struct sock *sk;
+
+	if (iovcnt < 0) {
+		return SET_ERRNO(EINVAL);
+	}
+
+	socket_idesc_check(fd, sk);
+
+	msg.msg_name = NULL;
+	msg.msg_namelen = 0;
+	msg.msg_iov = (struct iovec*)iov;
+	msg.msg_iovlen = iovcnt;
+	msg.msg_flags = 0;
+
+	ret = krecvmsg(sk, &msg, sk->idesc.idesc_flags);
+
+	if (ret < 0) {
+		return SET_ERRNO(-ret);
+	}
+
+	return msg.msg_iovlen;
+}
--- a/src/compat/posix/fs/readv.c
+++ b/src/compat/posix/fs/readv.c
@@ -1,26 +1,12 @@
-#include <unistd.h>
-#include <errno.h>
+/**
+ * @file
+ * @brief
+ *
+ * @date 28.04.2016
+ * @author Denis Chusovitin
+ */
 
-#include <fs/index_descriptor.h>
-#include <fs/idesc.h>
-#include <kernel/task/resource/idesc_table.h>
-
-#include <netinet/in.h>
-#include <sys/socket.h>
-
-#include <util/macro.h>
-#include <kernel/task/resource/idesc_table.h>
-#include <net/sock.h>
-#include <net/socket/socket_desc.h>
-#include <net/socket/ksocket.h>
-
-#define  socket_idesc_check(sockfd, sk) \
-	if (!idesc_index_valid(sockfd) || !index_descriptor_get(sockfd)) {\
-		return SET_ERRNO(EBADF);            \
-	}                                \
-	if (NULL == (sk = idesc_sock_get(sockfd))) { \
-		return SET_ERRNO(ENOTSOCK);            \
-	} \
+#include "vec_common.h"
 
 ssize_t readv(int fd, const struct iovec *iov, int iovcnt) {
 	int ret;
--- a/src/compat/posix/fs/readv.c
+++ b/src/compat/posix/fs/readv.c
@@ -31,5 +31,5 @@ ssize_t readv(int fd, const struct iovec *iov, int iovcnt) {
 		return SET_ERRNO(-ret);
 	}
 
-	return msg.msg_iovlen;
+	return ret;
 }
--- a/src/compat/posix/fs/readv.c
+++ b/src/compat/posix/fs/readv.c
@@ -31,5 +31,5 @@ ssize_t readv(int fd, const struct iovec *iov, int iovcnt) {
 		return SET_ERRNO(-ret);
 	}
 
-	return ret;
+	return msg.msg_iovlen;
 }
--- a/src/compat/posix/include/unistd.h
+++ b/src/compat/posix/include/unistd.h
@@ -17,6 +17,8 @@
 #include <sys/stat.h>
 #include <sys/cdefs.h>
 
+#include <sys/uio.h>
+
 #include <posix_environ.h>
 
 /*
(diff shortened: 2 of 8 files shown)


src/compat/posix/fs/readv.c
```diff
@@ -0,0 +1,49 @@
...
+#include <netinet/in.h>
+#include <sys/socket.h>
+
+#include <util/macro.h>
+#include <kernel/task/resource/idesc_table.h>
+#include <net/sock.h>
+#include <net/socket/socket_desc.h>
+#include <net/socket/ksocket.h>
+
+#define  socket_idesc_check(sockfd, sk) \
```
AntonKozlov: How about to extract this repetative macro definition to header file?
I'd suppose to place it near these .c files and include like  `#include "vec_common.h"` or something like this

src/compat/posix/fs/readv.c
```diff
@@ -0,0 +1,49 @@
...
+	if (!idesc_index_valid(sockfd) || !index_descriptor_get(sockfd)) {\
+		return SET_ERRNO(EBADF);            \
+	}                                \
+	if (NULL == (sk = idesc_sock_get(sockfd))) { \
+		return SET_ERRNO(ENOTSOCK);            \
+	} \
```
AntonKozlov: extra backslash

src/compat/posix/include/unistd.h
```diff
@@ -17,6 +17,8 @@
 #include <sys/stat.h>
 #include <sys/cdefs.h>
 
+#include <sys/uio.h>
```
AntonKozlov: Have I missed something? cannot find this file in PR

src/net/socket/ksocket.c
```diff
@@ -230,7 +230,7 @@ int ksendmsg(struct sock *sk, struct msghdr *msg, int flags) {
 	assert(sk);
 	assert(msg);
 	assert(msg->msg_iov);
-	assert(msg->msg_iovlen == 1); // FIXME add support of scatter gather
+	//assert(msg->msg_iovlen == 1); // FIXME add support of scatter gather
```
AntonKozlov: Delete it without fear! =)

src/net/socket/udp_sock.c
```diff
@@ -74,12 +84,19 @@ static int udp_sendmsg(struct sock *sk, struct msghdr *msg, int flags) {
 
 	udp_build(skb->h.uh, sock_inet_get_src_port(sk), to->sin_port, total_len);
 
-	memcpy(skb->h.uh + 1, msg->msg_iov->iov_base, data_len);
+	size = sizeof(msg->msg_iov[i].iov_base[0]);
+	buf = buf_ = malloc(data_len);
```
AntonKozlov: use of malloc in kernel is forbidden. In general, kmalloc is suited for this.
But in this case things are different. sk_buff widely used in network stack. It is buffer, optimized to hold network data. It have pointers to data on different layers, allows easy placing in queues, ...
Here, skb obtained via call to make_pack, which takes desired buffer length and returns length of actually allocated buffer. There is no need to allocate intermediate buffer and place data in it.

A loop is really required here, if actual_len is smaller than total_len. Then, make_pack/snd_pack should   be repeated until no data left

src/compat/posix/include/unistd.h
denis-chusovitin: This file defines iovec, so I can add delcaration of readv/writev in unistd.h

AntonKozlov: I mean, there is no this file in PR
28 апр. 2016 г. 11:02 пользователь "Denis Chusovitin" <

denis-chusovitin: Probably, I misunderstood something, but this file is already in embox repository:
https://github.com/embox/embox/blob/master/src/compat/posix/include/sys/uio.h

AntonKozlov: Oh, sorry, you are rigth.

Malloc issue is the critical that prevent merging PR
28 апр. 2016 г. 16:14 пользователь "Denis Chusovitin" <

denis-chusovitin: I hope, malloc issue was fixed in the last commit:
https://github.com/embox/embox/pull/932/commits/3e4c0c35ad5d2b94e631073c93576e2ef6fbd818

src/compat/posix/fs/readv.c
```diff
@@ -0,0 +1,35 @@
...
+	msg.msg_iovlen = iovcnt;
+	msg.msg_flags = 0;
+
+	ret = krecvmsg(sk, &msg, sk->idesc.idesc_flags);
+
+	if (ret < 0) {
+		return SET_ERRNO(-ret);
+	}
+
+	return msg.msg_iovlen;
```
AntonKozlov: Return in bytes, not in vector count

src/net/socket/udp_sock.c
```diff
@@ -46,43 +49,65 @@ static int udp_sendmsg(struct sock *sk, struct msghdr *msg, int flags) {
...
+		if (ret != 0) {
+			return ret;
+		}
 
-	if (msg->msg_name != NULL) {
-		to = (const struct sockaddr_in *)msg->msg_name;
-	} else {
-		to = (const struct sockaddr_in *)&to_inet_sock(sk)->dst_in;
-	}
+		if (msg->msg_name != NULL) {
```
AntonKozlov: could be moved out of loop

src/net/socket/udp_sock.c
```diff
@@ -46,43 +49,65 @@ static int udp_sendmsg(struct sock *sk, struct msghdr *msg, int flags) {
...
+			to = (const struct sockaddr_in *)msg->msg_name;
+		} else {
+			to = (const struct sockaddr_in *)&to_inet_sock(sk)->dst_in;
+		}
+
+		assert(skb);
+		assert(skb->h.uh);
+
+		udp_header = (unsigned char *) (skb->h.uh + 1);
+		size = sizeof(msg->msg_iov[i].iov_base[0]);
```
AntonKozlov: isn't it `sizeof(void)`? surprisingly, this is not error, gcc allows this and produces `1`.
Plain C doesn't allow that, http://stackoverflow.com/a/1666255.

src/net/socket/udp_sock.c
```diff
@@ -46,43 +49,65 @@ static int udp_sendmsg(struct sock *sk, struct msghdr *msg, int flags) {
...
 
-	assert(skb);
-	assert(skb->h.uh);
+		iov_len = i = skb_len = 0;
 
-	udp_build(skb->h.uh, sock_inet_get_src_port(sk), to->sin_port, total_len);
+		for (i = 0; i < msg->msg_iovlen; i++) {
+			iov_len = msg->msg_iov[i].iov_len * size;
```
AntonKozlov: read/write (and readv/wirtev) operates on bytes level, in contrast with stdlib fread/fwrite. This line may confuse the reader, so I'd suggest just delete the size and don't scale here

src/net/socket/udp_sock.c
```diff
@@ -33,11 +35,12 @@ EMBOX_NET_SOCK(AF_INET, SOCK_DGRAM, IPPROTO_UDP, 1,
 		udp_sock_ops_struct);
 
 static int udp_sendmsg(struct sock *sk, struct msghdr *msg, int flags) {
-	int ret;
-	size_t data_len, total_len, actual_len;
+	size_t data_len, data_len_, actual_len, skb_len, iov_len;
```
AntonKozlov: How do you think, isn't 5 different lengths are too much for little function?

AntonKozlov: data_len and data_len_ are hardly to distinguish visually, how about renaming?

src/net/socket/udp_sock.c
```diff
@@ -46,43 +49,65 @@ static int udp_sendmsg(struct sock *sk, struct msghdr *msg, int flags) {
...
-	udp4_set_check_field(skb->h.uh, skb->nh.iph);
+		udp_build(skb->h.uh, sock_inet_get_src_port(sk), to->sin_port,
+				skb_len + UDP_HEADER_SIZE);
 
-	assert(sk->o_ops->snd_pack);
-	ret = sk->o_ops->snd_pack(skb);
-	if (0 > ret) {
-		return ret;
+		data_len_ -= skb_len;
+		actual_len = data_len_ + UDP_HEADER_SIZE;
```
AntonKozlov: actual_len initializaton required only before make_pack call, consider move it there

src/net/socket/udp_sock.c
```diff
@@ -46,43 +49,65 @@ static int udp_sendmsg(struct sock *sk, struct msghdr *msg, int flags) {
...
-	assert(skb->h.uh);
+		iov_len = i = skb_len = 0;
 
-	udp_build(skb->h.uh, sock_inet_get_src_port(sk), to->sin_port, total_len);
+		for (i = 0; i < msg->msg_iovlen; i++) {
+			iov_len = msg->msg_iov[i].iov_len * size;
 
-	memcpy(skb->h.uh + 1, msg->msg_iov->iov_base, data_len);
+			if (skb_len + iov_len <= actual_len) {
+				memcpy(udp_header + skb_len, msg->msg_iov[i].iov_base, iov_len);
```
AntonKozlov: Consider using `skb_iovec_buf`/`skb_buf_iovec`, they may be useful

AntonKozlov: May be you miss some sort of another utility functions like that?

anton-bondarev:
It was closed by #989

PR #934, opened 2016-04-30
Rework ttys subsystems

anton-bondarev:
Enable ttys0 device for all 'arch/qemu' templates except microblaze.
Rework ttys subsytem for oldfs


alexkalmuk:
LGTM
I have tested only for stm32f4, it works.

PR #936, opened 2016-05-05
Change loading process

anton-bondarev:
Add command 'service' starting comand as daemon (create own task for it)
Enable 'system_init' routine in kernel_init


Deryugin:
LGTM :shipit:

PR #944, opened 2016-05-13
Developed 'mipsnet' driver

anton-bondarev:
Extended mips/qemu template for common qemu template
Desing 'mipsnet' nic driver
Fix some bugs for big endians architects


alexkalmuk:
Checked on mips/qemu, it works fine.
LGTM

PR #945, opened 2016-05-29
Fix gcc-arm-embedde ppa and package name

AntonKozlov:
See https://launchpad.net/gcc-arm-embedded/+announcement/13824


anton-bondarev:
Thanks, LGTM

PR #946, opened 2016-06-01
Update ti816x net driver

anton-bondarev:
Refactor ti816 net driver


Deryugin:
Net driver still works for the board :)
LGTM :shipit:

PR #962, opened 2016-07-01
travis: Fix unquoted ${COVERITY_SCAN_BRANCH} test

abusalimov:
This fixes PR builds missing the actual `make` step due to a bash syntax error in case when the `${COVERITY_SCAN_BRANCH}` expands to empty:

    $ if [ ${COVERITY_SCAN_BRANCH} != 1 ]; then make && ./scripts/continuous/run.sh "$TEMPLATE"; fi
    /home/travis/build.sh: line 45: [: !=: unary operator expected

    The command "if [ ${COVERITY_SCAN_BRANCH} != 1 ]; then make && ./scripts/continuous/run.sh "$TEMPLATE"; fi" exited with 0.

The worst thing is that this error used to make PR builds effectively no-op, resulting in a green build without any check.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,5 +55,6 @@ addons:
 
 notifications:
   slack:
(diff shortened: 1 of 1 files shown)


anton-bondarev:
LGTM

PR #965, opened 2016-07-22
Fix build for x86/test/fs

Deryugin:
mmap() failed to compile in this template, this issue was caused by rewriting mmap() in #952.


anton-bondarev:
LGTM

PR #975, opened 2016-08-12
task: Make children of a died task daemons

alexkalmuk:
When a task died its children became orphans. To solve this problem we simply attach each orphan to the kernel task (like in unix).

### DIFF
--- a/src/kernel/task/multi.c
+++ b/src/kernel/task/multi.c
@@ -254,6 +254,18 @@ void task_init(struct task *tsk, int id, struct task *parent, const char *name,
 	task_resource_init(tsk);
 }
 
+static void task_make_children_daemons(struct task *task) {
+	struct task *child;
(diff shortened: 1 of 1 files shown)


src/kernel/task/multi.c
```diff
@@ -267,6 +279,10 @@ void task_do_exit(struct task *task, int status) {
 	/* Deinitialize all resources */
 	task_resource_deinit(task);
 
+	/* Make all children of the specified task daemons (or, more generally, orphans).
+	 * It is made by simply setting up the parent task of each child to kernel_task. */
+	task_make_children_daemons(task);
```
anton-bondarev: I think that damonization should be done during creating a new  task

alexkalmuk: Why?

anton-bondarev:
LGTM

PR #977, opened 2016-08-17
Fix vmem issue #968

### LINKED ISSUE #968: VMEM on IMX6
When you use commands its breake on an assert in do_map_region() 
`assert(!mmu_pte_present(pte + pte_idx)); `

Deryugin:
Working on ARM MMU to fix #968 
Along the way, a number of vmem-related issues were fixed.

### DIFF
--- a/src/arch/arm/cortexa8/mmu.c
+++ b/src/arch/arm/cortexa8/mmu.c
@@ -48,18 +48,6 @@ static int mmu_init(void) {
 	);
 
 	return 0;
-
-	/* Setup physical address of the first level translation table */
-	__asm__ __volatile__ (
-		"ldr r0, =translation_table\n\t"
-		"mcr p15, 0, r0, c2, c0, 0\n\t"
-#if 0
-		"mcr p15, 0, r0, c2, c0, 1"
-#endif
-		: :
-	);
-
-	return 0;
 }
 
 /**
@@ -113,6 +101,12 @@ mmu_ctx_t mmu_create_context(mmu_pgd_t *pgd) {
 }
 
 void mmu_set_context(mmu_ctx_t ctx) {
+	void *table = &translation_table;
+
+	__asm__ __volatile__ (
+		"mcr p15, 0, %[addr], c2, c0, 0\n\t"
+		: [addr] "=r" (table) :
+	);
 }
 
 /**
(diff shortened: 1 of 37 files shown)


src/arch/arm/cortexa8/mmu.c
```diff
@@ -48,18 +45,6 @@ static int mmu_init(void) {
 	);
 
 	return 0;
-
-	/* Setup physical address of the first level translation table */
-	__asm__ __volatile__ (
```
anton-bondarev: There isn't init for arm mmu?

Deryugin: There's setup of P15 register.
The physical address of the translation table is unique for each task and it's set on `mmu_set_context()`.

src/drivers/clock/integratorcp_clock.c
```diff
@@ -45,6 +47,15 @@
 EMBOX_UNIT_INIT(integratorcp_init);
 
 static int integratorcp_clock_setup(struct time_dev_conf * conf) {
+	if (NULL == mmap_device_memory(
+		(void*) TIMER_BASE,
+		0x28,
```
anton-bondarev: Why region size initialization is different for integrator_clock & cortexa9_timer

third-party/lib/linenoise/linenoise.c
```diff
@@ -662,8 +662,10 @@ static int linenoiseEdit(int fdin, int fdout, char *buf, size_t buflen, const ch
         switch(c) {
         case 10:    /* '\n' enter */
         case 13:    /* '\r' enter */
-            history_len--;
-            sysfree(history[history_len]);
+	    if (history_len) {
```
anton-bondarev: This is right branch for these changes

src/drivers/clock/integratorcp_clock.c
Deryugin: Thus size is just the highest register offset from periph base.

third-party/lib/linenoise/linenoise.c
Deryugin: In some templates there were no heap in config, so this variable was NULL. Without MMU it produced no errors, but with MMU on it led to exception

anton-bondarev:
LGTM

PR #979, opened 2016-08-25
ARM MMU small page mode

Deryugin:
Implement support for small page mode.
Previously, only section mode was available for Embox.


anton-bondarev:
LGTM

PR #992, opened 2016-09-22
Arm reworking

anton-bondarev:


src/arch/arm/armlib/entry.S
```diff
@@ -8,60 +8,80 @@
...
+	/* save temp register*/
+	stmfd  sp!, {r0}
+	/* storing LP */
+	stmfd  sp!, {lr}
+	/*storing SPSR and valuable regs*/
+	mrs    lr, SPSR
+	/* storing SPSR */
+	stmfd  sp!, {lr}
+	/* pointer to saved on special stack registers */
+	mov    r0, sp
```
AntonKozlov: After this stack is 
```

[r0]
[lr]
[SPSR] <- r0 points here
```
...

src/arch/arm/armlib/entry.S
```diff
@@ -8,60 +8,80 @@
...
-	ldr	r0, =this_spsr_lr
-	ldmia	r0, {r1, r2}
-	stmfd	sp!,{r0, r1, r2}
+	/* now we have previous mode and sp */
+	stmfd  sp!, {r1-r12, lr}
 
-	ldr	lr, =irq_leave
-	msr	CPSR, #ARM_MODE_IRQ | I_BIT | F_BIT
+	/* restore spsr, lr, tmp reg */
+	ldmfd  r0!, {r1} /* spsr */
```
AntonKozlov: This starts to load in decreasing addressed. After load stack will be 
```
[r0]
[lr]
[SPSR] 
... <- r0 points here
```

I suppose `ldmia r0, {r1, r2, r3}` should be used.

AntonKozlov: Ops, ldmfd is ldmia. So this line is correct

src/arch/arm/armlib/entry.S
anton-bondarev: I think this code is correct.
Why can 'r0' be changed?

PR #995, opened 2016-09-28
Clang minifixes

Kakadu:
Some warnings detected by clang compiler are fixed.

### DIFF
--- a/src/include/kernel/spinlock.h
+++ b/src/include/kernel/spinlock.h
@@ -33,7 +33,7 @@
 
 #ifdef SPIN_CONTENTION_LIMIT
 # define __SPIN_CONTENTION_FIELD      unsigned long contention_count;
-# define __SPIN_CONTENTION_FIELD_INIT .contention_count = SPIN_CONTENTION_LIMIT,
+# define __SPIN_CONTENTION_FIELD_INIT contention_count : SPIN_CONTENTION_LIMIT,
 #else /* SPIN_CONTENTION_LIMIT */
 # define __SPIN_CONTENTION_FIELD
 # define __SPIN_CONTENTION_FIELD_INIT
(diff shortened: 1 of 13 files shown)


src/include/kernel/spinlock.h
```diff
@@ -48,7 +48,7 @@ typedef struct {
 /* XXX use 'field : value' instead of '.field = value' syntax because g++ does not support
  * the second one, but supports the first one in the trivial order --Alexander */
 #define SPIN_INIT(state) \
-  { l : state, owner : -1u, __SPIN_CONTENTION_FIELD_INIT }
+  { .l = state, .owner = -1u, __SPIN_CONTENTION_FIELD_INIT }
```
anton-bondarev: This is required for C++ as described above

src/kernel/task/resource/sig_table.c
```diff
@@ -32,6 +32,15 @@ static void task_sig_handler_ignore(int sig) {
 #define Core task_sig_handler_terminate /* FIXME */
 #define Stop task_sig_handler_terminate /* FIXME */
 #define Cont task_sig_handler_ignore    /* FIXME */
+
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Winitializer-overrides"
```
anton-bondarev: This is very difficult can be make it easy?
If I right understand this problem, there are some syscall numbers with identical id for example 
```
#define SIGABRT     6       /* (ANSI)     Abort.  */
#define SIGIOT      6       /* (4.2 BSD)  IOT trap.  */
```
I think right way to fix this warning rewrite initialization array

Kakadu:
I think that I fixed issue about designator fields.

anton-bondarev:
@Kakadu Yes, thanks, Looks good!

PR #1012, opened 2016-12-23
Service as background task

anton-bondarev:
- Switched start system to service start
- Developed vfork for MIPS
- Clean telnetd

Now qemu templates for mips/arm/x86 starting 'httpd' and 'telnetd' by default as service for testing you should open default page and enter with telnet


src/cmds/net/telnetd/telnetd.c
```diff
@@ -48,23 +45,22 @@
...
+//TODO is not posix. Must be posix_openpt() and friends
 extern int ppty(int pptyfds[2]);
 
-static struct {
-	int fd;
-	struct sockaddr_in addr_in;
-} clients[TELNETD_MAX_CONNECTIONS];
-static int listening_descr;
+
+static int telnet_connections_count;
```
AntonKozlov: telned_connections_count modified from many threads simultaneously, race condition

Is the limit artificial or there are some implicit real limit?

src/init/system_start_service.c
```diff
@@ -22,12 +25,20 @@ static const char *script_commands[] = {
...
+	int argc;
+	const struct cmd *cmd;
 
 	setup_tty(OPTION_STRING_GET(tty_dev));
 
 	array_foreach(command, script_commands, ARRAY_SIZE(script_commands)) {
-
-		system(command);
+		argc = cmdline_tokenize((char *)command, argv);
+		if (0 == strncmp(argv[0], "pthread", 7)) {
```
AntonKozlov: only prefix checked, is this intentional? this matches `pthreadUnexpectedResult`

AntonKozlov: What's wrong with system as it was before? Could you comment it in code?

templates/mips/qemu/mods.config
```diff
@@ -70,8 +71,8 @@ configuration conf {
 	@Runlevel(1) include embox.test.util.hashtable_test
 
 	@Runlevel(2) include embox.cmd.sh.tish(prompt="%u@%h:%w%$", rich_prompt_support=1, builtin_commands="exit logout cd export mount umount")
-	@Runlevel(2) include embox.init.start_script(shell_name="tish", tty_dev="ttyS0", shell_start=1, stop_on_error=true)
-	//include embox.init.system_start_service
+	//@Runlevel(2) include embox.init.start_script(shell_name="tish", tty_dev="ttyS0", shell_start=1, stop_on_error=true)
```
AntonKozlov: Consider removing these lines instead of commenting out

src/cmds/net/telnetd/telnetd.c
anton-bondarev: Yes, it's possible but  it's not significant because there is incrementation only in one (accept thread) and we will not be able to open over connections

src/init/system_start_service.c
anton-bondarev: Ok, thanks. It's only for future

templates/mips/qemu/mods.config
anton-bondarev: I think it can be an example for switching modes.

src/init/system_start_service.c
anton-bondarev: After cmdline_tokenize argv[i] contains null terminated string and checked will be correct.
In the future we wants starting services as threads (not a task) for small devices but now I don't want design this code because it will require some harm dependences

src/cmds/net/telnetd/telnetd.c
AntonKozlov: As far as I understood, race condition could make telnet refuse connections. Why it's not significant? In some time system will be unreachable by telnet.
It also allows to bypass connection limit, accepting more connections than allowed. Since our usual configurations are strictly limited by number of threads/tasks, this could make the system to fail another thread/task creation.

src/init/system_start_service.c
AntonKozlov: Consider introducing a system call like linux-specific `clone` (for thread creation). Are the new syscall and `exec` followed solves the problem here?

anton-bondarev: May be, but I want to make them in static, for resource restrictions system.
For example robots in previous init mode (start_sctipt) we must set command as build_in in tish. Of cause tish must be enable and when it stating tish use the same stack as  running cmd (the only existing kernel stack). Now cmd can be call directly, of cause it save just a couple of bytes but in small systems it may be impotent/

PR #1020, opened 2017-01-08
Fix build & run on MAC OS X

sikmir:
Overview:
* (archlinux) no need to install `curl`, it's core package
* (macos) install prerequisites from MacPorts
  * crosstools: `i386-elf-gcc`, `i386-elf-binutils`
  * GNU version of tools: `cpio`(`gnucpio`) and `awk`(`gawk`)
* (qemu) fix auto_qemu script
  * Bash 4.0 or later is required (associative arrays support)
  * `kvm` is available only on Linux
  * fix qemu version detection

TODO: qemu networking needs to be fixed.


sikmir:
@AntonKozlov By the way, do you know a way to install all prerequisites via `brew` (for those who use it)? I just have no one.

AntonKozlov:
I've started from brew at first, cross-compiler was the main issue. But I stopped looking for native development environment on MacOSX because tun (required for qemu networking) is not officially supported by MacOSX kernel. Even brew/macports are rather intrusive for OS, requiring third-party kernel module appeared too much. 

I turned to virtualized environments (emdocker), which are relatively easy to setup and are isolated from the OS. Build speed is rather good and such environment allow us to provide developers environment that is known to be good (cross-compilers, emulators that we are using and tested with). Moreover, environment like that have nearly zero performance overhead on linux machines, so it could be used in everyday life by core developers. It should also be easy to provide environment for windows machines.

So, I don't know brew package names, but suppose they are straightforward to find from error log. I also encourage trying emdocker environment on macos/windows/linux, because it still young but should be able to resolve many questions with building/debugging on all these platforms.

sikmir:
Yes, sure, docker definitely is a better way.

PR #1023, opened 2017-01-10
Clang stm32f4cube support

Kakadu:
First of all it compiles and it doesn't break compilation with gcc.
Run `make confload-arm/stm32f4cube_clang all` to get the image.

There are some difficulties about supporting two different compilers.
Maybe I did it not in optiomal way but it works. The main issue that
command line switcher are different for gcc and clang (the latter uses
`-target` switch to specify a triplet). Also we can't compiler with clang
and _without_ gcc because clang is only a compiler and relies on cross
toolchain.

Also there some issues with clang codegeneration: generated object files
are bigger than with GCC. That's why some modules were commented out for
stm32f4cube.

### DIFF
--- a/mk/extbld/arch-embox-clang
+++ b/mk/extbld/arch-embox-clang
@@ -1,0 +1,1 @@
+arch-embox-gcc
--- a/mk/flags.mk
+++ b/mk/flags.mk
@@ -9,9 +9,6 @@ ARFLAGS ?=
 LDFLAGS ?=
 
 CROSS_COMPILE ?=
-
-CC      ?= $(CROSS_COMPILE)gcc
-CPP     ?= $(CC) -E
 CXX     ?= $(CROSS_COMPILE)g++
 AR      ?= $(CROSS_COMPILE)ar
 AS      ?= $(CROSS_COMPILE)as
@@ -20,6 +17,15 @@ NM      ?= $(CROSS_COMPILE)nm
 OBJDUMP ?= $(CROSS_COMPILE)objdump
 OBJCOPY ?= $(CROSS_COMPILE)objcopy
 SIZE    ?= $(CROSS_COMPILE)size
+
+ifeq ($(COMPILER),clang)
+CC      ?= clang
+# for clang LIBGCC_FINDER will be set externally to arm-none-eabi-gcc or something like that
+else
+CC      ?= $(CROSS_COMPILE)gcc
+LIBGCC_FINDER=$(CC) $(CFLAGS)
+endif
+CPP     ?= $(CC) -E
 
 comma_sep_list = $(subst $(\s),$(,),$(strip $1))
 
@@ -90,6 +96,7 @@ EXTERNAL_MAKE_FLAGS = \
 			CACHE_DIR, \
 		$(path_var)=$(abspath $($(path_var)))) \
 	BUILD_DIR=$(abspath $(mod_build_dir)) \
+	COMPILER=$(COMPILER) \
 	EMBOX_ARCH='$(ARCH)' \
 	EMBOX_CROSS_COMPILE='$(CROSS_COMPILE)' \
 	EMBOX_MAKEFLAGS='$(MAKEFLAGS)' \
@@ -177,19 +184,21 @@ override ASFLAGS += $(asflags)
 override COMMON_CCFLAGS := $(COMMON_FLAGS)
 override COMMON_CCFLAGS += -fno-strict-aliasing -fno-common
 override COMMON_CCFLAGS += -Wall -Werror
-override COMMON_CCFLAGS += -Wundef -Wno-trigraphs -Wno-char-subscripts 
+override COMMON_CCFLAGS += -Wundef -Wno-trigraphs -Wno-char-subscripts
 
 override COMMON_CCFLAGS += -Wno-gnu-designator
 
+ifneq ($(COMPILER),clang)
+# Not clang means gcc
 # This option conflicts with some third-party stuff, so we disable it.
 override COMMON_CCFLAGS += -Wno-misleading-indentation
 
-
 # GCC 6 seems to have many library functions declared as __nonnull__, like
-# fread, fwrite, fprintf, ...  Since accessing NULL in embox without MMU 
-# support could cause real damage to whole system in contrast with segfault of 
+# fread, fwrite, fprintf, ...  Since accessing NULL in embox without MMU
+# support could cause real damage to whole system in contrast with segfault of
 # application, we decided to keep explicit null checks and disable the warning.
 override COMMON_CCFLAGS += -Wno-nonnull-compare
+endif
 
 override COMMON_CCFLAGS += -Wformat
 
@@ -222,4 +231,3 @@ CCFLAGS ?=
 
 INCLUDES_FROM_FLAGS := \
 	$(patsubst -I%,%,$(filter -I%,$(CPPFLAGS) $(CXXFLAGS)))
-
--- a/third-party/lib/libgcc/Mybuild
+++ b/third-party/lib/libgcc/Mybuild
@@ -1,7 +1,7 @@
 package third_party.lib
 
 static module libgcc_toolchain extends embox.lib.LibGcc {
-	@Rule(script="$(CP) \"$$($(CC) $(CFLAGS) -print-file-name=libgcc.a | sed $$'s/\r$$//')\" $(OBJ_DIR)/third-party/lib/libgcc")
+	@Rule(script="$(CP) \"$$($(LIBGCC_FINDER) -print-file-name=libgcc.a | sed $$'s/\r$$//')\" $(OBJ_DIR)/third-party/lib/libgcc")
 	source "libgcc.a"
 	source "empty.c"
 }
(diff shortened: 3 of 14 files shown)


src/drivers/char/terminal/vtparse.c
```diff
@@ -46,10 +46,11 @@ static void do_action(struct vtparse *parser, vt_action_t action, char ch) {
 		break;
 
 	case VT_ACTION_COLLECT:
-		if (token->attrs_len == -1) {
```
anton-bondarev: It's not good idea. It will break some other templates.

src/net/util/show_packet.h
```diff
@@ -20,7 +20,7 @@ static inline void show_packet(uint8_t *raw, int size, char *title) {
 	if (!&mod_logger)
 		return;
 
-	if (&mod_logger.logging &&
+	if (//&mod_logger.logging &&
```
anton-bondarev: Why is it required?

templates/arm/stm32f4cube/mods.config
```diff
@@ -50,13 +50,13 @@ configuration conf {
 	include embox.init.setup_tty_diag
 	@Runlevel(3) include embox.init.start_script(shell_name="diag_shell")
 
-	include embox.cmd.fs.ls
+	include embox.cmd.fs.ls /*
```
anton-bondarev: It's reduced functionality for this template

src/drivers/char/terminal/vtparse.c
Kakadu: Do you want to say that in some other templates `attrs_len` is signed?

src/net/util/show_packet.h
Kakadu: It was a warning about redundant `&`. It seems that this code has no sense.

templates/arm/stm32f4cube/mods.config
Kakadu: See my remark about generated object files' sizes. I don't know another way to make template compilable with clang yet.

anton-bondarev: Yes. I understand. If You split these changes, we can investigate this problem too.

src/drivers/char/terminal/vtparse.c
anton-bondarev: No, I mean that this change (remove checking buffer overflow) can influence to another templates.
So, why is this required?

Kakadu:
Clang raises a warning. I decided to remove говнокод than to silence it.

src/drivers/char/terminal/vtparse.c
Kakadu: Let's move this discussion to  #1024

mk/extbld/arch-embox-clang
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
```
AntonKozlov: Is this a copy or a symlink? `arch-embox-g++` is already a symlink to `gcc`, couldn't `clang` be too? Or may be `arch-embox-gcc` could be used, even if it wraps clang?

mk/flags.mk
```diff
@@ -177,19 +184,33 @@ override ASFLAGS += $(asflags)
...
 # GCC 6 seems to have many library functions declared as __nonnull__, like
-# fread, fwrite, fprintf, ...  Since accessing NULL in embox without MMU 
-# support could cause real damage to whole system in contrast with segfault of 
+# fread, fwrite, fprintf, ...  Since accessing NULL in embox without MMU
+# support could cause real damage to whole system in contrast with segfault of
 # application, we decided to keep explicit null checks and disable the warning.
 override COMMON_CCFLAGS += -Wno-nonnull-compare
+else
+# happens in stm32_flash_cube.c
+override COMMON_CCFLAGS += -Wno-unused-const-variable
```
AntonKozlov: This is a good start, and probably should be deleted in future by incremental fixing the warnings mentioned.

However, I think we shouldn't add warning exceptions for code we can't control (stm cube). You can provide required warning exceptions in template.

templates/arm/stm32f4cube/build.conf
```diff
@@ -1,3 +1,8 @@
+include $(ROOT_DIR)/mk/core/compiler.mk
+
+# to see which target clang supports:
```
AntonKozlov: As this is a part of template, this checks supposed to replicate in all templates?
How about to create separate templates for clang?

templates/arm/stm32f4cube/mods.config
AntonKozlov: Separate template could resolve this too

mk/extbld/arch-embox-clang
Kakadu: I slightly changed the script. Maybe we can merge this back to `arch-embox-gcc` if we pass some env variable about clang and check for it. 

````
--- mk/extbld/arch-embox-gcc    2016-10-06 12:41:09.016126951 +0300
+++ mk/extbld/arch-embox-clang  2017-01-12 13:11:02.483323783 +0300
@@ -9,7 +9,7 @@
 
 cmd=$(basename $0)
 case $cmd in
-       *-gcc) C_CXX_FLAGS="$EMBOX_IMPORTED_CPPFLAGS $EMBOX_IMPORTED_CFLAGS";;
+       *-clang|*-gcc) C_CXX_FLAGS="$EMBOX_IMPORTED_CPPFLAGS $EMBOX_IMPORTED_CFLAGS";;
        *-g++) C_CXX_FLAGS="$EMBOX_IMPORTED_CPPFLAGS $EMBOX_IMPORTED_CXXFLAGS";;
        *)     echo "Unknown flags for $cmd"; exit 1;;
 esac
@@ -33,6 +33,6 @@
 
 ARG_LINE="$ARG_LINE $EMBOX_IMPORTED_CPPFLAGS"
 PWD_ARG_LINE="$(for i in $ARG_LINE; do echo ${i/$PWD/.}; done)"
-# echo "$EMBOX_CROSS_COMPILE${cmd#arch-embox-} $@ $PWD_ARG_LINE" >&2
-$EMBOX_CROSS_COMPILE${cmd#arch-embox-} "$@" $PWD_ARG_LINE
+# we hardcore target because we should rely on installed GCC toolchain
+clang "$@" -target arm-none-eabi $PWD_ARG_LINE
 exit $?
````

File: templates/arm/stm32f4cube/build.conf (hunk 6 above)
Comment (Kakadu) [inline templates/arm/stm32f4cube/build.conf]: Creating separate template did not come to my mind.... :-)

**Comment by Kakadu** [comment]
@AntonKozlov I added separate template.

File: mk/extbld/arch-embox-clang (hunk 4 above)
Comment (AntonKozlov) [inline mk/extbld/arch-embox-clang]: I'll be fine with variable passed. The first line can be used in new script, the compiler invocation should call clang or cross-gcc depending on $0

File: src/drivers/char/terminal/vtparse.c [hunk 7]
```diff
@@ -46,6 +46,7 @@ static void do_action(struct vtparse *parser, vt_action_t action, char ch) {
 		break;
 
 	case VT_ACTION_COLLECT:
+		// Buffer overflow check may be needed
```
AntonKozlov: This extra comment seems incorrect

src/net/util/show_packet.h
AntonKozlov: I'd suppose this and attr_len fix should be removed from this PR, they will be merged as #1024

src/drivers/char/terminal/vtparse.c
Kakadu: I need these changes to make template compilable. I'm going to do rebase later when all other issues will be resolved.

Kakadu:
Okay, any other complaints? @AntonKozlov  @anton-bondarev

anton-bondarev:
@Kakadu Please create a new PR with the only files:

- mk/core/compiler.mk
- mk/extbld.mk
- mk/extbld/arch-embox-clang
- mk/extbld/arch-embox-gcc
- mk/extbld/lib.mk
- mk/flags.mk
- mk/main-stripping.sh
- src/include/sys/cdefs.h

It very hard to review all changes in this PR
The best way is to put every change to separate commit

third-party/lib/libgcc/Mybuild
```diff
@@ -1,7 +1,7 @@
 package third_party.lib
 
 static module libgcc_toolchain extends embox.lib.LibGcc {
-	@Rule(script="$(CP) \"$$($(CC) $(CFLAGS) -print-file-name=libgcc.a | sed $$'s/\r$$//')\" $(OBJ_DIR)/third-party/lib/libgcc")
+	@Rule(script="$(CP) \"$$($(LIBGCC_FINDER) -print-file-name=libgcc.a | sed $$'s/\r$$//')\" $(OBJ_DIR)/third-party/lib/libgcc")
```
AntonKozlov: LIBGCC_FINDER defined in stm32_clang only, this will not work for other templates. As far as I understand, this line could be reverted?

AntonKozlov: Oh, my bad, libgcc_finder was introduced in #1025

anton-bondarev:
The changes was merged by #1027 
@Kakadu Thanks for your work!

PR #1024, opened 2017-01-11
Fix some redundant checks for non-NULL values raised up by clang

Kakadu:
Some warnings raised during compiling with clang for stm32f4cube. In `vtparse.c ` we compare unsigned value with -1. In another file we test that the pointer to global value is not null.

@anton-bondarev 

P.S. It is a part of #1023 for now.

````
src/drivers/char/terminal/vtparse.c:49:24: error: comparison of constant -1 with expression of type 'char' is always false
      [-Werror,-Wtautological-constant-out-of-range-compare]
                if (token->attrs_len == -1) {
                    ~~~~~~~~~~~~~~~~ ^  ~~

````

### DIFF
--- a/src/net/util/show_packet.h
+++ b/src/net/util/show_packet.h
@@ -20,7 +20,7 @@ static inline void show_packet(uint8_t *raw, int size, char *title) {
 	if (!&mod_logger)
 		return;
 
-	if (&mod_logger.logging &&
+	if (//&mod_logger.logging &&
 		mod_logger.logging.level >= LOG_DEBUG - 1) {
 
 		printk("\nPACKET(%d) %s:\n", size, title);
--- a/src/drivers/char/terminal/vt.h
+++ b/src/drivers/char/terminal/vt.h
@@ -41,7 +41,7 @@ struct vt_token {
 	vt_action_t action;
 	char        ch;
 	char        attrs[VT_TOKEN_ATTRS_MAX];
-	char        attrs_len;
+	int         attrs_len;
 	short      *params;
 	int         params_len;
 };

### CODE REVIEW

File: src/drivers/char/terminal/vtparse.c [hunk 1]
```diff
@@ -46,10 +46,11 @@ static void do_action(struct vtparse *parser, vt_action_t action, char ch) {
 		break;
 
 	case VT_ACTION_COLLECT:
-		if (token->attrs_len == -1) {
+		// Buffer overflow check may be needed
+		//if (token->attrs_len == -1) {
```
AntonKozlov: As far as I understood, attr overflow checked in other states (attrs buffer is VTPARSE_TOKEN_PARAMS_MAX (2) long). If it overflows, then len is assigned to `-1` and this as guard conditions to all states.
The problem seems to be that `attrs_len` is of `char` type, which 
a) could be signed or unsigned
b) is not what you'd expect for length type.

I'd suggest to revert this code and change attrs_len to `int` (probably reorganized vt_token structure to  get rid of unnecesary member padding)

@abusalimov could have something to say (as the author)

src/net/util/show_packet.h
```diff
@@ -20,7 +20,7 @@ static inline void show_packet(uint8_t *raw, int size, char *title) {
 	if (!&mod_logger)
 		return;
 
-	if (&mod_logger.logging &&
+	if (//&mod_logger.logging &&
```
AntonKozlov: This seems as inlined version of `logging_raw` function, which actually checks for this address. I'd suggest to totally remove commented line and replace also runtime check for `logging` in `logging_raw` with assert

We could also ask author (@vloginova) for better way

src/drivers/char/terminal/vtparse.c
Kakadu: @AntonKozlov I'm terribly confused by this. I thout that attr_len is unsigned but it seems that it have type `char`. But you can see how clang raises a warning above (I just added this).

anton-bondarev: @Kakadu May be in some platforms char can be unsigned? I offer to change type of **token->attrs_len** to **int** and remove comments

src/net/util/show_packet.h
AntonKozlov: Ok, let's leave this commented, but with comment why it's commented out

abusalimov: Checking for `&whatever` to be non-null is indeed redundant, I guess, this should be removed. AFAICS, this file, and these lines, in particular, were added by @Deryugin, Denis, could you elaborate on this, please?

AntonKozlov: Please consider also `logging_raw` function, which called like
```
#define log_logp(level, fmt, ...) \
        if (&mod_logger) \
                logging_raw(&mod_logger.logging, level, "%s: %s: " fmt "\n", \
```
and defined to be 
```
void logging_raw(struct logging *logging, int level, const char* fmt, ...) {
        if (logging
```

I find this definition and the redundant check at least consistent. 
In case the check deleted `logging_raw` should be altered similar way, I suppose

anton-bondarev:
Merged by #1029

PR #1029, opened 2017-01-18
Fix some redundant checks for non-NULL values raised up by clang

anton-bondarev:
This is re-based #1024 

@AntonKozlov commented log_level is not break any templates

### DIFF
--- a/src/net/util/show_packet.h
+++ b/src/net/util/show_packet.h
@@ -20,8 +20,7 @@ static inline void show_packet(uint8_t *raw, int size, char *title) {
 	if (!&mod_logger)
 		return;
 
-	if (&mod_logger.logging &&
-		mod_logger.logging.level >= LOG_DEBUG - 1) {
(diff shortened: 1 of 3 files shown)


src/net/util/show_packet.h
```diff
@@ -20,7 +20,7 @@ static inline void show_packet(uint8_t *raw, int size, char *title) {
 	if (!&mod_logger)
 		return;
 
-	if (&mod_logger.logging &&
+	if (//&mod_logger.logging &&
```
AntonKozlov: My concerns about this line https://github.com/embox/embox/pull/1024#discussion_r95756876 are still there. I'd prefer remove this and add assert in logging_raw

anton-bondarev:
@AntonKozlov I hope I resolved as You suggested.
If yes I will re-base  and merge

anton-bondarev:
This is re-based #1024

PR #1035, opened 2017-02-02
Nuklear porting

anton-bondarev:
Add ['nuklear'](https://github.com/vurtun/nuklear) gui library to third-party
fix debug & qt-app x86 templates
Update app_diff.sh

### DIFF
--- a/scripts/app_diff.sh
+++ b/scripts/app_diff.sh
@@ -18,3 +18,10 @@ mkdir -p $APP_NAME-orig
 tar xzf $EMBOX_DIR/download/${APP_NAME}.tar.gz -C $APP_NAME-orig --strip-components=1
 diff -aur -x configure $APP_NAME-orig $APP_NAME | grep -v "^Only in" | tee $SRC_APP_DIR/patch.txt
 rm -rf $APP_NAME-orig
+
+#ORIGIN_DIR=../$APP_NAME-orig
+#cd $EXTBLD_APP_DIR
+#mkdir -p $ORIGIN_DIR
+#unzip $EMBOX_DIR/download/${APP_NAME}.zip -d $ORIGIN_DIR
+#diff -aur -x configure $ORIGIN_DIR/$APP_NAME $APP_NAME | grep -v "^Only in" | tee $SRC_APP_DIR/patch.txt
+#rm -rf $ORIGIN_DIR
--- a/third-party/lib/nuklear/Makefile
+++ b/third-party/lib/nuklear/Makefile
@@ -1,0 +1,13 @@
+PKG_NAME := nuklear
+PKG_VER := ef2dcd3b779647e0140bb78863cb8439774e277b
+
+PKG_ARCHIVE_NAME :=$(PKG_NAME)-$(PKG_VER).zip
+PKG_SOURCES := https://github.com/vurtun/nuklear/archive/ef2dcd3b779647e0140bb78863cb8439774e277b.zip
+PKG_MD5     := 318736d16bc0f7f1808aba365826d6c0
+
+PKG_PATCHES := patch.txt
+
+include $(EXTBLD_LIB)
+
+$(INSTALL) :
+	cp $(PKG_SOURCE_DIR)/nuklear.h $(PKG_INSTALL_DIR)/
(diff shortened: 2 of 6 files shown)


third-party/lib/nuklear/patch.txt
```diff
@@ -0,0 +1,12 @@
+diff -aur -x configure ../nuklear-master-orig/nuklear-master/nuklear.h nuklear-master/nuklear.h
+--- ../nuklear-master-orig/nuklear-master/nuklear.h	2017-01-25 01:06:59.000000000 +0300
++++ nuklear-master/nuklear.h	2017-02-02 19:47:27.613793130 +0300
+@@ -272,7 +272,7 @@
```
sikmir: Changes of nuklear.h in upstream makes the patch inapplicable, since latest master snapshot is used.

third-party/lib/nuklear/Makefile
```diff
@@ -0,0 +1,13 @@
+PKG_NAME := nuklear
+PKG_VER := master
+
+PKG_ARCHIVE_NAME :=$(PKG_NAME)-$(PKG_VER).zip
+PKG_SOURCES := https://github.com/vurtun/nuklear/archive/master.zip
+PKG_MD5     := 8758316fd6114450627e15042c2371fc
```
sikmir: The same, master.zip is supposed to be changed.

scripts/app_diff.sh
```diff
@@ -16,5 +16,12 @@ SRC_APP_DIR=$EMBOX_DIR/$3 #third-party/fuse/fuse
 cd $EXTBLD_APP_DIR
 mkdir -p $APP_NAME-orig
 tar xzf $EMBOX_DIR/download/${APP_NAME}.tar.gz -C $APP_NAME-orig --strip-components=1
-diff -aur -x configure $APP_NAME-orig $APP_NAME | grep -v "^Only in" | tee $SRC_APP_DIR/patch.txt
-rm -rf $APP_NAME-orig
+diff -aur -x configure ../$APP_NAME-orig $APP_NAME | grep -v "^Only in" | tee $SRC_APP_DIR/patch.txt
```
sikmir: I think it's more correct to use `diff -Naur` here, without `grep -v "^Only in"`, isn't it?

anton-bondarev: Thanks. You are right. I fixed it.

third-party/lib/nuklear/patch.txt
anton-bondarev: Ops, You are right!
I suggested a temporary decision. I clone 'nuklear' to our repo and change link.
Now we have some troubles with google disk so I preferred  this variant.

sikmir: Why not just using fixed snapshot, e.g. https://github.com/vurtun/nuklear/archive/ef2dcd3b779647e0140bb78863cb8439774e277b.zip?

scripts/app_diff.sh
sikmir: What I mean is `diff -Naur -x configure ../$APP_NAME-orig $APP_NAME | tee $SRC_APP_DIR/patch.txt`. But as far as I can see you've just reverted changes by commit e6a534b834a563588098cf2ec7e9982393910011.

third-party/lib/nuklear/patch.txt
anton-bondarev: Ok, May be you're right. I did it.

sikmir:
By the way, I can't build `x86/qt-app` template to try out **nuklear**:
```
In file included from ../../include/QtCore/qlist.h:1:0,
                  from ../../include/QtCore/../../src/corelib/kernel/qobject.h:50,
                  from ../../include/QtCore/qobject.h:1,
                  from animation/qabstractanimation.h:45,
                  from animation/qabstractanimation.cpp:145:
 ../../include/QtCore/../../src/corelib/tools/qlist.h:54:20: fatal error: iterator: No such file or directory
  #include <iterator>
                     ^ compilation terminated.
make[7]: *** [Makefile:6438: .obj/debug-static/qabstractanimation.o] Error 1
```
gcc 6.3.1

anton-bondarev:
@sikmir For testing I commented Qt modules (Your troubles may be appear in gcc 6.3.1) and created a small command with nuklear calls like this
`#define NK_PRIVATE
#define NK_INCLUDE_FIXED_TYPES
#define NK_INCLUDE_STANDARD_IO
#define NK_INCLUDE_DEFAULT_ALLOCATOR
#define NK_INCLUDE_VERTEX_BUFFER_OUTPUT
#define NK_INCLUDE_FONT_BAKING
#define NK_INCLUDE_DEFAULT_FONT
#define NK_IMPLEMENTATION
#include "nuklear.h"

int main(int argc, char *argv[]) {
	struct nk_font_atlas atlas;
	int i, j;
	nk_font_atlas_init_default(&atlas);
	nk_font_atlas_begin(&atlas);
	struct nk_font *font = nk_font_atlas_add_from_file(&atlas, "Roboto-Black.ttf", 13, 0);
	const void* img = nk_font_atlas_bake(&atlas, &i, &j, NK_FONT_ATLAS_RGBA32);
...`

Next step must be a rendering support implementation for our 'framebuffer'

I suggest to create issue 'Qt build broken with gcc6.3.1'

PR #1040, opened 2017-02-15
Optimized build

Deryugin:
We had some problems with -O2 optimization as some variables could be used uninitialized in some cases. For some reason with -O0 these errors are not detected (even as warnings).

Now this problem is fixed for x86/qemu, arm/qemu, arm/overo.

Also fix minor dependency issues for arm/vfork.


src/mem/vmem/vmem_unmapper.c
```diff
@@ -119,7 +122,6 @@ void vmem_unmap_region(mmu_ctx_t ctx, mmu_vaddr_t virt_addr, size_t reg_size) {
 	}
 
 out_free:
-	try_free_pmd(pmd, pgd + pgd_idx);
```
anton-bondarev: Is it really required?

PR #1043, opened 2017-02-28
Fix stm32f4cube

alexkalmuk:
Fix stm32f4-cube -- clock was not initizalized. That was leading to some unpredicted behavior, for instance, network crash

### DIFF
--- a/src/drivers/interrupt/cortexm_nvic.c
+++ b/src/drivers/interrupt/cortexm_nvic.c
@@ -118,3 +118,17 @@ void irqctrl_force(unsigned int interrupt_nr) {
 	}
 }
 
+static void hnd_stub(void) {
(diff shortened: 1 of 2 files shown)


AntonKozlov:
Hi!

And what's the point of initialization of table in arch_init?

Beside that, I cannot see new dependency from arch.c containing module to interrupt_nvic module, that contains initialization. Am I missing something?

alexkalmuk:
Hi!
The point is some interrupts were generated during system clock initialization. More specifically, the problem is:
arch_init()
ipl_init()
<-- some interrupts are here, but the vector is not initialized
....
nvic_init()

AntonKozlov:
Got it, thanks!

PR #1049, opened 2017-03-03
Stm32f7discovery porting

alexkalmuk:
There is a porting on STM32746G-Discovery. One can try it by using the following template: `make confload-platform/stm32f7/stm32f7_discovery`

What's done:
* UART, system timer, i.e. minimal staff that is required to run Embox well
* two LCD examples. One of them draws yellow rectangle above the white one. Another example shows basic features of the stm cube LCD support like ascii symbols drawing


platform/stm32f7/cmds/stm32f7_lcd_test.c
```diff
@@ -0,0 +1,52 @@
+
```
anton-bondarev: May be create fb driver for stm32f7?

alexkalmuk:
I am agree, will do in the next PR. The reason is to make next branch shorter.

PR #1050, opened 2017-03-08
Workaround mips gcc bug of soft/hard float mixed

### LINKED ISSUE #949: mips/qemu: soft/hard-float ABI inconsistency
Here is a fragment from mips/qemu build log:

```
mips-elf-ld: Warning: build/base/obj/embox-2.o uses -msoft-float
  (set by ./build/base/obj/module/embox/arch/mips/kernel/interrupt.o),
  ./build/base/obj/module/embox/arch/mips/libarch.a(setjmp.o) uses -mhard-float
```

AntonKozlov:
Fixes #949

### DIFF
--- a/templates/mips/qemu/build.conf
+++ b/templates/mips/qemu/build.conf
@@ -7,8 +7,8 @@ CROSS_COMPILE = mips-elf-
 
 CFLAGS += -O0 -g -G0 -march=24kf -msoft-float -EB
 
-# Workaround for https://gcc.gnu.org/bugzilla/show_bug.cgi?id=64569
-# See also https://patchwork.linux-mips.org//patch/8355/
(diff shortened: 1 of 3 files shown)


templates/mips/qemu/build.conf
```diff
@@ -3,8 +3,12 @@ TARGET = embox
 
 ARCH = mips
 
-CROSS_COMPILE = mips-elf-
+CROSS_COMPILE = mips-linux-gnu-
```
AntonKozlov: Our "official" crosstool is mips-elf, so this template became broken out-of-box.
bare-metal toolchain should be mips-elf or modern mips-none-... mips-linux-gnu creates linux mips binaries, not bare-metal (standalone, freestanding)

PR #1051, opened 2017-03-10
Fix sparc and microblaze  templates build

anton-bondarev:


src/arch/sparc/lib/vfork.S
```diff
@@ -0,0 +1,23 @@
...
+	.align 4
+
+	.global vfork
+vfork:
+	SAVE_ALL
+
+	call vfork_body
+	 add %sp, STACK_FRAME_SZ, %o0
+
+	nop /* noreturn*/
```
AntonKozlov: May be replace with some sort of panic here? Or at least infinite loop?

AntonKozlov: still not resolved

anton-bondarev: This is only for remember:)
May be you are right of cause

anton-bondarev: Now SPARC is not work, i'm trying fix it

anton-bondarev:
@AntonKozlov 
I fixed sparc/qemu. Now it works with vfork_test.
Also please check microblaze/qemu I added telnetd & httpd servers by default

src/kernel/task/task_resource.c
```diff
@@ -98,11 +100,16 @@ static int task_resource_module_init(void) {
...
+
 		offset += binalign_bound(res->resource_size, sizeof(void *));
+		/* if we have aligned resource we have to reserve space for alignment*/
+		if(res->resource_align) {
+			offset += binalign_bound(res->resource_align, sizeof(void *));
+		}
 	}
 
-	assert(offset == binalign_bound(TASK_RESOURCE_SIZE, sizeof(void *)));
+	//assert(offset == binalign_bound(TASK_RESOURCE_SIZE, sizeof(void *)));
```
AntonKozlov: TASK_RESOURCE_SIZE is cumulative size of all resources, if it's not, it's a bug.
I think this issue is mandatory to fix.

src/arch/sparc/include/asm/ptrace.h
```diff
@@ -29,6 +29,12 @@ struct pt_regs {
 	uint32_t ins[8];     /**< Input registers. */
 }__attribute__ ((aligned (8)));
 
+#define PT_REGS_ALIGN    8
```
AntonKozlov: What exactly requires pt_regs to be aligned on 8?

anton-bondarev: Unfortunately yes!
Some architectures (SPARC for example) required stack alignment more than sizeof(void *) when we work with pt_regs we simulate work with stack

src/kernel/task/task_resource.c
anton-bondarev: Of couse it's mandatory, but if we have alignment this assertion rises
May be change the condition to '=>'?

src/arch/microblaze/kernel/irq_handler.c
```diff
@@ -25,20 +25,20 @@ void irq_handler(void) {
...
-				/* disable interrupt and clear it later, since ack
-					 * have no effect on level interrupt */
+				/* disable interrupt and clear it later, since ACK
+				 * have no effect on level interrupt */
 				irqctrl_disable(irq_num);
 
-				/*now we allow nested irq*/
-				ipl_enable();
+				/*now we allow nested IRQ*/
+				//ipl_enable();
```
anton-bondarev: I'm angry. I create an issue before this PR will merged if we don't resolve this problem

src/kernel/task/task_resource.c
AntonKozlov: TASK_RESOURCE_SIZE is a size of allocation to hold all resources of task. Only `<=` acceptable, and in `<` case it's just a waste of memory. In `>` case (now) it's a bug

src/arch/sparc/include/asm/ptrace.h
AntonKozlov: Are other implementations (that are possible on SPARC) require same alignment, or just this one? For example we could not simulate operating with stack.
IMO, In this particular case, concrete implementation should not make so big impact on high-level structure. Having alignment is useful, for performance reasons for example, but as far as I understand that is not the reason this alignment appeared.

anton-bondarev:
@AntonKozlov I removed high level logic. Please look at this, I'll rebase and squash commits after your approving

AntonKozlov:
Looks good!

PR #1056, opened 2017-03-21
Implement support for multiple UARTS on i.MX6

Deryugin:
Configure pins on iwave rainbow g15s to use UARTs on 84-pin extension.

### DIFF
--- a/src/drivers/serial/imx_uart/imx_uart.c
+++ b/src/drivers/serial/imx_uart/imx_uart.c
@@ -67,6 +67,7 @@ EMBOX_UNIT_INIT(uart_init);
 #define ONEMS		0xB0
 #define UTS		0xB4
 # define UTS_RST	(1 << 0)
+# define UTS_TXEMPTY	(1 << 6)
 #define UMCR		0xB8
 
 #include <kernel/printk.h>
@@ -155,7 +156,7 @@ static int imxuart_getc(struct uart *dev) {
 static int imxuart_putc(struct uart *dev, int ch) {
 	UART(TXR) = ch;
 
-	while (!(UART(UTS) & (1 << 6))) {
+	while (!(UART(UTS) & UTS_TXEMPTY)) {
 	}
 
 	return 0;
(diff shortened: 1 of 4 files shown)


src/drivers/serial/imx_uart/imx_uart.c
```diff
@@ -69,49 +153,49 @@ static int imxuart_getc(struct uart *dev) {
 }
 
 static int imxuart_putc(struct uart *dev, int ch) {
-	while (!(UART(USR2) & USR2_TXFE)) {
-	}
-
 	UART(TXR) = ch;
 
+	while (!(UART(UTS) & (1 << 6))) {
```
anton-bondarev: Magic constant?
Why are you wait free tx buffer?

src/drivers/serial/imx_uart/imx_uart.c
```diff
@@ -11,40 +11,124 @@
...
+	/* Reset */
+	UART(UCR2) = 0;
+	while(UART(UTS) & UTS_RST);
+
+	UART(UCR1) = UCR1_UARTEN;
+	UART(UCR2) = UCR2_NORESET | UCR2_RXEN | UCR2_TXEN | UCR2_WORDSIZE | UCR2_IRTS;
+	UART(UCR3) = UCR3_RXDMUXSEL | UCR3_RI | UCR3_DCD | UCR3_DSR;
+	UART(UCR4) = 0;
+	UART(UFCR) = 4 << UFCR_RFDIF;
+	UART(UBIR) = 0x0F;
```
anton-bondarev: Is it for a frequency. May be formula for calculating this fields?

src/drivers/serial/imx_uart/imx_uart.c
Deryugin: Fixed constant in the following commit.
Waiting is just to make sure symbol was sent, nothing special.

src/drivers/serial/imx_uart/imx_uart.c
Deryugin: This frequency is tied to reference clock, and I don't know how to configure it. So I suggest just to leave it as is.

Issue #36, opened 2010-03-29
remove comercial link from code


anton-bondarev:
Delete some reference to commercial project

Issue #32, opened 2010-03-26
Impelement ipl_interface in microblaze architecture


anton-bondarev:
ipl_xxx interface doesn't correspond ipl_xxx interface from HAL.

PR #1061, opened 2017-03-28
Fix cortexm irq handling

alexkalmuk:
There is a problem with interrupts handling on cortex-m. When we are inside an interrupt handler, another interrupts of the same priority (including the current one) are locked. The only way to enable it is to exit interrupt handle. One can check the issue using the test proposed in this PR and see that without this changes thread2 started only after thread1 finished.

This PR was tested on stm32f4-discovery with httpd and telnetd also

### DIFF
--- a/src/drivers/interrupt/cortexm_irq_handle.S
+++ b/src/drivers/interrupt/cortexm_irq_handle.S
@@ -17,15 +17,11 @@ interrupt_handle_enter:
     # due to ARMv7-M manual (see AAPCS 5.2.1.2)
     sub    sp, #64
     str    r0, [sp, #56]
-    stmia  sp, {r0 - r12}
-    str    lr, [sp, #52]
+    stmia  sp, {r0 - r12, lr}
 
     mov    r0, sp
     bl     interrupt_handle
 
-.text
-.thumb
-.syntax unified
 .global __irq_trampoline
 __irq_trampoline:
 
@@ -35,9 +31,6 @@ __irq_trampoline:
     ldr    r14, [r0, #36]
     bx     r14
 
-.text
-.thumb
-.syntax unified
 .global __pendsv_handle
 __pendsv_handle:
 
@@ -45,9 +38,6 @@ __pendsv_handle:
     add    sp, #32
     bx     r14
 
-.text
-.thumb
-.syntax unified
 .global __pending_handle
 __pending_handle:
 
--- a/src/drivers/interrupt/cortexm_irq_handle.S
+++ b/src/drivers/interrupt/cortexm_irq_handle.S
@@ -1,0 +1,67 @@
+/**
+ * @file
+ * @brief
+ *
+ * @author  Alex Kalmuk
+ * @date    27.03.2017
+ */
+
+.text
+.thumb
+.syntax unified
+.global interrupt_handle_enter
+interrupt_handle_enter:
+
+    mov    r0, sp
+    sub    sp, #58
+    str    r0, [sp, #56]
+    stmia  sp, {r0 - r12}
+    str    lr, [sp, #52]
+
+    mov    r0, sp
+    bl     interrupt_handle
+
+.text
+.thumb
+.syntax unified
+.global __irq_trampoline
+__irq_trampoline:
+
+    cpsid  i
+    # r0 is a pointer to struct irq_saved_state
+    ldr    sp,  [r0, #32]
+    ldr    r14, [r0, #36]
+    bx     r14
+
+.text
+.thumb
+.syntax unified
+.global __pendsv_handle
+__pendsv_handle:
+
+    # 32 == sizeof (struct cpu_saved_ctx)
+    add    sp, #32
+    bx     r14
+
+.text
+.thumb
+.syntax unified
+.global __pending_handle
+__pending_handle:
+
+    # push initial saved context (state.ctx) on top of the stack
+    add    r0, #32
+    ldmdb  r0, {r4 - r11}
+    push   {r4 - r11}
+
+    add    r1, r1, #52
+    ldmdb  r1, {r2 - r12}
+
+    cpsie  i
+    bl     critical_dispatch_pending
+    cpsid  i
+
+    bl     nvic_set_pendsv
+    cpsie  i
+    # DO NOT RETURN
+1: b       1
--- a/src/drivers/interrupt/cortexm_irq_handle.S
+++ b/src/drivers/interrupt/cortexm_irq_handle.S
@@ -13,7 +13,9 @@
 interrupt_handle_enter:
 
     mov    r0, sp
-    sub    sp, #58
+    # It should be a size of struct context, but we align it up to multiple of 8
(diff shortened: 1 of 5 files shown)


src/drivers/interrupt/cortexm_irq_handle.S
```diff
@@ -0,0 +1,67 @@
...
+.thumb
+.syntax unified
+.global __irq_trampoline
+__irq_trampoline:
+
+    cpsid  i
+    # r0 is a pointer to struct irq_saved_state
+    ldr    sp,  [r0, #32]
+    ldr    r14, [r0, #36]
+    bx     r14
```
AntonKozlov: This suppose to return from exception or to just call further? I can't understand the rest without understanding this.

(!)Anyway, sp is going to be restored at the state of exception (start of interrupt_handle_enter). Exceptional state, context *regs and irq_saved_state `state` could be overwritten

src/drivers/interrupt/cortexm_nvic.c
```diff
@@ -48,8 +55,58 @@ static uint32_t exception_table[EXCEPTION_TABLE_SZ] __attribute__ ((aligned (128
...
+ *         |
+ *         v
+ *         |--------> interrupt_handle (irq context)
+ *                        |
+ *         <--------------|
+ *         |
+ *         |
+ *         v
+ *
+ * This method:
```
AntonKozlov: Is this drawing correct? 
If it is and interrupt handle just calls another function, many of these could be rewritten in C.

alexkalmuk:
Thanks.

The main idea:
* Roughly speaking, interrupt_handle() is called when an interrupt is generated. In this function we handle the current interrupt at first, nothing special. Then...
* Then we want to call critical_dispatch_pending(). To do that we modify CPU-stored context on the stack for "returning" to __pending_handle(), not to the place we were interrupted at. This "returning" is made by (simplified):
Store the current context from the stack somewhere:
`memcpy(&state.ctx, ctx, sizeof *ctx);`
Now we can modify it:
`ctx->lr = (uint32_t) __pending_handle`
`ctx->pc = ctx->lr;`
Well, after this preparation, __irq_trampoline simply makes usual "returning" from the interrupt. But now we are inside __pending_handle!
* Ok, we are here, in __pending_handle. Now we are doing critical_dispatch_pending, but already in non-irq context. It is the point this magic is done for.
* Well, critical_dispatch_pending is done, what else? Let's recall that we need to finish interrupt handling by returning to the place we were interrupted at. But to do this, we need to return back to interrupt somehow and simply returning like nothing happened :) To do this we need to restore CPU-saved interrupt context at first. 
`# push initial saved context (state.ctx) on top of the stack`
`add    r0, #32`
`ldmdb  r0, {r4 - r11}`
`push   {r4 - r11}`
Well, now we have the same CPU-stored context like on the first interrupt entering. The last thing we would like to do is to generate PendSV interrupt.
`bl     nvic_set_pendsv`
After that __pendsv_handle() will be called as an PendSV handler, and we simply return from the interrupt: 
`__pendsv_handle:`
`# 32 == sizeof (struct cpu_saved_ctx)`
`add    sp, #32`
`bx     r14`

src/drivers/interrupt/cortexm_irq_handle.S
```diff
@@ -0,0 +1,67 @@
+/**
```
alexkalmuk: The only reason for doing that is to keep this file along with cortexm_nvic.c, because it has only supporting magic for interrupt_handle realization and no its own separated logic.

src/drivers/interrupt/cortexm_irq_handle.S
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+.text
+.thumb
+.syntax unified
+.global interrupt_handle_enter
+interrupt_handle_enter:
+
+    mov    r0, sp
+    sub    sp, #58
```
alexkalmuk: Yes, I forgot about this.

src/drivers/interrupt/cortexm_irq_handle.S
alexkalmuk: Returning from exception.

Yeah, it is a cheating to keep some stuff behind the stack pointer for a while. But we have disabled interrupts previously:
`cpsid  i`
`# r0 is a pointer to struct irq_saved_state`
`ldr    sp,  [r0, #32]`
...

src/drivers/interrupt/cortexm_nvic.c
alexkalmuk: Yes, I think so. But may be not very representative.
In my opinion, it is really hard to keep this code is C, because of compiler-specifiec push's on function entering, cheating with sp, and some other issues. I believe that it is not very big bunch of functions in asm.

src/drivers/interrupt/cortexm_nvic.c
```diff
@@ -71,8 +147,9 @@ static int nvic_init(void) {
 	void *ptr;
 
 	for (i = 0; i < EXCEPTION_TABLE_SZ; i++) {
-		exception_table[i] = ((int) interrupt_handle) | 1;
+		exception_table[i] = ((int) interrupt_handle_enter) | 1;
 	}
+	exception_table[14] = ((int) __pendsv_handle) | 1;
```
alexkalmuk: Yes, thank you

src/drivers/interrupt/cortexm_irq_handle.S
AntonKozlov: Aha, got it. I think both statements deserve comments in code.

I have a doubt `bx r14` have a meaning of interrupt return. "B1.5.8 Exception return behavior" states:
```
Exception returns occur when one of the following instructions loads a value of 0xFXXXXXXX into the PC while in Handler mode ...
```
R14 holds __pending_handle address here? Then it should be just a jump.

alexkalmuk: Sure, I will comment these two things.

> R14 holds __pending_handle address here? 

No, R14 holds 0xFFFFFFXX value here. Look, it was restored one line above `ldr    r14, [r0, #36]` . That is, `[r0, #36]` points to the initial value, which was loaded in R14 by CPU on irq handler entering. The actual address of __pending_handle was already stored on the stack in cortexm_nvic.c:
`ctx->lr = (uint32_t) __pending_handle;`

AntonKozlov: Got it, was confused by ctx->lr asignment just above. Ok, looks correct.
If you don't mind, I'd recommend to change __irq_trampoline interface: 2nd arg not used, only sp and restore lr used from 1st. If you just pass return sp and lr, then __irq_trampoline could be just 
```
mov sp, r0
bx r1
```

alexkalmuk: Yes, looks like a good way to simplification.

alexkalmuk:
@AntonKozlov I made changes due to your remarks. Please take a look when you have time.

AntonKozlov:
Looks good! (and a bit clearer for me =))

PR #1069, opened 2017-04-08
Fix ip fragmentation

anton-bondarev:


alexkalmuk:
I checked the fix on stm32f4-discovery in the following way:
At first I send a big ICMP packet of about 8192 bytes, so there it no more space available for new packets. But after 4 seconds, this fragments are deleted, and embox resumes running is the usual manner.
LGTM

PR #1071, opened 2017-05-01
Autoformatting codebase

AwesomeLemon:
For some reason on github some empty lines at files' ends disappear: I can see them locally, but not here.
Another problem: in empty whiles there's space before semicolon, so it looks like this: "while (true) ;"

PR #1072, opened 2017-05-05
Uncrustify

AwesomeLemon:
Disclaimer: this pull request is not in its final state. Its purpose is to provide opportunity to review and discuss code changes and code style changes.

Following in the steps of https://github.com/embox/embox/pull/1071

* empty while still has space before semicolon
* there seems to be no options about 'small' structs (this is the answer to a question from the previous pull request)
* I don't know why comment is moved in platform/lego_nxt/drivers/lcd/display.c : 21. Haven't found option that may change it
* for now I disabled column limit for easier discussion of other changes

Guide on playing around with uncrustify:
* config file is uncrustify_cfg.ini
* to format all files, start format_files.sh: the only argument is a file with list of files to process
	* if you don't want to format inplace, comment 3 last lines in format_files.sh


platform/stm32f4_cnc/http_admin/http_admin_backend.c
```diff
@@ -63,30 +63,30 @@ static char *http_admin_build_iface_list(void) {
...
-        }
+	}
 
 	json_list = cJSON_PrintUnformatted(iface_array);
 
 	cJSON_Delete(iface_array);
 	freeifaddrs(ifa);
 	return json_list;
-outerr:
+	outerr:
```
AntonKozlov: labels should not be indented (something desired is https://www.kernel.org/doc/html/v4.10/process/coding-style.html#centralized-exiting-of-functions)

src/arch/arm/armlib/exception_handler.c
```diff
@@ -10,14 +10,14 @@
...
-		"r4=%08x r5=%08x r6=%08x r7=%08x\n"
-		"r8=%08x r9=%08x r10=%08x r11=%08x\n"
-		"r12=%08x r14=%08x\n"
-		"cpsr=%08x spsr=%08x\n",
-		regs[2], regs[3], regs[4], regs[5],
-		regs[6], regs[7], regs[8], regs[9],
-		regs[10], regs[11], regs[12], regs[13],
-		regs[14], regs[15],
-	        regs[0], regs[1]);
+		  "r0=%08x r1=%08x r2=%08x r3=%08x\n"
```
AntonKozlov: This change is a mystery for me. I see nothing wrong with 2 level indentation everywhere, but there are 2 extra spaces and 1 indentation level below

src/cmds/module/test.c
```diff
@@ -53,7 +53,9 @@ int main(int argc, char **argv) {
 	int test_nr = -1;
 	int opt;
 	/* TODO it must be agreed with shell maximum command length */
-	char test_name[100] = { 0 };
+	char test_name[100] = {
```
AntonKozlov: This looks ugly. User should not be forced to change prior variation with this.
Are you sure there is no option to disable formatting structs (may be at all?)

src/cmds/module/test.c
```diff
@@ -75,7 +77,7 @@ int main(int argc, char **argv) {
 		case '?':
 		case 'h':
 			print_usage();
-			/* FALLTHROUGH */
+		/* FALLTHROUGH */
```
AntonKozlov: wrong indent

src/kernel/lthread/lthread_sched_wait.c
```diff
@@ -90,7 +90,7 @@ int sched_wait_timeout_lthread(struct lthread *self, clock_t *remain) {
 	}
 
 	if ((res = timer_set(&info->tmr, TIMER_ONESHOT, jiffies2ms(info->remain),
-			sched_wait_timeout_handler, &self->schedee))) {
+					sched_wait_timeout_handler, &self->schedee))) {
```
AntonKozlov: wrong indent change

src/kernel/lthread/sync/mutex.h
```diff
@@ -27,15 +27,15 @@
...
+		struct mutex *mutex) {
 	return WAITQ_WAIT_LTHREAD(self, &mutex->wq, ({
-			int done;
-			done = (mutex_trylock_schedee(&self->schedee, mutex) == 0);
-			if (!done) {
-				mutex_priority_inherit(&self->schedee, mutex);
-			}
-			done;
-		}));
+				int done;
```
AntonKozlov: wrong indent again

src/kernel/thread/fixed_stack.h
```diff
@@ -12,7 +12,7 @@
 
 struct thread_stack {
 	void              *stack;         /**< Allocated thread stack buffer. */
-	size_t             stack_sz;      /**< Stack size. */
+	size_t stack_sz;                  /**< Stack size. */
```
AntonKozlov: The former version have members aligned (probably with some extra indentation), but new version is worse

AntonKozlov:
Oh, and could update be made in this PR? I suppose there will be a number of iteration, but it won't be good to have same number of pull requests.

AwesomeLemon:
Sure, I'll be working in this PR from now on. I've created a new one because of my problems with git, github, and branches, which I've discussed with you. Now all seems fine.

src/kernel/lthread/sync/mutex.h
AwesomeLemon: Well, technically, it's a continuation of a parameter, so we indent it with two tabs, so that's kind of logical. I'll look into whether this behaviour can be changed.

src/kernel/thread/fixed_stack.h
AwesomeLemon: There're a lot of options for alignment, but I haven't worked with them yet.

src/kernel/lthread/sync/mutex.h
AwesomeLemon: Wait, it isn't wrong ident at all: we continue code block with ident of two tabs - because it's our desired indent. And then there's indent inside code block, because it's a code block - so we have an indent of one tab. It's also exactly what we want in new blocks of code. So it seems to me that this behaviour is desired.

src/arch/arm/armlib/exception_handler.c
AntonKozlov: Ok, I see. Can the option be disabled?
And is there an explanation why `regs[2], regs[3], ...` get indented by level 3?

AwesomeLemon: Well, 'panic' is indented by 1 level, arguments of 'panic' are indented by continuation indent, which is 2. That gives 3 levels total. I think that it's an expected behavior, am I wrong?

AntonKozlov:
Hmm, seems legit. I'm doing crosscheck with clang-fromat, and it gives same
output. While I'm not happy with this output (perhaps there is an issue
with definition of continuation block) let's say it's ok for

AwesomeLemon:
Do we really want to have 2 tabs as continuation indent? It looks weird, and isn't widely used in the existing code.

AntonKozlov:
>Do we really want to have 2 tabs as continuation indent?

The main reason is function defininitions with long argument lists, like
```
void foo(int a1, int a2, ...
        int a16, int a17, ...) { 
    int local; 
    if (local = bar(a1, a2, a3, ...
            a16, a17) {
        int local2;
        ...
    }
}
```

With 1 level indent it will be harder to distinguish continuation vs block content
```
void foo(int a1, int a2, ...
    int a16, int a17, ...) { 
    int local; 
    if (local = bar(a1, a2, a3, ...
        a16, a17) {
        int local2;
        ...
    }
}
```

Another option is to align parameters, like

```
void foo(int a1, int a2, ...
         int a16, int a17, ...) {
    ...
}
```
but it will be harder to maintain, i.e. after function rename.

src/kernel/lthread/lthread_sched_wait.h
```diff
@@ -25,9 +25,9 @@ struct sys_timer;
  */
 struct sched_wait_info {
 	struct sys_timer *tmr;
-	clock_t remain;
-	clock_t last_sync;
-	int status;
+	clock_t			  remain;
```
AntonKozlov: Tab used to align. Tabs should be used only to indentation, alignment should be only with spaces, that will make good look with every tab length in most cases

src/cmds/module/test.c
AwesomeLemon: I understand that this is struct initialization? In this case there's current feature request in uncrustify issues: https://github.com/uncrustify/uncrustify/issues/1200
The problem is that newlines are added after every '{' and there's no way (yet) to disable it in this specific situation.

scripts/codestyle/format_files.sh
```diff
@@ -0,0 +1,26 @@
...
+  mkdir -p out/$dn
+  ~/uncrustify_git/uncrustify/build/uncrustify -f $item -c uncrustify_cfg.ini > out/$item
+  #uncrustify -f $item -c uncrustify_cfg.ini > out/$item
+  #mv -f out/$item $item
+
+done
+
+find out -type f -empty -delete
+rsync -ah out/ ./
+rm -rf out
```
AntonKozlov: There is no newline at the end of file here and other files

src/tst2.c
```diff
@@ -0,0 +1,99 @@
+
```
AntonKozlov: There are 2 test files, that should be deleted prior merging.
However, they show how UNCRUSTIFY job will fail =)

scripts/codestyle/run_uncrustify.sh
```diff
@@ -0,0 +1,69 @@
...
+
+rm -rf out
+
+if [ -s uncrustify_diff.txt ]
+then
+	echo "The differences that don't comply with the code style are written to uncrustify_diff.txt"
+	if $VERBOSE; then
+		cat uncrustify_diff.txt
+	fi
+	exit 1
```
AntonKozlov: Can we just list files with suggested changes in not-verbose mode? 
In the end of script, uncrustify_diff.txt should be deleted

src/tst2.c
AwesomeLemon: I don't think that they should be deleted: if someone were to continue this taks, they would prove helpful.

scripts/codestyle/run_uncrustify.sh
```diff
@@ -0,0 +1,69 @@
...
+	    VERBOSE=true
+	    ;;
+	    *)
+	    HASHES+=($key)
+	    ;;
+	esac
+	shift
+done
+
+if [ ${#HASHES[@]} -eq 2 ];
```
AntonKozlov: Could you specifiy usage in help message or comment.
Is one hash mode supported (semantic: from that commit to current)
Two hash mode is doubtfully needed from my point, are you find it useful?

src/tst2.c
AntonKozlov: Let's move them to scripts/codestyle/testcases/ then?

scripts/codestyle/run_uncrustify.sh
AntonKozlov: Thanks!
Could you change help format to more formal one? Like (`man git-diff`):
``` 
      git diff [options] [<commit>] [--] [<path>...]
      git diff [options] [--no-index] [--] <path> <path>
```
where optional parts are in square brackets? I suppose
```
     $0 [<SHA1> <SHA2>]
``` 
should work?

scripts/codestyle/travis_uncrustify.sh
```diff
@@ -0,0 +1,7 @@
+echo $PWD
+if ! $TRAVIS_PULL_REQUEST; then
+	bash ./run_uncrustify.sh -v $(git rev-parse HEAD~1) $(git rev-parse HEAD)
```
AntonKozlov: we need to filter out codestyle/testcases/ path while codestyle check there fails (so at least this PR will not fail). Let's check `src` only for a while

AwesomeLemon: Modifying script to check only ``src`` would look ugly. Maybe a better option would be to have a file with names of files we want to ignore?

AntonKozlov: May be, or just a list with path to scan as a variable in script.
I think it should be include list (src), not exclude list (third-party, scripts,...)

AwesomeLemon: Okay, I've done it.

AwesomeLemon:
Sure. #1094 looks fine, I think it can be merged.

scripts/codestyle/run_uncrustify.sh
```diff
@@ -0,0 +1,106 @@
...
+then 
+	diffargs="${HASHES[0]} ${HASHES[1]}"
+fi
+
+files=$(git diff --name-only $diffargs);
+#files=$(find ../../ -type f -name "*.[ch]") #all files - I'll leave it just in case
```
AntonKozlov: Could you remove commented lines?

scripts/codestyle/run_uncrustify.sh
```diff
@@ -0,0 +1,106 @@
...
+
+folders_to_format=$(cat folders_to_format.txt)
```
AntonKozlov: Is the script intended to run from scripts/codestyle? It's not very convinient, could you make it run from arbitrary current directory? Useful way is to define something like `UNCRUSTIFY_BASE=$(dirname $0)` and use it as a path to folders_to_format.txt, etc

This also will allow to remove explicit processing of path (`../..`) on line 65

AwesomeLemon: I don't see how it'll help with removal of ``../../`` on line 65: suppose we're in ``src`` and running uncrustify like this: ``../scripts/codestyle/run_uncrustify.sh``. Then ``UNCRUSTIFY_BASE == ../scripts/codestyle``. From ``git diff `` we get, say, ``src/fs/smth.c``.
Now to correctly find file we need somehow to get the path ``fs/smth.c`` (or ``../src/fs/smth.c``), and I don't see how ``UNCRUSTIFY_BASE`` can help with that.

AwesomeLemon: For now I've implemented all other changes that you've requested.
As for this request, I've added to the beginning of the file ``cd $UNCRUSTIFY_BASE``, and to the end ``cd -``, which I think solves any problems.

AntonKozlov: Thanks, it's better now. 

Regarding `cd`, it's is error-prone, and seems like there is an error. When I change `src/kernel/init.c`, run `./scripts/codestyle/run_uncrustify.sh` (from top directory), then `scripts/src/kernel/init.c` appear. After some debugging, seems like `$item` is `../../src/kernel/init.c` and temporary files created in `out/../../src/.....`

Also, output changed to 
```
Some files don't comply with code style:
../../src/kernel/init.c
```
I'd expect `src/kernel/init.c` in output

AwesomeLemon: I think I've fixed that bug (and made output better).

AntonKozlov:
Seems good now. I've rebased this again in #1094 and merging it. This one is closed and remain as historical reference. Further work should be done in new branches/PRs. 
Thanks for your work!

PR #1079, opened 2017-06-09
stm32f7: Add sdram_init as a module for stm32f7discovery

anton-bondarev:


Deryugin:
@anton-bondarev do we have any examples of using this SDRAM?

PR #1080, opened 2017-06-09
Fix heap

anton-bondarev:
Add special implementation of heap where using fixed address and it's not presented in ld script


Deryugin:
@anton-bondarev may be we should add template with fixed heap? I guess this branch is related with #1079

PR #1085, opened 2017-06-19
mmap() child exit test

anna-marina:


anton-bondarev:
@anna-marina Thanks for your PR!

Can you rework to use only POSIX callings vfork() + exec()?
Anyway you should fix building. You forgot to include some headers.

AntonKozlov:
@anton-bondarev What is the purpose of vfork+exec? I see no pros, only cons:
1. Creating command for simple test case is an overkill
2. These 2 processes should communcate (amount of memory available to allocation at first try should be equal to one at second try). With fork+exec it will be harder to pass this info from child to parent, it's possible, but for what?

anton-bondarev:
@AntonKozlov Ok, I've just suggested. But I think it's better to separate POSIX and non-standard way. And fix build certainly.

As pros I can offer that communicating between 2 processes requires right implementation of shared-memory, but It's difficult of cause.

anna-marina:
Build works, test is intended to be failing

AntonKozlov:
Test looks good! Lets proceed with mmap itself. 
I hope it's OK to not merge this until x86/qemu stop failing test

PR #1087, opened 2017-06-23
Tracking of regions not mapped with vmem_map_region

anna-marina:
These changes allow mmap test with child to pass. That's only a draft.


src/mem/mmap/mmap_mmu.c
```diff
@@ -112,12 +122,19 @@ void mmap_free(struct emmap *mmap) {
...
+	struct phy_page *phy_page;
 
 	dlist_foreach_entry(marea, &mmap->marea_list, mmap_link) {
 		vmem_unmap_region(mmap->ctx, marea->start, mmu_size_align(marea->end - marea->start));
 
 		marea_destroy(marea);
 	}
+
+	dlist_foreach_entry(phy_page, &mmap->page_list, page_link) {
+		phymem_only_free(phy_page->page, phy_page->page_number);
```
AntonKozlov: This is correct, but something similar should be performed in munmap to prevent double free in such code
```c
p = mmap(ANON)
munmap(p) // phymem_free called
exit(0) // here mmap_clear will try to free phymem again
```

src/mem/phymem.c
```diff
@@ -51,3 +51,7 @@ void phymem_free(void *page, size_t page_number) {
 	munmap(page, PAGE_SIZE() * page_number);
 	page_free(__phymem_allocator, page, page_number);
 }
+
+void phymem_only_free(void *page, size_t page_number) {
```
AntonKozlov: Seems like you encounter phymem_free behavior that calles munmap for freeing memory. That is not correct, so was fixed recently https://github.com/embox/embox/pull/1084/commits/d066c061a36a9d2f012d4cbed4d50346163607f8

Could you rebase this branch on top of master or merge master? You'll get fix in phymem_free and this function could be removed

anna-marina:
Rebased #1085 to this branch

AntonKozlov:
Thanks! Got better, but master still should be merged in and `munmap_anon` function (near `munmap` in vmem_device_memory.c) should be fixed to release corresponding `phy_page` structs

src/mem/vmem/vmem_device_memory.c
```diff
@@ -83,7 +85,16 @@ static int munmap_anon(void *addr, size_t size) {
 	 * actual number of pages is (size + 1)/MMU_PAGE_SIZE */
 	int pages = (addr + size - start + MMU_PAGE_SIZE - 1) / MMU_PAGE_SIZE;
 
+	phy_page = mmap_find_phy_page(emmap, start);
+	if (NULL == phy_page) {
+		return SET_ERRNO(ENOENT);
+	}
+
 	phymem_free(start, pages);
```
AntonKozlov: There is a memory leak if pages != phy_page->page_number. Could you check this and in inequal case return EINVAL? That's would be enough for merging this.

But in theory, munmap could be called on range different than was mmaped before.
Consider the following 
```
p1 = mmap(anon, 0x4000) // returns 0x1000
p2 = mmap(anon, 0x4000) // returns 0x6000
```
in result, there is 2 maps 0x1000 - 0x5000 and 0x6000 - 0xa000
As far as I understand munmap interface, munmap(0x0, 0xa000) should munmap both maps.
munmap(0x2000, 0x1000) should result in 3 maps: 0x1000 - 0x2000, 0x3000 - 0x5000 and 0x6000 - 0xa000 (last one unchanged).

I think we could iterate over mmap->page_list and check all kind of intersections of munmapping range and mmaped ranges.
Then we could just alter `page` or `page_number` fileds of phy_page if one of it's end is in unmapping range.
Munmapping range could split one phy_page in two if none of it's ends in munmapping range. One phy_page could be just changed plus new one should be created. Inserting in mmap->page_list is probably OK but potentionally dangerous, since we are modifying list we are iterting over.

So correct implementation is a bit complex, so I suppose merge this PR first, then consider fixing.

anna-marina:
Isn't it already uncommented? https://github.com/embox/embox/blob/master/templates/x86/qemu/mods.config#L85

src/mem/vmem/vmem_device_memory.c
```diff
@@ -90,6 +90,10 @@ static int munmap_anon(void *addr, size_t size) {
 		return SET_ERRNO(ENOENT);
 	}
 
+	if (pages != phy_page->page_number) {
```
vtols: There's possible type error (`size_t` and `int` comparison)

PR #1088, opened 2017-06-24
Trying to fix GCC 7 build errors

anna-marina:
Is it possible to change compiler flags only for single source file in project?
What does `-Wno-gnu-designator` mean? GCC says it doesn't know such option

### DIFF
--- a/src/drivers/block_dev/partition/partition.c
+++ b/src/drivers/block_dev/partition/partition.c
@@ -18,12 +18,14 @@
 #include <drivers/block_dev/partition.h>
 #include <framework/mod/options.h>
 
+#define PART_NAME_MAX 64
+
 /* TODO Create Partition as drive */
 int create_partitions(struct block_dev *bdev) {
 	struct mbr mbr;
 	int rc;
 	int part_n;
-	char part_name[50];
+	char part_name[PART_NAME_MAX];
 	struct block_dev *part_bdev;
 
 	/* Read partition table */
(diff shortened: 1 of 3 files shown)


Deryugin:
It seems that  `-Wno-gnu-designator`  is related to CLANG compiler and is used for some diagnostics ( https://clang.llvm.org/docs/DiagnosticsReference.html#wgnu-designator ). However, I have no problems with this flag with x86/qemu (gcc 6.3.0) and arm/qemu (arm-none-eabi-gcc 5.4.1) templates.

Which template are you trying to build and what is your GCC version? Please, provide `gcc --version` (`arm-none-eabi-gcc --version` if you build for ARM and so on).

anna-marina:
```
$ gcc --version
gcc (GCC) 7.1.1 20170528
```

Deryugin:
Got the same error with gcc-7.1.0
I think you should add check for compiler like this:

```
ifeq ($(COMPILER),clang)
override COMMON_CCFLAGS += -Wno-gnu-designator
endif
```

Deryugin:
Actually, in GCC-6 this changes cause build errors, so I added check for GCC version.

src/drivers/block_dev/partition/partition.c
```diff
@@ -23,7 +23,7 @@ int create_partitions(struct block_dev *bdev) {
 	struct mbr mbr;
 	int rc;
 	int part_n;
-	char part_name[0x16];
+	char part_name[50];
```
Deryugin: It's about compiler error:
`src/drivers/block_dev/partition/partition.c:45:26: error: ‘%s’ directive writing up to 31 bytes into a region of size 16 [-Werror=format-overflow=]`

sikmir:
Problem still exists. Got the same error (https://github.com/embox/embox/issues/1093).

PR #1098, opened 2017-07-16
Fix xen build

AntonKozlov:
First commit contain just whitespace fixes, second one really fixes build.
I think they are easier to review one by one


AntonKozlov:
e746c1f have codestyle errors fixed, but there are more issues. I consider those invalid and will fix by checker config changes. Any thoughts?

AntonKozlov:
I keet hope to fix codestyle errors to pass the check, but end up on bug in checker. 
I'll make codestyle fixes in separate merge (in some time)

PR #1100, opened 2017-07-20
Codestyle check improvement

AntonKozlov:
1. Only changes checked on codestyle
2. Check under CI will be error only on pull request. That's mean there will be no error if check fail on push (no more failed travis badge after)


AntonKozlov:
Please don't merge until https://github.com/embox/embox/pull/1100/commits/d1372d1a7213ffd905adb7e15b3418a431a75592 (DONT_PUSH Break codestyle for example) commit removed

AntonKozlov:
Codestyle check job for PR will fail if style issue found: https://travis-ci.org/embox/embox/builds/255726587
The check is passing for push hook of same commit: https://travis-ci.org/embox/embox/builds/255723830

AntonKozlov:
@embox/devs, could you please review this?

scripts/codestyle/run_uncrustify.sh
```diff
@@ -1,103 +1,68 @@
...
+while getopts :qh opt; do
+	case $opt in
+	q)
+		verbose=""
+		;;
+	h)
+		echo "run_uncrustify.sh [options] [<git diff args>]"
+		echo ""
+		echo "Check codestyle of a change. Change generated by bypassing"
+		echo "diff-generating arguments to git-diff"
```
abusalimov: I don't get the second sentence TBH. Do you mean that the change being checked for code style conformance is generated by running git-diff with specified arguments? Or am I missing something?

scripts/codestyle/run_uncrustify.sh
```diff
@@ -1,103 +1,68 @@
...
 
-rm -rf out
 
-if [ -s uncrustify_diff.txt ]
-then
-	echo "Some files don't comply with code style:"
-	if $VERBOSE; then
-		cat uncrustify_diff.txt
+	if [ $verbose ]; then 
+		gitarg="--no-pager diff"
```
abusalimov: I'd also add `--color`, since Travis supports colored output.

scripts/codestyle/travis_uncrustify.sh
```diff
@@ -1,5 +1,18 @@
...
-else
-	bash ./run_uncrustify.sh -v $(git rev-parse HEAD~1) $TRAVIS_PULL_REQUEST_SHA
+#!/bin/sh
+
+git fetch origin master:master
+git fetch --depth=1000 # unlikely the branch will have more than 1000 commits
+
+revs="$(git rev-parse --short master)...$(git rev-parse --short HEAD)"
+
+echo Checking $revs
```
abusalimov: `... for code style conformance` (otherwise it's hard to get it from the surrounding output)

scripts/codestyle/travis_uncrustify.sh
```diff
@@ -1,5 +1,18 @@
...
+
+$(dirname $0)/run_uncrustify.sh $revs
+result=$?
+
+if [ $result != 0 ] && [ x"$TRAVIS_PULL_REQUEST" = xfalse ]; then
+	echo "Codestyle issues are not fatal for not Pull-Request jobs"
```
abusalimov: How about `Ignoring code style violations (non-PR job)`

scripts/codestyle/run_uncrustify.sh
```diff
@@ -1,103 +1,68 @@
...
 	fi
-	rm uncrustify_diff*
-	cd -
-	exit 1
-fi
 
-rm uncrustify_diff* &> /dev/null
-cd - &> /dev/null
-exit 0 #neccessary if 'rm' fails
+	git $gitarg --no-index -- $from $to
```
abusalimov: I'd also strip the first two lines using `tail`. Compare:
```diff
diff --git a/out/undiff/./src/kernel/init.c+20,8 b/out/fixed/./src/kernel/init.c+20,8
index 0163fe2..bb002c8 100644
--- a/out/undiff/./src/kernel/init.c+20,8
+++ b/out/fixed/./src/kernel/init.c+20,8
@@ -1,8 +1,7 @@
 /**
  * The setup of the system, the run level and execution of the idle function.
  */
-void kernel_start(void)
-{
+void kernel_start(void) {
 
 	kernel_init();
```
---
```diff
--- a/out/undiff/./src/kernel/init.c+20,8
+++ b/out/fixed/./src/kernel/init.c+20,8
@@ -1,8 +1,7 @@
 /**
  * The setup of the system, the run level and execution of the idle function.
  */
-void kernel_start(void)
-{
+void kernel_start(void) {
 
 	kernel_init();
```

abusalimov: Ideally, the header lines should also be more descriptive, at least:
```diff
--- a/BAD/./src/kernel/init.c+20,8
+++ b/GOOD/./src/kernel/init.c+20,8
```
or just
```diff
--- a/src/kernel/init.c+20,8
+++ b/src/kernel/init.c+20,8
```
to reduce the visual clutter (colors are self-descriptive enough). This is nitpicking though.

scripts/codestyle/run_uncrustify.sh
AntonKozlov: Yeah, that's correct. Obviously, I spent not enough time in writing clear messages :-)
How about 
```
Check codestyle of a change. The change generated by passing
<git diff args> to git diff. No args means changes in worktree.
```

abusalimov: > The change generated by passing
A verb is missing. `The change IS generated ...`

> No args means changes in worktree.
I still don't get it. What are other options besides using changes in the worktree? (BTW the "the" article in front of "worktree").
I though these args are supposed to alter the diff output, like the number of context lines, etc. What else could they affect?

scripts/codestyle/run_uncrustify.sh
AntonKozlov: git-diff should determ color support and seems it thinks there is no one for travis log.
I think I'll force color as you suggest. It will make parsing this script output much harder if possible, but for now it's already really complicated.

scripts/codestyle/run_uncrustify.sh
AntonKozlov: This needs careful implementation. I beleive `git diff | tail` will break the script, as return code will be reported for the last process of pipe. I haven't found any magic option suppressing these 2 lines.

scripts/codestyle/run_uncrustify.sh
abusalimov: On the second though, if you suppose that the `run_uncrustify.sh` script can be used outside Travis, like in a batch mode, it would be better to force the colored output from the `travis_uncrustify.sh` side:
```
git config --global color.ui true
```
But then you need to pass the explicit `--color=never` option to other git invocations throughout the script: `git diff --no-prefix $diffargs -- 'src/*.[ch]' | ...`

scripts/codestyle/run_uncrustify.sh
abusalimov: Hmm, that makes sense. In Bash, there's a way to do that though:
```console
$ false | true
$ echo "${PIPESTATUS[0]} ${PIPESTATUS[1]}"
1 0
```

From [here](https://unix.stackexchange.com/questions/14270/get-exit-status-of-process-thats-piped-to-another).

scripts/codestyle/run_uncrustify.sh
AntonKozlov: Primary intention was to specify change source, like
```
Check codestyle of a change. The change is generated by passing
<git diff args> to git diff. Like:

run_uncrustify.sh                      # check working tree
run_uncrustify.sh --cached             # check index
run_uncrustify.sh origin/master...HEAD # check current branch before PR
```
I think I'll change help message to this :-)

abusalimov: Ah, OK, now I see. Thanks for clarifying! =)

AntonKozlov:
Eldar, thanks for the review! 
If I haven't forgot something, I've fixed all issues except file names in diff output. They are still
```
--- a/out/diff/./src/kernel/init.c+20,8
+++ b/out/fixed/./src/kernel/init.c+20,8
```
Fix for this as I imagine it (sed/awk/perl in different proportions) is really harder to create than the issue it fixes, IMHO.

scripts/codestyle/run_uncrustify.sh
```diff
@@ -13,8 +13,12 @@ while getopts :qh opt; do
...
 		echo "run_uncrustify.sh [options] [<git diff args>]"
 		echo ""
-		echo "Check codestyle of a change. Change generated by bypassing"
-		echo "diff-generating arguments to git-diff"
+		echo "Check codestyle of a change. The change is generated by passing"
+		echo "<git diff args> to git diff. Like:"
+		echo ""
+		echo "run_uncrustify.sh                      # check working tree"
+		echo "run_uncrustify.sh --cached             # check index"
+		echo "run_uncrustify.sh origin/master...HEAD # check current branch"
```
abusalimov: git-related: is it really `origin/master...HEAD` or rather `origin/master..HEAD`? I always confuse the notation.

abusalimov:
You're welcome!

> Fix for this as I imagine it (sed/awk/perl in different proportions) is really harder to create than the issue it fixes, IMHO.

This makes sense indeed.

scripts/codestyle/run_uncrustify.sh
AntonKozlov: That's a good question! Usually I try both until I'm satisfied with result :-)
Seems like in diff case triple-dot form is likely desirable.
```
       git diff [--options] <commit> <commit> [--] [<path>...]
           This is to view the changes between two arbitrary <commit>.

       git diff [--options] <commit>..<commit> [--] [<path>...]
           This is synonymous to the previous form. ...
       
       git diff [--options] <commit>...<commit> [--] [<path>...]
           This form is to view the changes on the branch containing and up to the second <commit>, starting at a common ancestor of both <commit>. "git
           diff A...B" is equivalent to "git diff $(git-merge-base A B) B".
```

I assume usually one want to check their branch changes without changes from master. Let's suppose such (we were about to break record of number of messages in a talk about git without a picture, so I'm sorry :-))
```
$ tig
db467e3 Anton Kozlov         M─┐ [master] {origin/master} Merge pull request #1098 from embox/fix-xen-build
fb87b2c Anton Kozlov         │ o [HEAD] [fix-xen-build]  Remove termios dependency on sched
3fe7991 Anton Kozlov         M─│─┐ Merge pull request #1099 from embox/xen-vagrant
79be763 Anton Kozlov         │ │ o Add vagrant defintion for xen dev vm
7184093 Anton Bondarev       M─┴─┘ Merge pull request #1097 from embox/openlibm-improve
...

$ git --no-pager diff --stat 3fe7991..fb87b2c 
 .gitignore                    |  3 ---
 Vagrantfile                   | 27 ---------------------------
 src/drivers/tty/termios_ops.c |  4 +---
 src/drivers/tty/termios_ops.h |  2 ++
 src/drivers/tty/tty.c         |  4 ++++
 5 files changed, 7 insertions(+), 33 deletions(-)

$ git --no-pager diff --stat 3fe7991...fb87b2c 
 src/drivers/tty/termios_ops.c | 4 +---
 src/drivers/tty/termios_ops.h | 2 ++
 src/drivers/tty/tty.c         | 4 ++++
 3 files changed, 7 insertions(+), 3 deletions(-)
```
Note that tripple dot don't try to revert 3fe7991 just because it's not accessible from current branch.

Funny part, that git log takes revset described in gitrevisions(7).Double and tripple dot notation have _opposite_ semantics there. I don't know what forced git developers to make so.

PR #1102, opened 2017-07-27
Fix template pjsip/stm32f4discovery

alexkalmuk:
Fix build of pjsip/stm32f4discovery template

### DIFF
--- a/mk/extbld/toolchain.mk
+++ b/mk/extbld/toolchain.mk
@@ -48,15 +48,19 @@ else
 root2dist = $1
 endif
 
+# TODO: Probably it is more correct to impove Embox's cxx includes, rather than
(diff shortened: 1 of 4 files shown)


alexkalmuk:
@AntonKozlov Hi, it seems something went "wrong" with uncrustify suggestion in case of ifdef/endif block.
https://travis-ci.org/embox/embox/jobs/258137421
Or I'm missed something and my code should be fixed in the proposed way?

AntonKozlov:
Yep, uncrustify is wrong. Seems like it take correct indentation level but fails to reduce it with closing braces.
Anyway, code changes from codestyle job are suggestions, so it's up to devoleper/reviewer to decide.

alexkalmuk:
@AntonKozlov Ok, thanks.

PR #1111, opened 2017-08-22
Fix FS for STM32F4Discovery

Deryugin:
* Use new version of CUBE
* Fix FAT


src/fs/driver/fat/fat_common.c
```diff
@@ -23,7 +23,7 @@
 #define SYSTEM16 "FAT16   "
 #define SYSTEM32 "FAT32   "
 
-uint8_t fat_sector_buff[FAT_MAX_SECTOR_SIZE]; /* XXX */
+uint8_t fat_sector_buff[FAT_MAX_SECTOR_SIZE] __attribute__((aligned(16))); /* XXX */
```
anton-bondarev: Is it a global variable?
May be add an option for alignment?

Deryugin: Yes, it's supposed to be global because it's used in different implementations of FAT (old vfs and dvfs). I don't think that it worth to add another parameter for alignment as it's common problem of many blockdev- and dma-related buffers, and adding options for every buffer is not good. We rather need a system-wide alignment for such buffers, so i want to leave this problem unresolved for now.

PR #1113, opened 2017-10-21
More extensive testing on Travis

AntonKozlov:
Emdocker suggested to build and test embox on travis. This allowed to extend test base up to all templates supported by `continious/run.sh`

Some tests are failing
* arm/qemu: broken by #1112 (https://travis-ci.org/embox/embox/builds/279502569)
* x86/test/lang: languages are running, until lua trying to load http://habrahabr.ru hangs. Ping to habrahabr is successful.
* x86/test/fs: 
  * nfs: some tests are passing, some tests are failing; failed tests are on embox side, seems like nfs driver issue
  * cifs connect timed out; this is reported from embox and host side. Haven't investigated further
* x86/test/net:
  * dropbear is failed to start
  * (when dropbear is disabled) at some time embox got `assert(skb)` failed, leading to tests failures
* sparc/qemu, mips/qemu, ppc/qemu, microblaze/petalogix: 
  * failed to build as architectures is not supported by openlibm


AntonKozlov:
Hi,

> Lets try to fix most errors before merge to master

Some from these errors looks rather hard to investigate and may be not worth doing (cifs, for example)
From my point of view, it's better to mask out all big (or may be all) issues and keep the list of such issues in github issue. Then it will be possible to fix them one-by-one

> If you rebase under master, fix arm/qemu template

Thanks! Seems to work. But I'd rather postpone rebase prior final merging, for now I just merged master into the branch

> Why do you use microblaze/petalogic instead of microblaze/qemu?

I've took a list of tested templates from `continuous/run.sh` Some templates are gone, but this one survived. So, it's obsolete too, I suppose?

anton-bondarev:
> From my point of view, it's better to mask out all big (or may be all) issues and keep the list of such issues in github issue. Then it will be possible to fix them one-by-one

I'm agree. This is good way.

AntonKozlov:
Hi!

I made several changes in this:
 * failed tests disabled, making all checks green. Each disabled check have a commit with "CI FIXME" prefix, description of failure is in the commit message. #1118 have the list of disabled checks and commits. Some failures seems major, like telnet getting garbage, rlogin causing panic, etc.
 * microblaze template name fixed
 * rebased on current master

Ready to merge?

PR #1129, opened 2017-12-23
Add extra args option to docker run

AntonKozlov:
Docker exec bins specific keys to detach from running process (Ctrl-p
Ctrl-q by default). Ctrl-p is emacs up, also is up in readline,
also shell previous command. To make use of container's interactive
shell more plesant, this commit adds abillity to pass extra args to
docker exec, thus remap detach keys to something else, like:

	$ . ./scripts/docker/docker_rc.sh
	$ EMDOCKER_EXEC_EXTRA_ARGS="--detach-keys ctrl-_"

### DIFF
--- a/scripts/docker/docker_rc.sh
+++ b/scripts/docker/docker_rc.sh
@@ -9,5 +9,11 @@ dr() {
 	local tty_opt=
 	[ -t 0 ] && tty_opt="-t"
 
-	docker exec -u user -i $tty_opt emdocker bash -lc "$*"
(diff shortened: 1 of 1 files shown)


AntonKozlov:
Hi, @sikmir, thanks for review. I changed the way args passed because old one (via positional arguments to sourced script), if called from other script, captured outer one argument if no was provided to sourced one. I'll wait with merge before get any opinion from you on this.
Thanks.

PR #1135, opened 2018-01-10
Enable different implementations libm

anton-bondarev:


Deryugin:
If we use stubs for math operations, there's no need in `third_party.lib.NoOpenLibm` module (and `third_party.lib.OpenLibm` shold be non-abstract).

templates/msp430/exp430g2/mods.config
```diff
@@ -17,7 +17,7 @@ configuration conf {
 	include embox.kernel.cpu.bkl_api
 
 	include embox.compat.libc.stdio.print(support_floating=0)
-	include third_party.lib.NoOpenLibm
+	include third_party.lib.libm_none
```
Deryugin: I think it should be `embox.compat.libc.math_libm_none` (same with other templates).

anton-bondarev:
@Deryugin Thanks, I fixed it. Please look again.

src/compat/libc/math/stub.c
```diff
@@ -0,0 +1,39 @@
+/**
+ * @file stub.c
```
AntonKozlov: This implementation should not be used in any circumstances, right? Why should it exist?

anton-bondarev: @AntonKozlov It's for future porting on E2K :)

PR #1136, opened 2018-01-10
net: Fix skb allocation

alexkalmuk:
These changes revert 3107afe030b1c56e82cf648d2d3a977238401610 which is partially wrong since all huge packets are dropped.

This also fixes template x86/qt-vnc

### DIFF
--- a/src/net/skbuff/skb_data.c
+++ b/src/net/skbuff/skb_data.c
@@ -16,6 +16,7 @@
 #include <util/log.h>
 #include <util/member.h>
 #include <util/binalign.h>
+#include <mem/sysmalloc.h>
 
 #include <hal/ipl.h>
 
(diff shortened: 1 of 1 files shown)


src/net/skbuff/skb_data.c
```diff
@@ -125,9 +139,16 @@ void skb_data_free(struct sk_buff_data *skb_data) {
 	sp = ipl_save();
 	{
 		if (--skb_data->links == 0) {
-			assert(pool_belong(&skb_data_pool, skb_data));
-
-			pool_free(&skb_data_pool, skb_data);
+			assert(skb_data->alloc_type == ALLOCATED_POOL ||
+					skb_data->alloc_type == ALLOCATED_MALLOC);
+			switch (skb_data->alloc_type) {
+			case ALLOCATED_POOL:
```
anton-bondarev: default: assert()

alexkalmuk: Thanks, fixed.

src/net/skbuff/skb_data.c
```diff
@@ -125,9 +139,17 @@ void skb_data_free(struct sk_buff_data *skb_data) {
 	sp = ipl_save();
 	{
 		if (--skb_data->links == 0) {
-			assert(pool_belong(&skb_data_pool, skb_data));
-
-			pool_free(&skb_data_pool, skb_data);
+			switch (skb_data->alloc_type) {
+			case ALLOCATED_POOL:
```
AntonKozlov: Another option is to use `pool_belong`, in this case no need to store allocation metainfo

alexkalmuk: Yes, that's exactly like it was previously done, but as for me, it is more safely and clearly to state explicitly how the object is allocated.

PR #1146, opened 2018-04-02
ARM VFP/NEON support

Deryugin:
Add support for VFP/NEON instructions. Currently it works only with NEON, but VFP is subset of NEON, so it's possible to cut it if necessary in future.

### DIFF
--- a/src/arch/arm/fpu/vfp.h
+++ b/src/arch/arm/fpu/vfp.h
@@ -1,0 +1,68 @@
+/**
+ * @file vfp.h
+ * @brief
+ * @author Denis Deryugin <deryugin.denis@gmail.com>
+ * @version
+ * @date 28.03.2018
+ */
+
+#ifndef ARM_FPU_VFP_H_
+#define ARM_FPU_VFP_H_
+
+#define FPU_DATA_LEN 65 /* One word for coprocessor state
+			   and 64 words for VPF registers */
+
+#ifndef __ASSEMBLER__
+#include <string.h>
+static inline void arm_fpu_context_init(void *opaque) {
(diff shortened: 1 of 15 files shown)


templates/arm/iwave_imx6/mods.config
```diff
@@ -12,6 +12,8 @@ configuration conf {
 	include embox.arch.arm.vfork
 	@Runlevel(0) include embox.driver.cache.pl310(base_addr=0x00A02000)
 
+	include embox.arch.arm.fpu.vfp(log_level=4)
```
anton-bondarev: log_level=4?

src/arch/arm/fpu/vfp.h
```diff
@@ -0,0 +1,68 @@
...
+#ifndef ARM_FPU_VFP_H_
+#define ARM_FPU_VFP_H_
+
+#define FPU_DATA_LEN 65 /* One word for coprocessor state
+			   and 64 words for VPF registers */
+
+#ifndef __ASSEMBLER__
+#include <string.h>
+static inline void arm_fpu_context_init(void *opaque) {
+	memset(opaque, 0, sizeof(uint32_t) * FPU_DATA_LEN);
```
anton-bondarev: It's really  necessary ?

templates/arm/iwave_imx6/mods.config
Deryugin: Fixed

src/arch/arm/fpu/vfp.h
Deryugin: Well, probably this can be removed, but I think it's ok to make sure there's no garbage in this memory.

PR #1147, opened 2018-04-04
Interrupt controller

miha2733:


src/drivers/interrupt/xen.c
```diff
@@ -6,18 +6,43 @@
...
 
-int test_irqctrl_div_zero;
+#define CLEAR_BIT(field, bit) __asm__ __volatile__ ("lock btrl %1,%0":"=m" ((field)):"Ir"((bit)):"memory")
+
+extern shared_info_t xen_shared_info;
+shared_info_t *shared_info;
 
 EMBOX_UNIT_INIT(xen_int_init);
 
+int port;
```
AntonKozlov: `static int port`

src/arch/xen/traps.c
```diff
@@ -47,7 +47,7 @@ void machine_check(void);
 void do_divide_error(struct pt_regs *regs, int error_code) {
 	printk("DIVIDE ERROR called\n");
 
-	extern int test_irqctrl_div_zero;
+	int test_irqctrl_div_zero = 10; //TODO: Em... ?
```
AntonKozlov: `test_irqctrl_div_zero` is not required anymore, should be deleted

src/drivers/interrupt/xen.c
```diff
@@ -29,15 +54,20 @@ void irqctrl_disable(unsigned int irq) {
...
-	test_irqctrl_div_zero = 0;
-	printk("%d\n", y); /* avoid optimizing out by compiler */
+	struct evtchn_send event;
+	event.port = port;
+	shared_info = &xen_shared_info;
+	int size = sizeof(xen_ulong_t) * 8;
+	//shared_info->evtchn_mask[irq / size] &= ~(1UL << irq % size); //TODO: Need to choose..
+	//CLEAR_BIT(shared_info->evtchn_mask[irq / size], irq % size);
+	shared_info->evtchn_mask[irq / size] &= 0;
+	shared_info->vcpu_info[0].evtchn_upcall_mask = 0; //TODO: Searching for the right bit..
```
AntonKozlov: ouch :)

src/arch/xen/event.c
```diff
@@ -126,6 +127,9 @@ void do_hypervisor_callback(struct pt_regs *regs)
 
 			/* Combine the two offsets to get the port */
 			evtchn_port_t port = (pending_selector << 5) + event_offset;
+
+			irq_dispatch(port);
```
AntonKozlov: My bad, I suggested to call just `irq_dispatch`. Actually, interrupt handling scheme is rather good defined, for example, src/arch/x86/kernel/irq_handler.c. This required to be extended to that in order to support nested interrupts

PR #1149, opened 2018-04-11
Added basic xenstore support

Columpio:
1. Breakpoint on 49 line of src/arch/xen/kernel_init.c
2. Continue
3. print list
4. got: `$1 = "EACCES\000\000\362\313\000\000\363\313\000\000\364\313\000\000\365\313\000\000\000\000\000\000 \00 0\000"`

### DIFF
--- a/src/arch/xen/kernel_init.c
+++ b/src/arch/xen/kernel_init.c
@@ -14,12 +14,14 @@
 #include <xen/event_channel.h>
 #include <xen/io/console.h>
 #include <xen/event.h>
+#include <xenstore.h>
 
 /* Embox interface */
 extern void kernel_start(void);
 
 /* Xen interface */
 extern void trap_init(void);
+extern int xenstore_init(start_info_t *);
 
 uint8_t xen_features[XENFEAT_NR_SUBMAPS * 32];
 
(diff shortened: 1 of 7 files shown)


src/arch/xen/xenstore.c
```diff
@@ -0,0 +1,221 @@
+#include "xenstore.h"
```
AntonKozlov: There is no copyright here. Have the source any? If so, please add it (and yours too)

src/arch/xen/kernel_init.c
```diff
@@ -38,6 +42,12 @@ void xen_kernel_start(start_info_t * start_info) {
 	
 	init_events();
 
+
+	xenstore_init(start_info);
+	char list[32];
+	uint32_t len = xenstore_ls("/", list, 32);
+	printk("len = %d\n", len);
```
AntonKozlov: you're calling printk before `diag` initialized, which performed later in `kernel_start`

Columpio:
@AntonKozlov I asked you not to merge this because this is testing branch. What I mean is that calling xenstore_ stuff in kernel_init.c file is totally wrong and is made for testing only.

src/arch/xen/kernel_init.c
Columpio: thx

AntonKozlov:
That's true, but it is easy to fix. Do you have objections for me merging https://github.com/AntonKozlov/embox/tree/pull-1149? It's this PR with my commit on top

Columpio:
Seems to work ^_^

AntonKozlov:
Was unable to compile, starting from unit init function incompatible type up to missing `test_assert*` declarations (definitions, actually). I addressed that in next commit by moving test function to separate test module. 

Please let me know if my modifications caused any harm. Thanks!

PR #1154, opened 2018-05-19
EMAC improvement

Deryugin:
* Added support for EMAC on AM3505
* Now really initialize PHY (previously it was hardcoded or taken from uboot)


anton-bondarev:
LGTM

PR #1156, opened 2018-05-19
Gold linux was added

miha2733:


linuxKernel/run_linux
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+sudo xl -vvv create -c -f config.cfg
```
AntonKozlov: The script only works only if started from it's directory. 
```
root@vagrant:/embox# ./linuxKernel/run_linux
Failed to read config file: config.cfg: No such file or directory
```
But, and that's interesting, it also allows to specify relative path to kernel as just `"linux"`. I used more sophisticated scheme in emox `run_xen`, but this one looks simplier (better).

If you add next before `xl create ...`, the issue would be fixed without script compilcation
```
cd $(dirname $0)
```

AntonKozlov:
Thanks, merged in by 18ed499f69b2a17f3219884d9559c661107ced05

PR #1159, opened 2018-05-25
Fix backtrace for ARM

Deryugin:
Now it works for Integrator/CP and Cortex-A9

### DIFF
--- a/src/arch/arm/lib/debug/stack_iter.c
+++ b/src/arch/arm/lib/debug/stack_iter.c
@@ -18,11 +18,11 @@
 #include "stack_iter.h"
 
 int stack_iter_next(stack_iter_t *f) {
-	f->fp = (void*) *((int*)f->fp - 3);
-	f->pc = (void*) *((int*)f->fp);
-	f->lr = (void*) *((int*)f->fp - 1);
+	f->fp = (void *) *((int *) (f->fp - 0xc));
+	f->pc = (void *) *((int *) f->fp);
+	f->lr = (void *) *((int *) (f->fp - 0x4));
 
-	return *((int*)f->fp - 3) != 0;
+	return (void *) *((int *) (f->fp - 0xc)) != 0;
 }
 
 void stack_iter_context(stack_iter_t *f, struct context *ctx) {
@@ -48,11 +48,10 @@ void stack_iter_current(stack_iter_t *f) {
 	 * by hand */
 
 	stack_iter_next(f);
-	f->fp = (void*) *((int*)f->fp - 3);
 }
 
 void *stack_iter_get_fp(stack_iter_t *f) {
-	return f->fp - 3;
+	return f->fp - 4;
 }
 
 void *stack_iter_get_retpc(stack_iter_t *f) {
--- a/src/arch/arm/lib/debug/stack_iter.c
+++ b/src/arch/arm/lib/debug/stack_iter.c
@@ -50,10 +50,6 @@ void stack_iter_current(stack_iter_t *f) {
 	stack_iter_next(f);
 }
 
-void *stack_iter_get_fp(stack_iter_t *f) {
-	return f->fp - 4;
-}
-
 void *stack_iter_get_retpc(stack_iter_t *f) {
 	return f->pc;
 }
--- a/src/arch/sparc/lib/debug/stack_iter.c
+++ b/src/arch/sparc/lib/debug/stack_iter.c
@@ -36,10 +36,6 @@ int stack_iter_next(stack_iter_t *f) {
 	}
 }
 
-void* stack_iter_get_fp(stack_iter_t *f) {
-	return (*f)->reg_window.fp;
-}
-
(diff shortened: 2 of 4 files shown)


src/arch/arm/lib/debug/stack_iter.c
```diff
@@ -18,11 +18,11 @@
 #include "stack_iter.h"
 
 int stack_iter_next(stack_iter_t *f) {
-	f->fp = (void*) *((int*)f->fp - 3);
-	f->pc = (void*) *((int*)f->fp);
-	f->lr = (void*) *((int*)f->fp - 1);
+	f->pc = (void *) *((int *)f->fp);
+	f->lr = (void *) *((int *)f->fp);
+	f->fp = (void *) *((int *) (f->fp - 4));
```
AntonKozlov: considering old and new version assumes 4 word stack frame header, it's APCS. But in APCS header is 
```
    save code pointer       [fp]        <-fp points to here
    return link value       [fp, #-4]
    return sp value         [fp, #-8]
    return fp value         [fp, #-12]
```
(https://www.cl.cam.ac.uk/~fms27/teaching/2001-02/arm-project/02-sort/apcs.txt)

Could you describe why old code is not correct, when it's apparently is?

Deryugin: Basicly, new version works with `-fno-omit-frame-pointer` and old version doesn't 🤔
However, it works with `-mapcs-frame`, so ok, I will roll back these changes :)

Deryugin: (by the way, old code was incorrect anyway because we need to do `f->fp = (void*) *((int*)f->fp - 0xc);`, not `f->fp = (void*) *((int*)f->fp - 3);`

AntonKozlov: As far as I understand, both flags should be specified (`-fno-omit-frame-pointer -mapcs-frame`) to unwind stack in such way. 
Without `-fno-omit-frame-pointer` compiler is allowed to not generate full frame header for leaf functions.
Without `-mapcs-frame`, frame can't be unwinded without consulting debug info, generated by compiler (those .extab, .exidx sections)
I believe, with other options it will break from time to time

src/arch/arm/lib/debug/stack_iter.c
```diff
@@ -18,11 +18,11 @@
 #include "stack_iter.h"
 
 int stack_iter_next(stack_iter_t *f) {
-	f->fp = (void*) *((int*)f->fp - 3);
-	f->pc = (void*) *((int*)f->fp);
-	f->lr = (void*) *((int*)f->fp - 1);
+	f->fp = (void *) *((int *) (f->fp - 0xc));
+	f->pc = (void *) *((int *)f->fp);
+	f->lr = (void *) *((int *)f->fp - 0x4);
```
AntonKozlov: Access 4 _words_ below `fp`, are parentheses required? `(f->fp - 0x4)`

src/arch/arm/lib/debug/stack_iter.c
Deryugin: It seems that you are right. Thanks for your remark.

src/arch/arm/lib/debug/stack_iter.c
```diff
@@ -48,11 +48,10 @@ void stack_iter_current(stack_iter_t *f) {
 	 * by hand */
 
 	stack_iter_next(f);
-	f->fp = (void*) *((int*)f->fp - 3);
 }
 
 void *stack_iter_get_fp(stack_iter_t *f) {
-	return f->fp - 3;
+	return f->fp - 4;
```
AntonKozlov: `stack_iter_get_fp` is not used, can't check correctness of this change

Deryugin: 🤔
Let's just delete this function :)

Deryugin: It's trivial for x86 and sparc, so I think it's ok to remove it from these architectures as well

src/arch/arm/lib/debug/stack_iter.c
AntonKozlov: > By the way, old code was incorrect anyway because we need to do f->fp = (void*) *((int*)f->fp - 0xc);, not f->fp = (void*) *((int*)f->fp - 3);

Sorry, saw this too late. Just to be clear.

`(int*)fp - 3` (was) is the same as `(int*) (fp - 12)` (in your recent commit). But `(int*)f->fp - 0xc` from the message is not and it's not correct, it's access 12 words below `fp`.

Actually, I would prefer `(int*)fp - 3` than `(int*) (fp - 12)` because fp is `void*` and it's gcc extension to C that allows that. Maybe `uint32_t*` is better that `int*` here...

src/arch/arm/lib/debug/stack_iter.c
Deryugin: Yep. Added it.

PR #1161, opened 2018-05-28
Implement generic PHY interface

Deryugin:


src/drivers/net/phy/phy.h
```diff
@@ -0,0 +1,22 @@
...
+ * @date 28.05.2018
+ */
+
+#ifndef NET_PHY_H_
+#define NET_PHY_H_
+
+#include <net/mii.h>
+#include <net/netdevice.h>
+
+int phy_detect(struct net_device *dev);
```
anton-bondarev: Use 'extern' in functions declarations

PR #1167, opened 2018-06-09
Multitasking tests fix

anton-bondarev:


src/tests/kernel/sched/running_threads_test.c
```diff
@@ -6,54 +6,71 @@
...
+		if (args.end1 && args.end2) {
+			break;
+		}
+		ksleep(2);
+	}
 }
 
 static void *t1_run(void *arg) {
-	int i = 0;
+	struct thread_args *volatile args = arg;
```
alexkalmuk: I am not sure this 'volatile' qualifier is required for the local variable.

src/compat/posix/proc/fork_nommu/fork.c
```diff
@@ -50,28 +50,28 @@ void _NORETURN fork_body(struct pt_regs *ptregs) {
...
+		fork_stack_store(child_adrspc, child->tsk_main);
+		fork_heap_store(&child_adrspc->heap_space, task_self());
+		fork_static_store(&child_adrspc->static_space);
 
-	memcpy(&child_adrspc->pt_entry, ptregs, sizeof(*ptregs));
+		memcpy(&child_adrspc->pt_entry, ptregs, sizeof(*ptregs));
 
-	sched_lock();
-	{
 		child = task_table_get(child_pid);
```
alexkalmuk: This is redundant, the same at line 62.

src/compat/posix/proc/fork_nommu/fork_addr_space_heap.c
```diff
@@ -28,20 +29,33 @@ static inline struct dlist_head *task_mspace(struct task *tk) {
...
 	}
+	assert(hpspc->heap);
 	mspace_deep_store(task_mspace(tk), &hpspc->store_space, hpspc->heap);
 }
 
 void fork_heap_restore(struct heap_space *hpspc) {
+	assert(hpspc);
+
+	if (NULL == hpspc->heap) {
+		return;
```
alexkalmuk: Why could it happen? Is it correct situation?

src/tests/posix/sys/timerfd/timerfd_tests.c
```diff
@@ -79,7 +82,7 @@ TEST_CASE("timerfd set and get (one-time timer)") {
 	test_assert_true(timespec_is_zero(&curr_value.it_interval));
 
 	// second gettime: timer expired
-	usleep(USEC_PER_SEC / 4);
+	usleep(USEC_PER_SEC * 2);
```
alexkalmuk: It seems its really so bad with our timers because the timer was set up at 1 second, and now you check whether it expired only after 2 seconds. Why? Because the timer didn't expire after 1 sec like it was assumed?

src/compat/posix/proc/fork_nommu/fork_addr_space_heap.c
```diff
@@ -28,20 +29,33 @@ static inline struct dlist_head *task_mspace(struct task *tk) {
 void fork_heap_store(struct heap_space *hpspc, struct task *tk) {
 	size_t size;
 
+	assert(hpspc);
+	assert(tk);
+
 	size = mspace_deep_copy_size(task_mspace(tk));
+	if (!size) {
+		return;
```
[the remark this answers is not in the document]
anton-bondarev: In this case heap will not store/restore.

src/compat/posix/proc/fork_nommu/fork_addr_space_heap.c
anton-bondarev: it's happend if we is the commetn case
```case  in +	if (!size) {
	return;```

src/compat/posix/proc/fork_nommu/fork_addr_space_heap.c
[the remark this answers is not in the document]
alexkalmuk: But is it normal situation? Which value fork return in that case?

src/compat/posix/proc/fork_nommu/fork.c
anton-bondarev: sched_lock is required because when fork_addr_space_create is happened  irq event may be occurred and then contex_switch in where try save data in a new addr_space which is not initialized in that place

src/compat/posix/proc/fork_nommu/fork_addr_space_heap.c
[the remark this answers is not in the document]
anton-bondarev: It's normal for processes which not use heap (malloc/free)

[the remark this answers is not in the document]
alexkalmuk: Ah, finally I got this, `mspace_deep_copy_size` doesn't allocate memory, it just calculates a summarized size of allocated objects. So now it seems to be a correct code.

src/tests/posix/sys/timerfd/timerfd_tests.c
alexkalmuk: Ok, I created trello card.

PR #1168, opened 2018-06-18
Drivers net

anton-bondarev:
Just move most of network drivers to own folders


Deryugin:
Why stm32f4_eth, loopback and tun are not moved to separate folders?

Also `src/drivers/net/xilinx_emaclite.c` is not mentioned in any Mybuild files, can it be safely removed?

PR #1173, opened 2018-06-29
Templates fix

anton-bondarev:
Fix build for following templates:

- arm/stm32_vl_light
- arm/omap
- arm/stm32_dvfs
- arm/stm32_modbus

### DIFF
--- a/src/arch/arm/embox.lds.S
+++ b/src/arch/arm/embox.lds.S
@@ -20,9 +20,9 @@ _ram_size = LENGTH(RAM);
 
 SECTIONS {
 	.text : {
+		STATIC_IRQ_TABLE
 		*(.trap_table)
 		*(.monitor_trap_table)
-		STATIC_IRQ_TABLE
 
 		*(.flash)
 
--- a/src/drivers/flash/stm32f4/Mybuild
+++ b/src/drivers/flash/stm32f4/Mybuild
@@ -2,6 +2,10 @@ package embox.driver.flash
 
 @BuildDepends(third_party.bsp.st_f4.core)
 module stm32_f4 {
+	option number page_size=0x4000
+	option number flash_size=0xc000
+	option number word_size=4
+
 	source "stm32f4_flash.c"
 	source "stm32f4_flash.lds.S"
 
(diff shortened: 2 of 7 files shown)


templates/arm/ti816x-qt/build.conf
```diff
@@ -16,6 +16,8 @@ CXXFLAGS += -march=armv7-a -mtune=cortex-a8 -mno-unaligned-access
 CXXFLAGS += -fno-rtti
 CXXFLAGS += -fno-exceptions
 CXXFLAGS += -fno-threadsafe-statics
+ /* C++14 */
+CXXFLAGS += -fsized-deallocation
```
AntonKozlov:
```
Enable the built-in global declarations
                   void operator delete (void *, std::size_t) noexcept;
                   void operator delete[] (void *, std::size_t) noexcept;
```
Intention is not clear. Could you describe for what is this?

anton-bondarev: without this flag I got error
```/home/anton/embox/src/compat/cxx/include/standalone_new:49:6: error: 'void operator delete(void*, size_t)' is a usual (non-placement) deallocation function in C++14 (or with -fsized-deallocation) [-Werror=c++14-compat]```

AntonKozlov: Description of the warning is https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wc_002b_002b14-compat
seems like it should catch things that will break when the c++ change language version, i.e. switch from c++11 to c++14.
c++ code also seems to built with -Wall -Werror, compiler have more warnings than expected. Embox should not insist on third-party to compile with those options. If third-party project request them, it's not right for them, suppose there is something like `--disable-warning-as-erros` configure switch.
Does `include/standalone_new` lay in include directory passed as `-isystem`? System includes should not trigger all warnings.
It's also possible to disable the `c++14-compat` warning for the third-party or, at last, globally

anton-bondarev: Yes, It's a new for c++14.
The warning appears during src/compat/cxx/core/new.cpp compilation not third-party
We can disable c++14-compat globaly, but I don't understand why -fsized-deallocation is bad?

anton-bondarev: @AntonKozlov I move commits with c++ to another branch

PR #1175, opened 2018-07-10
add service_list

FilippChubukov:


src/cmds/service.c
```diff
@@ -12,43 +12,68 @@
 #include <unistd.h>
 #include <stdlib.h>
 #include <mem/misc/pool.h>
-
 #include <util/dlist.h>
+#include <sys/types.h>
+#include <signal.h>
+
+#define len 32
```
anton-bondarev: Our codestyle points that macros names must be in upper case
Use more informative  symbols names

src/cmds/service.c
```diff
@@ -12,43 +12,68 @@
...
+static DLIST_DEFINE(head);
 
-static void add_element_in_list(char *name) {
+static void add_element_in_list(const char *name, pid_t service_pid) {
   	struct service_list *new_element;
 
   	new_element = pool_alloc(&pool);
    
-  	strcpy(new_element -> service_name,name);
+  	strncpy(new_element->service_name,name,len);
```
anton-bondarev: better use sizeof(array_name) instead of a constant

src/cmds/service.c
```diff
@@ -12,43 +12,68 @@
...
+	new_element->pid = service_pid;
    
-  	dlist_init(&new_element -> dlist_item);
+  	dlist_init(&new_element->dlist_item);
   
-  	dlist_add_next(&new_element -> dlist_item, &head -> dlist_item); 
+  	dlist_add_next(&new_element->dlist_item, &head); 
 }
 
+static void delete_elem(const char *name) {
```
anton-bondarev: service_delete?

src/cmds/service.c
```diff
@@ -12,43 +12,68 @@
...
 
 POOL_DEF(pool, struct service_list, sizeof(struct service_list));		 
 
-struct service_list *head;
-
-int init_head_flag = 0;
+static DLIST_DEFINE(head);
 
-static void add_element_in_list(char *name) {
+static void add_element_in_list(const char *name, pid_t service_pid) {
```
anton-bondarev: service_add?

src/cmds/service.c
```diff
@@ -11,15 +11,69 @@
...
-	printf("PRINT LIST OF SERVICES IS NOT SUPPORTED YET\n");
+#define len 32
+
+struct service_list {
+	struct dlist_head dlist_item;
+	char service_name[len];
+	pid_t pid;
+};
+
+POOL_DEF(pool, struct service_list, sizeof(struct service_list));
```
Deryugin: Third argument of POOL_DEF is a number of pool objects, so `sizeof struct` has no sense here.

src/cmds/service.c
```diff
@@ -11,15 +11,69 @@
...
+POOL_DEF(pool, struct service_list, sizeof(struct service_list));		 
+
+static DLIST_DEFINE(head);
+
+static void add_element_in_list(const char *name, pid_t service_pid) {
+  	struct service_list *new_element;
+
+  	new_element = pool_alloc(&pool);
+   
+  	strncpy(new_element->service_name,name,len);
```
Deryugin: Here you dereference the pointer, but it can be null (`new_element` can be NULL if pool is full), which will cause system crash, and that's not what we want. It's better to return error code and/or print appropriate error message.

src/cmds/service.c
```diff
@@ -11,15 +11,69 @@
...
+	struct dlist_head dlist_item;
+	char service_name[len];
+	pid_t pid;
+};
+
+POOL_DEF(pool, struct service_list, sizeof(struct service_list));		 
+
+static DLIST_DEFINE(head);
+
+static void add_element_in_list(const char *name, pid_t service_pid) {
```
Deryugin: I guess the better way to name it is `add_element_to_list` or `insert_element_in_list`

src/cmds/service.c
```diff
@@ -11,15 +11,69 @@
...
+  
+  	dlist_add_next(&new_element->dlist_item, &head); 
+}
+
+static void delete_elem(const char *name) {
+	struct service_list *tmp;
+	
+	dlist_foreach_entry(tmp, &head, dlist_item) {
+		if (strcmp(tmp->service_name, name) == 0) {
+			kill(tmp->pid, SIGKILL);
```
Deryugin: `kill` can return error code. For example, it can fail if target process is dead for some reason.

src/cmds/service.c
```diff
@@ -5,21 +5,93 @@
...
-static void srv_print_services(void) {
-	printf("PRINT LIST OF SERVICES IS NOT SUPPORTED YET\n");
+static void service_delete(const char *name) {
+	struct service_list *tmp;
+	
+	dlist_foreach_entry(tmp, &head, dlist_item) {
+		if (strcmp(tmp->service_name, name) == 0) {
+			int kill_result = kill(tmp->pid, SIGKILL);
+			
+			if(kill_result != 0) {
```
Deryugin: Please add a whitespace between `if` and `(`

src/cmds/service.c
```diff
@@ -5,21 +5,93 @@
...
+				printf("Error: the service was launched but is not working now.\n");
+				return;
```
Deryugin: If service task is not working, it's still necessary to remove it from the service list.

src/cmds/service.c
```diff
@@ -42,12 +115,19 @@ int main(int argc, char **argv) {
...
 	command = argv[1];
-	argv[argc] = NULL;
-	printf("Starting service: %s\n", command);
 
+	if (strcmp(argv[2],"stop") == 0) {
+		service_delete(command);
+		return 0;
+	}
+
+	argv[argc] = NULL;
```
Deryugin: Why this is necessary?

src/cmds/service.c
```diff
@@ -42,12 +115,19 @@ int main(int argc, char **argv) {
...
-		srv_print_services();
+		listprint();
 		return 0;
 	}
+
 	command = argv[1];
-	argv[argc] = NULL;
-	printf("Starting service: %s\n", command);
 
+	if (strcmp(argv[2],"stop") == 0) {
```
Deryugin: Need to add one whitespace after comma inside the parameter list for `strcmp`.

src/cmds/service.c
```diff
@@ -30,6 +101,7 @@ static int service_run(const char *path, char *const argv[]) {
 		execv(path, argv);
 		exit(1);
 	}
+	service_add(path,pid);
```
Deryugin: Insert whitespace after the comma, please

PR #1176, opened 2018-07-10
find problem with services(arm/qemu)

FilippChubukov:


src/mem/pagealloc/bitmask.c
```diff
@@ -112,7 +112,7 @@ struct page_allocator *page_allocator_init(char *start, size_t len, size_t page_
 	unsigned int pages;
 	size_t bitmap_len;
 
-	if (len < page_size) {
+	if (len < page_size + 1) {
```
alexkalmuk: As I can see page_allocator_init() doesn't assume len % page_size == 0. That means, even after your fix if you pass len = page_size + 2 into this function you will get the assertion failure anyway.

My proposition is to change assert(pages>0) to return NULL. In that case this function will handle all possible values of 'len' without assertion failures.

Another possible way to make this function correct is to force 'len % page_size == 0'. I mean assert(len % page_size == 0); at the beginning.

src/tests/mem/Mybuild
```diff
@@ -47,3 +55,6 @@ module mmap {
 
 	depends embox.mem.vmem
 }
+
```
alexkalmuk: Extra newlines, remove, please.

src/tests/mem/page_allocator.c
```diff
@@ -0,0 +1,30 @@
...
+#include <mem/phymem.h>
+#include <util/binalign.h>
+
+EMBOX_TEST_SUITE("page_alloc_init test");
+
+TEST_CASE("Init page_allocator with one page") {
+	struct page_allocator *allocator;
+
+	extern char _reserve_end;
+	char *const phymem_alloc_start = (char *)
```
alexkalmuk: It is not a very good idea to use phymem_alloc_start, because of this address is already in use by phymem_allocator. A better way is to prealloc some memory region with phymem_alloc() and then initialize page allocator on this new region:

ptr = phymem_alloc(pages_count);
page_allocator_init(ptr,...);

src/tests/mem/page_allocator.c
```diff
@@ -0,0 +1,30 @@
...
+		binalign_bound((uintptr_t) &_reserve_end, PAGE_SIZE());
+
+	const size_t mem_len = 4096;
```
alexkalmuk: I guess mem_len should be equal to PAGE_SIZE(). If it is so, then it's better to write directly 'mem_len = PAGE_SIZE();'

src/tests/mem/Mybuild
AntonKozlov: I would recommend looking in "uncrustify" job output 

https://travis-ci.org/embox/embox/jobs/403160384

(can be reached from PR travis checks https://travis-ci.org/embox/embox/builds/403160369)

It suggests diff that should fix codestyle (as it understand it), so It's good for quick look for possible issues. But it should be taken with grain of scepsis, sometime it recommend strangely bad changes :)

src/tests/mem/page_allocator.c
```diff
@@ -0,0 +1,34 @@
...
+		
+}
+
+TEST_CASE("Init page_allocator with len < page_size") {
+	struct page_allocator *allocator;
+	char buff[0x100];
+
+	allocator =  page_allocator_init(buff, 0x10, 0x100);
+	test_assert_null(allocator);
+
```
anton-bondarev: remove redundant empty lines

src/tests/mem/page_allocator.c
```diff
@@ -0,0 +1,34 @@
...
+#include <mem/phymem.h>
+#include <util/binalign.h>
+
+EMBOX_TEST_SUITE("page_alloc_init test");
+
+TEST_CASE("Init page_allocator with one page") {
+	struct page_allocator *allocator;
+	char buff[0x10];
+
+	allocator =  page_allocator_init(buff,0x10 ,0x10);
```
alexkalmuk: Space at the wrong place. Should be page_allocator_init(buff, 0x10, 0x10);

src/tests/mem/page_allocator.c
```diff
@@ -0,0 +1,34 @@
...
+
+#include <embox/test.h>
+#include <mem/page.h>
+#include <mem/heap.h>
+#include <mem/phymem.h>
+#include <util/binalign.h>
+
+EMBOX_TEST_SUITE("page_alloc_init test");
+
+TEST_CASE("Init page_allocator with one page") {
```
alexkalmuk: I do not see serios differences between these two test cases. As I understand you are testing two cases: len == page_size and len < page_size. In both cases, you should get NULL. So you just can propose united test case for len <= page_size and test both variants inside it.

United variant:

allocator =  page_allocator_init(buff, 0x100, 0x100);
test_assert_null(allocator);

allocator =  page_allocator_init(buff, 0x10, 0x100);
test_assert_null(allocator);

alexkalmuk: But it is not critical, so we can leave as it is.

src/tests/mem/page_allocator.c
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * @date 12.07.2018
+ * @author Filipp Chubukov
+ */
+
+#include <embox/test.h>
+#include <mem/page.h>
+#include <mem/phymem.h>
+
+EMBOX_TEST_SUITE("page_alloc_init test");
```
alexkalmuk: page_allocator_init?

src/tests/mem/Mybuild
```diff
@@ -19,6 +19,14 @@ module page {
 	depends embox.framework.LibFramework
 }
 
+module page_allocator {
+	source "page_allocator.c"
+
+	depends embox.mem.page_api
+	depends embox.mem.phymem
```
alexkalmuk: This dependence is also redundant now. You can remove it.

alexkalmuk:
Now it looks good for me. After squashing several commits, these changes could be merged into master.

PR #1177, opened 2018-07-10
Added an infomem command.

jackbogdanov:


src/drivers/common/memory.c
```diff
@@ -22,14 +22,12 @@
 #include <mem/mmap.h>
 #include <embox/unit.h>
 
+
```
Deryugin: Remove this empty line, please.

src/mem/phymem.c
```diff
@@ -44,10 +40,28 @@ static int phymem_init(void) {
...
+    extern char _reserve_end;
+
+    return (char *) binalign_bound((uintptr_t) &_reserve_end, PAGE_SIZE());
+}
+
+char *const phymem_allocated_end(void) {
+	extern char _ram_base;
+	extern char _ram_size;
+
+    return (char *)
```
Deryugin: Replace spaces with tabs at there two lines, please.

src/mem/vmem/vmem_alloc.c
```diff
@@ -19,6 +19,7 @@
 #include <mem/vmem/vmem_alloc.h>
 #include <util/binalign.h>
 
+
```
Deryugin: Remove this empty line, please.

src/cmds/mem/infomem.c
```diff
@@ -0,0 +1,94 @@
...
+
+#include <mem/vmem.h>
+#include <drivers/common/memory.h>
+
+static void print_vmem(struct page_allocator *pgd_allocator,
+		struct page_allocator *pmd_allocator, struct page_allocator *pte_allocator) {
+	int is_on_vmem = vmem_mmu_enabled();
+
+	printf("VIRTUAL MEMORY:\n");
+	printf("IS_ON - %d\n", is_on_vmem);
```
Deryugin: I recommend to use something like `printf("Virtual memory is %s\n", is_on_vmem ? "on" : "off");`, it's more clear.

src/drivers/common/memory.h
```diff
@@ -11,15 +11,24 @@
...
 #include <stdint.h>
 
+#define MAX_SEGMENTS 64
+
 struct periph_memory_desc {
 	uint32_t start;
 	uint32_t len;
 };
 
+struct _segment {
```
Deryugin: This struct has almost the same content as `struct periph_memory_desc`, are you sure it's necessary to add one more structure?

src/drivers/common/memory.c
```diff
@@ -100,4 +98,15 @@ static int periph_memory_init(void) {
 	return 0;
 }
 
+void periph_description(struct _segment **buff, int *size) {
+
+	for (int i = 0; i < MAX_SEGMENTS; i++) {
+		if (!_segments[i].start) {
+			*size = i;
```
Deryugin: This size is constant and can be obtained with `ARRAY_SPREAD_SIZE(__periph_mem_registry)`. Also it can be just a return value of this function, so no need in the second argument.

mk/image.lds.S
```diff
@@ -30,7 +30,7 @@ SECTIONS {
 
 		*(.text)
 		*(.text.*)
-
+		_text_end = .;
```
alexkalmuk: I am not sure you actually need these _text_end, _rodata_len and _data_end since for all these sections starting address and length are already defined. That is, to get the end of .text you can  use something like this &_text_vma + &_text_len.

mk/image.lds.S
```diff
@@ -92,15 +94,23 @@ SECTIONS {
 
 		ALIGNMENT();
 		_reserve_end = .;
+		_bss_reserve = _reserve_end - _bss_end;
```
alexkalmuk: The same is for this. Once you have _bss_vma, _bss_len, and _bss_len_with_reserve, you have no need to define something else, because it would be redundant. You can derive end address and length of both bss and bss_with_reserve in the C code just using these three.

mk/image.lds.S
```diff
@@ -92,15 +94,23 @@ SECTIONS {
...
 
 	_bss_len = _bss_end - _bss_vma;
 	_bss_len_with_reserve = SIZEOF(.bss);
 
+	_heap_vma = 0;
+	_heap_end = 0;
+	_heap_len = 0;
 #if HAS_SECTION_REGION(heap)
 	.heap : ALIGN(DATA_ALIGNMENT) {
+		_heap_vma = .;
```
alexkalmuk: Unfortunately, as I can see, the section .heap is completely unused. The actual heap is just a part of the .bss..reserve section and can be found here src/mem/heap/heap.lds.S There also HEAP_START and HEAP_END are defined, so you can use these instead on head_vma etc.

alexkalmuk: It's my mistake, this section is used for static heap on stm32.

jackbogdanov: I can't understand, how I can use HEAP_START and HEAP_END variables, becouse they defined by macros OPTION_GET(STRING,start_symbol) and I don't know real name of this variables.
Should I add knew variables, describing heap start and end, or exist a way to solve this problem better?

alexkalmuk: You can use in your file something like this
```
#define HEAP_START OPTION_MODULE_GET(embox__mem__static_heap,STRING,start_symbol)
extern char HEAP_START;
```
But it is only for one heap implementation (static_heap).

I'd propose to leave this as is and print only text, data, rodata, and bss. The problem with heap  could be deferred for the future :)

src/drivers/common/memory.c
jackbogdanov: Fixed

src/mem/phymem.c
jackbogdanov: Fixed

src/mem/vmem/vmem_alloc.c
jackbogdanov: Fixed

src/cmds/mem/infomem.c
jackbogdanov: Fixed

src/drivers/common/memory.h
jackbogdanov: Ok, I removed the structure _segment from this file.

src/drivers/common/memory.c
jackbogdanov: Fixed

src/mem/phymem.c
```diff
@@ -19,20 +19,16 @@ EMBOX_UNIT_INIT(phymem_init);
...
+	char *const phymem_alloc_start = phymem_allocated_start();
+	char *const phymem_alloc_end = phymem_allocated_end();
 	const size_t mem_len = phymem_alloc_end - phymem_alloc_start;
 	void *va;
 
 	log_boot_start();
 	log_boot("start=%p end=%p size=%zu\n", phymem_alloc_start, phymem_alloc_end, mem_len);
 
-	va = mmap_device_memory(phymem_alloc_start,
+
```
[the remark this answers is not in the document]
jackbogdanov: Fixed

src/mem/phymem.c
```diff
@@ -19,20 +19,16 @@ EMBOX_UNIT_INIT(phymem_init);
...
+	va = mmap_device_memory (phymem_alloc_start,
```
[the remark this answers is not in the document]
jackbogdanov: Fixed

mk/image.lds.S
jackbogdanov: Ok, I removed this variables

mk/image.lds.S
jackbogdanov: Fixed

mk/image.lds.S
jackbogdanov: Ok, I removed all information about heap.

src/drivers/common/memory.h
```diff
@@ -11,15 +11,20 @@
 
 #include <stdint.h>
 
+#define MAX_SEGMENTS 64
```
alexkalmuk: There is no need to define MAX_SEGMENTS in the header. You can move it to memory.c

src/mem/phymem.c
```diff
@@ -44,10 +39,28 @@ static int phymem_init(void) {
 	return phymem_alloc_start == va ? 0 : -EIO;
 }
 
+char * const phymem_allocated_start(void) {
+	extern char _reserve_end;
+
+    return (char *) binalign_bound((uintptr_t) &_reserve_end, PAGE_SIZE());
```
alexkalmuk: Something wrong with tabs/spaces in this function.

mk/image.lds.S
```diff
@@ -92,6 +91,7 @@ SECTIONS {
 
 		ALIGNMENT();
 		_reserve_end = .;
+
```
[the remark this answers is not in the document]
jackbogdanov: Done.

src/mem/heap/heap_simple.c
```diff
@@ -6,7 +6,6 @@
  * @author Nikolay Korotky
  */
 
-
```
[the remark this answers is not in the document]
jackbogdanov: Done.

src/drivers/common/memory.c
```diff
@@ -21,15 +21,16 @@
...
 #include <mem/mmap.h>
 #include <embox/unit.h>
+#include <util/array.h>
 
 ARRAY_SPREAD_DEF(struct periph_memory_desc *, __periph_mem_registry);
 
 EMBOX_UNIT_INIT(periph_memory_init);
 
-static struct _segment {
+struct _segment {
```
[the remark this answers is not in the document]
jackbogdanov: Fixed.

src/drivers/common/memory.h
jackbogdanov: Fixed.

src/mem/phymem.c
jackbogdanov: Fixed.

src/cmds/mem/infomem.c
```diff
@@ -0,0 +1,92 @@
...
+		struct page_allocator *pmd_allocator, struct page_allocator *pte_allocator) {
+	int is_on_vmem = vmem_mmu_enabled();
+
+	printf("VIRTUAL MEMORY:\n");
+	printf("vmem is %s\n", is_on_vmem ? "ON" : "OFF");
+
+	if (is_on_vmem) {
+		if (pte_allocator != NULL) {
+			printf("max pages count / free - %llu / %zu\n",
+					(unsigned long long) (pgd_allocator->pages_n * pmd_allocator->pages_n) * pte_allocator->pages_n,
```
alexkalmuk: I am not sure the total count of pages is a multiple of these three values because of the following reasons. 1. These allocators do not allocate the pages, they allocate tables for the pages. That is, each single page allocated by the allocator, represents a table with multiple entries, which then used to get physical address and page's attributes. 2. These allocator do not form a tree: you cannot allocate  pmd_allocator->pages_n for every pgd entry, you can allocate only pmd_allocator->pages_n at all.

I'd propose to print something like this:
```
printf("PGD tables count / free - %llu / %zu\n", pgd_allocator->pages_n,
          pgd_allocator->free / pgd_allocator->page_size);
printf("PMD tables count / free - %llu / %zu\n", pmd_allocator->pages_n,
          pmd_allocator->free / pmd_allocator->page_size);
if (pte_allocator) {
    printf("PTE tables count / free - %llu / %zu\n", pte_allocator->pages_n,
             pte_allocator->free / pte_allocator->page_size)
} else {
    printf("PTE do not available in this configuration\n");
}
```

P.S.
If and when you have time you can investigate virtual memory tables structure by looking at src/mem/vmem and especially at vmem_mapper.c

alexkalmuk: Yes, of course, there is a way to calculate the actual number of the mapped pages, and you can investigate how to do that further.

jackbogdanov: Ok, thank you, I'm not right understand the allocators meaning.
I fixed output.

Also there is a somthing strange with PGD tables: number of free tables is more than number of tables at all. (2048 / 2049 - all / free).

alexkalmuk: Well, I tried infomem on arm/qemu and reproduced the bug. After looking a bit more in depth, gdb shows me that there were happened 2 page_alloc's and 3 page_free on pgd_allocator. The last page_free worked well and incremented allocator->free up to the maximum available space + page_size. Apparently, it's bug in bimask.c and somewhere else (in there who try free non-allocated page). So the output of your command is well.

Thank you, I will add an issue.

alexkalmuk: I created the issue - https://github.com/embox/embox/issues/1185

src/cmds/mem/Infomem.my
```diff
@@ -0,0 +1,18 @@
...
+		SYNOPSIS
+			infomem
+		DESCRIPTION
+
+		AUTHOR
+			Evgenij Bogdanov
+	''')
+module infomem {
+	source "infomem.c"
+}
```
alexkalmuk: Probably you need dependencies on what you're using. `depends embox.mem.phymem` etc.

alexkalmuk:
Now it looks good for me. I believe the merge is coming soon enough.

@Deryugin Denis, do you have any comments and/or suggestions?

src/cmds/mem/infomem.c
```diff
@@ -0,0 +1,97 @@
...
+	printf("BSS:    begin - %p, end - %p, len - %p\n        "
+			"reserve - %p, end_with_reserve - %p, len_with_reserve - %p\n",
+			&_bss_vma, &_bss_end, &_bss_len,
+			-(int) &_bss_end + &_reserve_end,
+			&_reserve_end, &_bss_len_with_reserve);
+
+	printf("-----------------------------------------\n\n");
+}
+
+static void print_segments() {
```
alexkalmuk: `static void print_segments(void)`

src/cmds/mem/infomem.c
```diff
@@ -0,0 +1,97 @@
...
+		struct periph_memory_desc desk;
+		periph_elem_desc(&desk, i);
+		printf("seg_num %d: start - %x, end - %x, len - %x\n",
+				i, desk.start, desk.start + desk.len, desk.len);
+	}
+
+	printf("-----------------------------------------\n");
+}
+
+int main() {
```
alexkalmuk: `int main(int argc, char **argv)`

alexkalmuk: Please, add a simple help `infomem -h` like it is done in other commands. It will help you to understand why these arguments are required.

src/cmds/mem/Infomem.my
jackbogdanov: Done.

src/cmds/mem/infomem.c
jackbogdanov: Fixed.

src/cmds/mem/infomem.c
jackbogdanov: Ok, I added help and some other options to the command.

src/cmds/mem/infomem.c
```diff
@@ -0,0 +1,129 @@
...
+			return 0;
+		case 'v':
+			print_vmem(get_pgd_allocator(), get_pmd_allocator(), get_pte_allocator());
+			return 0;
+		case 'i':
+			print_segments();
+			return 0;
+		default:
+			printf("unknown argument!\n");
+			return 0;
```
alexkalmuk: Since it's an unknown parameter, you should return `-1` to indicate failure. https://www.gnu.org/software/coreutils/manual/coreutils.html#Exit-status

src/cmds/mem/infomem.c
```diff
@@ -0,0 +1,129 @@
...
+	print_phymem(phy_allocator());
+	print_vmem(get_pgd_allocator(), get_pmd_allocator(), get_pte_allocator());
+	print_segments();
+	print_sections();
+}
+
+int main(int argc, char **argv) {
+	int opt;
+
+	while (-1 != (opt = getopt(argc, argv, "h:p:s:v:i"))) {
```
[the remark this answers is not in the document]
jackbogdanov: Oh, I pushed wrong version of file, sorry..

src/cmds/mem/infomem.c
jackbogdanov: Fixed.

src/cmds/mem/infomem.c
```diff
@@ -0,0 +1,129 @@
...
+	printf("-----------------------------------------\n");
+}
+
+static void print_phymem(struct page_allocator *allocator) {
+	char *const phymem_alloc_start = phymem_allocated_start();
+	char *const phymem_alloc_end = phymem_allocated_end();
+	const size_t mem_len = phymem_alloc_end - phymem_alloc_start;
+
+	printf("\nPHYSICAL MEMORY:\n");
+	printf("start=%p end=%p size=%zu\n",
```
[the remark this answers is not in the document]
jackbogdanov: Fixed.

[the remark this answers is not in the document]
alexkalmuk: Thank you.

src/drivers/common/memory.c
```diff
@@ -100,4 +103,13 @@ static int periph_memory_init(void) {
 	return 0;
 }
 
+int periph_mem_size() {
+	return ARRAY_SPREAD_SIZE(__periph_mem_registry);
+}
+
+void periph_elem_desc(struct periph_memory_desc *elem, int num) {
```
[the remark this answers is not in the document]
jackbogdanov: I think, that the problem was caused by devision into MMU_PAGE_SIZE, during _segments[] init: variables were round, so we couldn't restore correct len using them.

Fixed that by using __periph_mem_registry instead of _segments[] array.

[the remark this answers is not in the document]
alexkalmuk: You are right, periph_desc corresponds to the memory region, which the corresponding peripheral device has requested, while _segments correspond to the _pages_ that will be allocated for these regions. I think now it's well done.

src/drivers/common/memory.h
```diff
@@ -16,10 +16,14 @@ struct periph_memory_desc {
 	uint32_t len;
 };
 
+#define MAX_SEGMENTS 64
```
alexkalmuk: Since you made this macro global, you should assign a unique name to it. Let's call it PERIPH_MAX_SEGMENTS, for example.

src/drivers/common/memory.c
```diff
@@ -100,4 +101,16 @@ static int periph_memory_init(void) {
 	return 0;
 }
 
+int periph_desc(struct periph_memory_desc **buff) {
+	struct periph_memory_desc *elem;
+
+	int i = 0;
+	array_spread_foreach(elem, __periph_mem_registry) {
+		buff[i] = elem;
+		i++;
```
alexkalmuk: By the way, you can just write `buf[i++] = elem;`. That is, `i` will be post-incremented immediately after assignment.

src/drivers/common/memory.c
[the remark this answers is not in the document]
alexkalmuk: By the way, it would be nice to print also the name of the corresponding memory segment. But let's defer this task for the future.

alexkalmuk:
@jackbogdanov Well, I have no more significant remarks, let's prepare the branch for merge.
There are 13 commits in your branch, many of them are fixes and small improvments of infomem command. Let's squash them all into a single commit with a commit message "cmds: Add infomem command". You can read about `git rebase -i` and `git squash` in google, for example here http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

src/cmds/mem/infomem.c
```diff
@@ -0,0 +1,128 @@
...
+	if (is_on_vmem) {
+		printf("PGD tables count / free - %zu / %zu\n", pgd_allocator->pages_n,
+					pgd_allocator->free / pgd_allocator->page_size);
+		printf("PMD tables count / free - %zu / %zu\n", pmd_allocator->pages_n,
+					pmd_allocator->free / pmd_allocator->page_size);
+		if (pte_allocator) {
+		    printf("PTE tables count / free - %zu / %zu\n", pte_allocator->pages_n,
+		    		pte_allocator->free / pte_allocator->page_size);
+		} else {
+		    printf("PTE do not available in this configuration\n");
```
alexkalmuk: https://travis-ci.org/embox/embox/jobs/409956601

Please remove whitespaces and add tabs instead.

jackbogdanov: Fixed.

src/cmds/mem/infomem.c
```diff
@@ -0,0 +1,128 @@
...
+	printf("vmem is %s\n", is_on_vmem ? "ON" : "OFF");
+
+	if (is_on_vmem) {
+		printf("PGD tables count / free - %zu / %zu\n", pgd_allocator->pages_n,
+					pgd_allocator->free / pgd_allocator->page_size);
+		printf("PMD tables count / free - %zu / %zu\n", pmd_allocator->pages_n,
+					pmd_allocator->free / pmd_allocator->page_size);
+		if (pte_allocator) {
+			printf("PTE tables count / free - %zu / %zu\n", pte_allocator->pages_n,
+		    		pte_allocator->free / pte_allocator->page_size);
```
alexkalmuk: Extra whitespaces still remained at this line. Change these whitespaces with tab. And squash the fix in your previous commit, please.

jackbogdanov: Fixed.

alexkalmuk:
@jackbogdanov Nice! Finally, please add your command into arm/qemu template (make a separate commit for this)

PR #1181, opened 2018-07-19
Wc util

FilippChubukov:


src/cmds/wc.c
```diff
@@ -0,0 +1,252 @@
...
+	}
+
+	if (argv[1] == NULL) {
+		printf("You need to specify the parameter(file)\n");
+		return 0;
+	}
+
+	
+	while (-1 != (opt = getopt(argc, argv, "lLcw"))) {
+		switch(opt) {
```
anton-bondarev: default?

src/cmds/wc.c
```diff
@@ -0,0 +1,252 @@
...
+	if (argc == 0) {
+		return -EINVAL;
+	}
+
+	if (argv[1] == NULL) {
+		printf("You need to specify the parameter(file)\n");
+		return 0;
+	}
+
+
```
anton-bondarev: Remove extra line

src/cmds/wc.c
```diff
@@ -0,0 +1,252 @@
...
+			break;
+
+			case 'w':
+			print_words = true;
+			param_number++;
+			break;
+			
+		}
+	}
+
```
anton-bondarev: extra line

src/cmds/wc.c
```diff
@@ -0,0 +1,252 @@
...
+	
+	
+	if (! (print_lines || print_words || print_bytes
+         || print_linelength))
+		print_lines = print_words = print_bytes = true;
+
```
anton-bondarev: extra lines

src/cmds/wc.c
```diff
@@ -0,0 +1,252 @@
...
+	if (argv[1] == NULL) {
+		printf("You need to specify the parameter(file)\n");
+		return 0;
+	}
+
+	
+	while (-1 != (opt = getopt(argc, argv, "lLcw"))) {
+		switch(opt) {
+			case 'l':
+			print_lines = true;
```
anton-bondarev: it's a bad formate style for switch case

src/cmds/wc.c
```diff
@@ -0,0 +1,252 @@
...
+	
+	
+	if (! (print_lines || print_words || print_bytes
+         || print_linelength))
+		print_lines = print_words = print_bytes = true;
+
+
+
+	
+	if (argc-param_number-1 == 1) {
```
anton-bondarev: this case is really required? whether If argc > 1 the loop from else branch show all required parameters or not .

src/cmds/wc.c
```diff
@@ -0,0 +1,252 @@
...
+}
+
+static int number_of_words(char *name) {
+	int   byte_file = 0;
+	int   count = 0;
+     
+	bool  prev_symbol = false;
+	bool  next_symbol = false;
+     
+	FILE*    file_in = fopen(name, "rt");
```
anton-bondarev: it's better if file is opened only once

src/cmds/wc.c
```diff
@@ -0,0 +1,167 @@
...
+static int bytes_in_file = 0;
+static int max_line_in_file = 0;
+
+static int wc_util(FILE* file_in) {
+	int byte_walker;
+	int count_len;
+	bool  prev_symbol = false;
+	bool  next_symbol = false;
+	lines_in_file = 0;
+        words_in_file = 0;
```
sikmir: You are mixing tabs and spaces for indention here and bellow.

src/cmds/wc.c
```diff
@@ -0,0 +1,167 @@
...
+			
+		if (print_linelength) {
+			printf("%d  ", max_line_in_file);		
+			if (max_line < max_line_in_file) max_line = max_line_in_file;
+		}									 
+
+		printf("%s \n", file_name);
+	}
+
+	if (argc-param_number-1 > 1) {
```
sikmir: (argc-param_number > 2)?

sikmir: (argc - param_number > 2)?

src/cmds/wc.c
```diff
@@ -0,0 +1,167 @@
...
+		} else { 
+			if (byte_walker == '\t') {
+				count_len = count_len + 8; 
+
+			} else {
+				count_len++;
+			}
+		}
+
+		next_symbol = !isspace(byte_walker)  ?  true  :  false; // words
```
sikmir: next_symbol = !isspace(byte_walker);

src/cmds/wc.c
```diff
@@ -0,0 +1,167 @@
...
+            		lines_in_file++;	// lines
+
+			if (max_line_in_file < count_len) {	// max line
+				max_line_in_file = count_len;
+			}
+
+			count_len = 0;
+		} else { 
+			if (byte_walker == '\t') {
+				count_len = count_len + 8;
```
sikmir: count_len += 8;

src/cmds/wc.c
```diff
@@ -0,0 +1,167 @@
...
+	if (argv[1] == NULL) {
+		printf("You need to specify the parameter(file)\n");
+		return 0;
+	}
+
+	while (-1 != (opt = getopt(argc, argv, "lLcw"))) {
+		switch(opt) {
+			case 'l':
+				print_lines = true;
+				param_number++;
```
sikmir: You can move param_number++; out of switch.

src/cmds/wc.c
```diff
@@ -0,0 +1,170 @@
...
+	if (! (print_lines || print_words || print_bytes
+         || print_linelength))
+		print_lines = print_words = print_bytes = true;
+
+	int lines_count = 0;
+	int words_count = 0;
+	int bytes_count = 0;
+	int max_line = 0;
+
+	int lines_in_file;
```
Deryugin: These four variables are used only inside the following loop, so it's better to declare it inside the loop.

src/cmds/wc.c
```diff
@@ -0,0 +1,170 @@
...
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+#include <stdlib.h>
+#include <mem/misc/pool.h>
+#include <stdbool.h>
+#include <util/log.h>
+#define OBJECTS_QUANTITY 0x10
+
+POOL_DEF(pool, int, OBJECTS_QUANTITY);
```
Deryugin: Pool is not used, so no need to declare it and include <mem/misc/pool.h>

src/cmds/wc.c
```diff
@@ -0,0 +1,170 @@
...
+	int max_line = 0;
+
+	int lines_in_file;
+	int words_in_file;
+	int bytes_in_file;
+	int max_line_in_file; 
+	
+	for (int j = param_number + 1; j <= argc - 1; j++) {
+		file_name = argv[j];
+		FILE*    file_in = fopen(file_name, "rt");
```
Deryugin: What does "t" char mean in mode string? I think just "r" should be enough for this case.

src/cmds/Wc.my
```diff
@@ -0,0 +1,32 @@
...
+			-L 
+				dispays the lenght of longest line
+
+			(without options displays number of line, words and bytes)
+		AUTHORS
+			Filipp Chubukov
+	''')
+module wc {
+	source "wc.c"
+
```
Deryugin: Remove these two empty lines, please.

src/cmds/wc.c
```diff
@@ -0,0 +1,170 @@
...
+	}
+	return 0;
+}
+
+int main(int argc, char **argv) { 
+	char *file_name;
+	int opt, param_number;
+	param_number = 0;
+
+	static bool print_lines, print_words,print_bytes;
```
Deryugin: Declaring these variables as `static` can cause error.

Static functions and static local variables is not the same:
* Static funcions are only visible in current translation unit
* Static local variables are allocated statically (i.e. they are placed to .bss section, not on the stack), and are initialized once during OS boot. So they will have old value if you run command one more time.

Deryugin: Also, add a space after the comma between second and third variables, please.

src/cmds/wc.c
```diff
@@ -0,0 +1,170 @@
...
+	static bool print_linelength;
+	
+	print_lines = print_words = print_bytes = false;
+  	print_linelength = false;
+
+	if (argc == 0) {
```
Deryugin: `argc` generally can not be zero if OS is working correctly. You can replace this condition and the following with one, like

```
if (argc < 2) {
    printf("No file name given.\n");
}
```

src/cmds/Wc.my
```diff
@@ -0,0 +1,32 @@
+package embox.cmd
+
+@AutoCmd
+@Cmd(name = "wc",
+	help = "wc util(copies the functionality of wc from GNU coreutils)",
+	man = '''
+		NAME
+			wc - wc util(copies the functionality of wc from GNU coreutils)
+		SYNOPSIS
+			wc FILE_NAME
```
Deryugin: It's actually `wc [OPTIONS] [FILE_NAME]`.

It's important because in this implementation you really assume that file names are going just after the options (if they will be mixed like `wc file1 -c file2` it will fail).

src/cmds/wc.c
FilippChubukov: this check was used in another command, so I decided that it was needed

PR #1184, opened 2018-07-21
net: Improve ip_rcv if inet_dev isn't set

anton-bondarev:


src/net/l3/ipv4/ip_input.c
```diff
@@ -98,7 +98,11 @@ static int ip_rcv(struct sk_buff *skb, struct net_device *dev) {
 
 	/* Forwarding */
 	assert(skb->dev);
-	assert(inetdev_get_by_dev(skb->dev));
+	if (!inetdev_get_by_dev(skb->dev)) {
+		skb_free(skb);
```
Deryugin: May be it worth to add `log_debug()` for this case? It can be hard to debug if such packet is lost.

PR #1186, opened 2018-07-30
More command

FilippChubukov:


src/cmds/fs/more.c
```diff
@@ -17,71 +17,62 @@
...
-							memset(buff + y, ' ', TAB_SIZE);
-							y += TAB_SIZE - 1;
-							break;
+	lines = std->endy - std->begy;
+	endwin();
+
+	while (1) {
+		for (x = 0; x < lines - 1; x++) {
+
+			for (i = 0; i < columns; i++) {
```
Deryugin: Does it actually matter? If string is shorter than `columns` then according `buff[i > columns]` are not used at all.

src/cmds/fs/more.c
```diff
@@ -17,71 +17,62 @@
...
+					y += TAB_SIZE - 1;
+					break;
 			}
-			memset(info, '\0', columns);
-			memcpy(info, "---More---", 10);
-			memset(buff, ' ', columns);
-			if (endOfFile) {
-				strcat(info, " End of the file! Press 'q' to quit.");
+		}
+			/*	In case if we got out of the actual line size	*/
```
Deryugin: Fix identation here, please.

Deryugin: I mean for two '{' in the upper lines.

src/cmds/fs/more.c
```diff
@@ -17,71 +17,62 @@
 #define TAB_SIZE 4
 #define MAX_SCREEN_WIDTH 256
 
-static char buff[MAX_SCREEN_WIDTH], info[MAX_SCREEN_WIDTH];
+static char buff[MAX_SCREEN_WIDTH];
```
Deryugin: This variable is used only in `screen()`, move it there.

src/cmds/fs/more.c
```diff
@@ -17,71 +17,62 @@
...
+
+			for (i = 0; i < columns; i++) {
+				buff[i] = ' ';
+			}
+
+			for (y = 0; y < columns; y++) {
+				buff[y] = getc(fp);
+				switch ((int) buff[y]) {
+				case EOF:
+					buff[y] = ' ';
```
Deryugin: You don't actually use `buff[y]`, so no need to set it.

src/cmds/fs/more.c
```diff
@@ -17,71 +17,62 @@
...
+					for (i = 0; i < y; i++) {
+						printf("%c", buff[i]);
+					}
+					return;
+				case '\n':
+					buff[y] = ' ';
```
Deryugin: Why did you remove 'memset()` and comments here and in the next `case`? It seems to be ok.

src/cmds/fs/more.c
FilippChubukov: It was before and doesn't affect, i can remove it.

PR #1187, opened 2018-08-22
Fix static irq

alexkalmuk:
Extend the common part of trap table for all cortex-m for both static and dynamic irqs. It is required bacause we want to have informative exceptions handlers for any Embox configuration. So we just extend the trap table.

### DIFF
--- a/src/arch/arm/armmlib/exception_table.S
+++ b/src/arch/arm/armmlib/exception_table.S
@@ -30,9 +30,13 @@ trap_table_start:
     .word 0 /* Reserved */
     .word 0 /* Reserved */
     .word 0 /* Reserved */
+#if 0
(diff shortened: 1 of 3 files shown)


src/arch/arm/armmlib/exception_table.S
```diff
@@ -30,9 +30,6 @@ trap_table_start:
     .word 0 /* Reserved */
     .word 0 /* Reserved */
     .word 0 /* Reserved */
-    .word exp_default_entry /* SVCall */
```
anton-bondarev: May be comment this
#if 0 
/* we use static_irq_attach for these handlers */
#endif

alexkalmuk: Thanks, I added the corresponding comment.

PR #1188, opened 2018-08-23
Monocube port

anton-bondarev:
Continue porting to 'monocub'

- Improve AM85c30 serial driver
- Enable console
- Add sm750 video driver
- Add PCI chip
- Start working on IRQ


anton-bondarev:
@alexkalmuk Yes, I'm agree, it's a right way. We will do this in future, but now it requires too much time.

PR #1198, opened 2018-09-05
Pjsip qemu

alexkalmuk:
pjsip_simpleua is now working on qemu x86 (exactly the same thing as for stm32f4discovery)

P.S.
I cannot get record back to life and test it well, so I leave record as it is for now.


src/drivers/audio/portaudio/portaudio_lib.c
```diff
@@ -221,6 +222,8 @@ PaError Pa_OpenStream(PaStream** stream,
 		return paInvalidDevice;
 	}
 
+	audio_dev->ad_ops->ad_ops_ioctl(audio_dev, ADIOCTL_SET_RATE, &rate);
```
anton-bondarev: May be add ADIOCTL_GET_RATE and change a rate only if current rate differs from required?

alexkalmuk: Thanks, I will add it.

PR #1199, opened 2018-09-06
Robots pthread

alexkalmuk:
* Use pthreads in platform/stm32f4_multibots instead of kernel threads.
* Make SPI module which implements *Msp_Init function as non-static, it is required to overwrite weak declarations in STM32 Cube


platform/stm32f4_multibots/cmds/robot/main.c
```diff
@@ -170,7 +169,7 @@ static void robot_config_motors(struct stepper_motor *motor1,
 
 static int robot_config_gy_30(int i2c_nr, int mode) {
 	if (gy_30_init(i2c_nr) < 0) {
-		log_error("gy_30_init failed\n");
+		printf("gy_30_init failed\n");
```
Deryugin: Is it necessary to minimize build?

alexkalmuk: No, it's just because it's command. I think it's better do not use our log_* in user commands.

alexkalmuk: Of course, this function can be used in our commands, but here printf looks a bit better as for me. It's normal fix or do you propose to revert this change back to log_error?

PR #1202, opened 2018-09-13
cortex-m: Add additional information about exceptions

alexkalmuk:
Improve cortex-m exception info: Print out also MemManage and Bus fault address registers

### DIFF
--- a/src/arch/arm/armmlib/exception_handlers.c
+++ b/src/arch/arm/armmlib/exception_handlers.c
@@ -16,14 +16,30 @@
 #define SCB_BASE 0xe000ed00
 #define SCB_CONF_FAULT_STATUS (SCB_BASE + 0x28)
 #define SCB_HARD_FAULT_STATUS  (SCB_BASE + 0x2C)
+#define SCB_MEM_FAULT_ADDRESS  (SCB_BASE + 0x34)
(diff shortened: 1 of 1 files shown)


src/arch/arm/armmlib/exception_handlers.c
```diff
@@ -16,14 +16,30 @@
...
 
-	printk("Memory Management Fault Status register = %x\n", conf_faults & 0xf);
-	printk("Bus Fault Status register = %x\n", (conf_faults >> 8) & 0xf);
-	printk("Usage Fault Status register = %x\n", (conf_faults >> 16) & 0xff);
-	printk("Hard Fault Status register = %lx\n", REG_LOAD(SCB_HARD_FAULT_STATUS));
+	mem_fault_status = conf_faults & 0xf;
+	bus_fault_status = (conf_faults >> 8) & 0xf;
+	usage_fault_status = (conf_faults >> 16) & 0xff;
+
+	printk("MemManage Fault Status register = %x\n", mem_fault_status);
```
Deryugin: I guess `MemManage Fault Status register` should be `MemManage Fault Status` here (because it's not a register), and the same stuff for following lines.

Also, I think it's better to use `0x%0Nx` format so it's clear that the value is hex.

PR #1204, opened 2018-09-16
Pjsip stm32f7discovery

alexkalmuk:
pjsip example works now on STM32F7-Discovery.

Some other contributions:
* Improve NVIC to make it work well with interrupt priorities
* Play command can play .wav from ROM
* Use 320K memory for STM32F7 instead of 256K as it was before.


src/cmds/play.c
```diff
@@ -14,10 +14,17 @@
...
 #include <util/math.h>
+#include <framework/mod/options.h>
+
+#define PLAY_FROM_FLASH OPTION_GET(BOOLEAN, play_from_flash)
+#define FRAMES_PER_BUFFER OPTION_GET(NUMBER, frames_per_buffer)
 
 static void print_usage(void) {
 	printf("Usage: play [WAVAUDIOFILE]\n"
-	       "       play -s\n");
+	       "       play -s\n"
```
Deryugin: I guess it's good idea to update manual page (in `Play.my`) for this command if you change the help message.

src/cmds/play.c
```diff
@@ -151,7 +187,11 @@ int main(int argc, char **argv) {
 
 		printf("Progress:\n");
 
-		_bl = min(fread(_fbuffer, 1, 64 * 1024 * 1024, fd), _bl);
+#if PLAY_FROM_FLASH
```
Deryugin: We can get rid of this `#ifdef` by moving `_bl = ...` to the previous `#ifdef` as it's not used between this macros.

src/cmds/play.c
```diff
@@ -83,7 +96,11 @@ static int fd_callback(const void *inputBuffer, void *outputBuffer,
 int main(int argc, char **argv) {
 	int opt;
 	int err;
+#if PLAY_FROM_FLASH
+	uint32_t audio_flash_addr;
```
Deryugin: Is this variable really necessary? I think `audio_start_ptr` can be used directly to parse arguments.

src/cmds/play.c
```diff
@@ -103,21 +120,39 @@ int main(int argc, char **argv) {
...
+	while (-1 != (opt = getopt(argc, argv, "nshf:"))) {
 		switch (opt) {
 		case 'h':
 			print_usage();
 			return 0;
 		case 's':
 			callback = &sin_callback;
 			break;
+		case 'f':
+#if !PLAY_FROM_FLASH
```
Deryugin: I think this option is ok even for big platforms. For example, this can be used in debug purpose to play audio from RAM.

src/cmds/play.c
```diff
@@ -14,10 +14,17 @@
 #include <drivers/audio/portaudio.h>
 #include <fs/file_format.h>
 #include <util/math.h>
+#include <framework/mod/options.h>
+
+#define PLAY_FROM_FLASH OPTION_GET(BOOLEAN, play_from_flash)
```
Deryugin: I think it's better to replace this option with something like `USE_LOCAL_BUFFER` and just fail it there's not local buffer, but user requests playing audio from file. This approach will reduce number of `#ifdef .. #else .. #endif`

alexkalmuk: Thank you, I will try something like this

alexkalmuk: We can implement two different functions - play_from_memory and play_from_file. What do you think? In this case we can get rid from this ifdef's. Because now we must maintain both buffers and files in the same function which leads to ifdef's...

alexkalmuk:
@Deryugin Thank you. I hope I introduced fixed which are close to your suggestions.

Deryugin:
Yes, now the code looks much better, thanks!

PR #1211, opened 2018-09-21
Improve system start service

anton-bondarev:
* Enable different ttys
* Enable output commands during starting


src/init/system_start_service.c
```diff
@@ -23,6 +24,16 @@ static const char *script_commands[] = {
...
+#if OPTION_GET(NUMBER,log_level) >= LOG_INFO
+#	define MD(x) do {\
+		x;\
+	} while (0);
+#else
+#	define MD(x) do{\
+	} while (0);
+#endif
+
+#define PRINTF printf
```
Deryugin: This macro is not used

src/init/system_start_service.c
```diff
@@ -32,6 +43,7 @@ int system_start(void) {
 	setup_tty(OPTION_STRING_GET(tty_dev));
 
 	array_foreach(command, script_commands, ARRAY_SIZE(script_commands)) {
+		MD(printf(">%s\n", command));
```
Deryugin: May be `log_boot()` can be used here?

src/init/system_start_service.c
anton-bondarev: Thanks, fixed

src/init/system_start_service.c
anton-bondarev: No, I don't think so. it's a first user process, it should use printf in my mind

templates/x86/qemu/mods.config
```diff
@@ -94,8 +94,8 @@ configuration conf {
 	@Runlevel(1) include embox.test.util.hashtable_test
 
 	@Runlevel(2) include embox.cmd.sh.tish(prompt="%u@%h:%w%$", rich_prompt_support=1, builtin_commands="exit logout cd export mount umount")
-	@Runlevel(2) include embox.init.start_script(shell_name="tish", tty_dev="ttyS0", shell_start=1, stop_on_error=true)
-	include embox.init.system_start_service
+	//@Runlevel(2) include embox.init.start_script(shell_name="tish", tty_dev="ttyS0", shell_start=1, stop_on_error=true)
```
Deryugin: Is it necessary to leave this line?

PR #1213, opened 2018-09-25
ARM exception handlers

Deryugin:
1) Improve ARM unresolvable exception handler. Now it's able to print error even if stack pointer or even MMU tables are corrupted.
2) Improve C-part of exception handler. Now it has sanity check for arguments.
3) Now data abort handler uses separate C-part.


src/arch/arm/armlib/exceptions/exceptions.S
```diff
@@ -121,14 +122,36 @@ prefetch_abt_handler:
...
+
+data_abt_handler:
+	/* Same as with arm_unresolvable_exception we need
+	 * to make sure that RAM access will be safe */
+	mrc p15, 0, r0, c1, c0, 0
+	and r0, r0, #~CR_M
+	mcr p15, 0, r0, c1, c0, 0
+	ldr sp, =_stack_top
+	nop
+	stmfd r13!, {r0-r12,r14}
```
alexkalmuk: Just a small remark. Probably to use the same notation as above? sp instead of r13?

src/arch/arm/armlib/exceptions/exceptions.S
```diff
@@ -121,14 +122,36 @@ prefetch_abt_handler:
...
 	b arm_exception_handler
+
+data_abt_handler:
+	/* Same as with arm_unresolvable_exception we need
+	 * to make sure that RAM access will be safe */
+	mrc p15, 0, r0, c1, c0, 0
+	and r0, r0, #~CR_M
+	mcr p15, 0, r0, c1, c0, 0
+	ldr sp, =_stack_top
+	nop
```
alexkalmuk: nop added for safety? Can we remove it or something can go wrong?

PR #1215, opened 2018-09-25
A bunch of fixes for memory-related functions

Deryugin:
The most significant change is implementation for Idesc-specific `mmap()`.

### DIFF
--- a/src/tests/mem/mmap.c
+++ b/src/tests/mem/mmap.c
@@ -7,20 +7,20 @@ EMBOX_TEST_SUITE("anonymous mmap");
 
 TEST_CASE("allocate one page") {
 	void *ptr = mmap(NULL, PAGE_SIZE(), PROT_READ | PROT_WRITE, MAP_SHARED | MAP_ANONYMOUS, -1, 0);
-	test_assert_not_null(ptr);
+	test_assert_not_equal(ptr, MAP_FAILED);
 	test_assert_zero(munmap(ptr, PAGE_SIZE()));
 }
 
 TEST_CASE("request a non-whole page") {
 	int len = PAGE_SIZE() + 1;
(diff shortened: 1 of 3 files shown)


src/tests/mem/mmap.c
```diff
@@ -7,20 +7,20 @@ EMBOX_TEST_SUITE("anonymous mmap");
...
 	void *ptr = mmap(NULL, len, PROT_READ | PROT_WRITE, MAP_SHARED | MAP_ANONYMOUS, -1, 0);
-	test_assert_not_null(ptr);
+	test_assert_not_equal(ptr, MAP_FAILED);
 	test_assert_zero(munmap(ptr, len));
 }
 
 TEST_CASE("request zero-length") {
 	void *ptr = mmap(NULL, 0, PROT_READ | PROT_WRITE, MAP_SHARED | MAP_ANONYMOUS, -1, 0);
-	test_assert_null(ptr);
+	test_assert_equal(ptr, MAP_FAILED);
```
anton-bondarev: Check errno?

src/tests/mem/mmap.c
```diff
@@ -36,7 +36,7 @@ static void *mmap_test_child(void *arg) {
 	int allocated = 0;
 	while (1) {
 		void *ptr = mmap(NULL, PAGE_SIZE(), PROT_READ | PROT_WRITE, MAP_SHARED | MAP_ANONYMOUS, -1, 0);
-		if (!ptr)
+		if (ptr != MAP_FAILED)
```
anton-bondarev: check errno?

alexkalmuk: It was `if (!ptr)`. So I think it should be `ptr == MAP_FAILED`, not `ptr != MAP_FAILED`.
The test logic, as I can see, suppose that we break only when all pages (per task) were already mapped.

PR #1220, opened 2018-09-26
MESA/Etnaviv

Deryugin:
* Implement Etnaviv GPU driver
* Add limited support for MESA with Etnaviv gallium dirver
* Add templates for MESA + Etnaviv on i.MX6
* Refactoring/fixes/other minor improvements


platform/etnaviv/cmds/cube_cmd.h
```diff
@@ -0,0 +1,841 @@
+
```
anton-bondarev: May be change file extention from .h to .inc because it looks very strange header file

Deryugin: Fixed it.

PR #1222, opened 2018-09-28
Cpp14 templates fix

anton-bondarev:
Add CXXFLAGS += -Wno-error=c++14-compat for some templates

### DIFF
--- a/templates/arm/omap-qt/build.conf
+++ b/templates/arm/omap-qt/build.conf
@@ -1,7 +1,7 @@
 
 TARGET = embox
 
-PLATFORM = olimex
+PLATFORM = SECO-Q7-796
 
 ARCH = arm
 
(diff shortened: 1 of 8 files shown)


templates/arm/omap-qt/mods.config
```diff
@@ -1,10 +1,12 @@
 package genconfig
 
 configuration conf {
+	@Runlevel(0) include embox.arch.arm.armlib.high_excpt_table
 	@Runlevel(0) include embox.arch.generic.arch
-	@Runlevel(0) include embox.arch.arm.mmu_test
+	//@Runlevel(0) include embox.arch.arm.mmu_test
```
alexkalmuk: If this template without mmu then we can simply remove mmu_test, there is no need to comment it out.

templates/arm/omap-qt/build.conf
```diff
@@ -14,5 +14,9 @@ CXXFLAGS += -march=armv7-a -mtune=cortex-a8
 CXXFLAGS += -fno-rtti
 CXXFLAGS += -fno-exceptions
 CXXFLAGS += -fno-threadsafe-statics
+CXXFLAGS += -fsized-deallocation
```
alexkalmuk: The same flag is commented out 2 line below..

PR #1223, opened 2018-09-28
arm: Add support for fast bss clear & data copy

Deryugin:
Now `Embox kernel start` is printed almost immediately after boot.


src/arch/arm/head.S
```diff
@@ -4,53 +4,123 @@
...
+	mcr p15, 0, r0, c2, c0, 0
+
+	/* Map whole address space 1-1 */
+	mov r1, #0
+	mov r2, #0x1000
+l1:
+	/* Build supersection desc in R3 */
+	orr r3, r1, # 0x6
+	orr r3, r3, # 0xC00
+	#orr r3, r3, # (1 << 18)
```
anton-bondarev: It's required or the same as line above?

Deryugin: Removed.

PR #1224, opened 2018-09-29
Add sort util

oniich:
Add sort util that works like standard GNU sort. It uses mergesort algorithm and has -b -f and -r options.


src/cmds/fs/sort.c
```diff
@@ -0,0 +1,178 @@
...
+    while (k <= mid)
+        buff_strs[i++] = in_strs[k++];
+    while (l <= end)
+        buff_strs[i++] = in_strs[l++];
+
+    for(i = begin; i <= end; ++i)
+        in_strs[i] = buff_strs[i];
+}
+
+static void sort(char **in_strs, int begin, int end, char **buff_strs) {
```
alexkalmuk: @oniich First of all, thank you for the contribution!
Now, as far as I can see you implemented Merge Sort inside the command. But it is better to use some existing stuff if it is presented. In your case, you could try libc/stdlib/qsort.c. It's from standard C library. If it will work, it would be a better choice.
In that case all you need is to use qsort, and write your own string comparator for it.

src/cmds/fs/sort.c
```diff
@@ -0,0 +1,166 @@
...
+		"\nOther options:\n\n"
+		"  -h\tdisplay this help and exit\n"
+		"\n*** WARNING ***\n"
+		"The locale specified by the environment affects sort order.\n"
+		"This implementation assumes that LC_ALL=C set to get the\n"
+		"traditional sort order that uses native byte values.\n"
+	);
+}
+
+static int compare(const void *fst_ptr, const void *snd_ptr) {
```
alexkalmuk: It looks like a sort of string comparison. If it's so, can I ask to try `strcmp` instead?

src/cmds/fs/sort.c
```diff
@@ -0,0 +1,166 @@
...
+		}
+		while (snd[j] == ' ') {
+			++j;
+		}
+	}
+
+	while (fst[i] != '\0' && snd[j] != '\0') {
+		if (ign_case) {
+			if (
+				((64 < fst[i] && fst[i] < 91) || (96 < fst[i] && fst[i] < 123)) &&
```
anton-bondarev: Try to don't use magic constant. Extract definitions of it through macros

src/cmds/fs/sort.c
oniich: I need to compare strings with different conditions depending on flags used, which is not supported by `strcmp`. If it's needed though, I can use `strcmp` when there is no flags. But I'm not sure this will speed up comparision process.

alexkalmuk: Yes, I understand, I think your solution is working well. But my proposition is concerning code simplification. Look, instead of `while (fst[i] == ' ')` you can use `while (isspace(fst[i])`, because the second variant prоcesses tabs and whitespaces, while the first one processes only white spaces.

Now, after skipping white spaces you have two options - usual string comparison, or comparison ignoring case (-f). So in case if option `-f` is specified, you can use `strncasecmp`, which compares two strings ignoring case. In case if `-f` was not specified, you can use usual `strcmp`.
What do you think about it? Probably, it can do this job?

alexkalmuk: But in any case, it's just a proposition to think about it, not a requirement.

oniich: Thanks for advice. I totally forgot about tabs, but they're need to be ignored with `-b` option. So I will use `isspace`. And I'll check if `strcasecmp` suits `-f` option, because it may be tricky how `sort` compares with case insensitivity.

oniich: `isspace` works like a charm. `strcmp` suits my needs too. Problems appears with `strcasecmp`. It compares strings with lowering case of characters. But `sort -f` folds lower case to upper case. So sort results differs. If it's not vital to follow up GNU sort implemintation, I can leave things as it is. If needed, I can do GNU-like solution. What do you think?

alexkalmuk: > It compares strings with lowering case of characters. But sort -f folds lower case to upper case. So sort results differs.

Hm, it's interesting.. Could you, please, paste here an example of two strings when `sort -f` and `strcasecmp` will get different results? I am not sure about all the languages in the world, but for English and many other European languages there should be no matter which case (lower or upper) is used during string comparision.. Are you using English language in your test files?

alexkalmuk: By the way, since `sort` requires converting to upper case, could we firstly convert the string to upper case using `toupper` character by character, and then apply `strcmp` to the obtained strings?

oniich: Yes, I'm using English language. I suppose, it's not language problem.
Here's input `atest` file:
```
Aab
AA
b
aa
AAb
Aa
B
aab
aA
aAb
```
Standard GNU sort ─ `LC_ALL=C sort -f atest`:
```
AA
Aa
aA
aa
AAb
Aab
aAb
aab
B
b
```
My sort with `strcasecmp` as string comparator ─ `./sort -f atest`:
```
AA
aa
Aa
aA
Aab
AAb
aab
aAb
b
B
```
Problem here is that for `strcasecmp` "AA" == "Aa". Because of that, output line's position depends on position in input file.
GNU sort does it's own magic. For him "AA" < "Aa" < "aA" and so on. As I figured out it happens because with `-f` option, `sort` folds lower case to upper case, but still thinks that "A" < "a". And what important here, "a" become less than "B". So comparision of ASCII characters will be like that:
```
... < '?' < '@' < 'A' < 'a' < 'B' < 'b'  < ... < 'Z' < 'z' < '[' < '\' < ...
```
Therefore, `strcasecmp` here is useless and `strcmp` with `toupper()` here is useless too (because will be "AA" == "Aa" again). So I guess for `-f` option I will leave my solution (it works somehow).

alexkalmuk: Well, so it seems gnu `sort` preserves the original sorting order among equal strings. Yes, let's keep your solution here.

After your last commit, `compare` looks pretty well! Can you, please, add a small comment in the code explaining that we do that because gnu sort works in the following way with `-f` option: (e.g. Input: `aa AA`, output: `AA aa`)

alexkalmuk: P.S.
An alternative solution could be:
```
    if (ign_case) {
        int ret;
        ret = strcasecmp(fst, snd);
        return ret ? ret : strcmp(fst, snd);
    } else {
```

oniich: Oh, it looks like your solution is working, didn't thought about that, thanks.
Btw, I've got new problem. Now GNU sort does some magic with `-b` option.
Here's `test` file (`|` character represents beginning of the line):
```
|b
| B
|AA
|B
|aa
| Aa
| b
|    aA
|    B
```
`./sort -b test`:
```
|AA
| Aa
| B
|B
|    B
|    aA
|aa
|b
| b
```
`LC_ALL=C sort -b test`:
```
|AA
| Aa
|    B
| B
|B
|    aA
|aa
| b
|b
```
Thing here is that GNU sort always try to sort. In my implementation "[space]B" == "B" == "[tab]B". Therefore they're placed like in input file. 
For GNU sort "[tab]B" < "[space]B" < "B". So they're sorted like that.
It has logic in that. Even with "equal" lines one still can expect somehow sorted output. Don't know if that implementation was necessary or just for eyecandy. Anyway, it seems like I'll need little more time to solve this.

src/cmds/fs/sort.c
```diff
@@ -32,33 +32,48 @@ static void print_usage(void) {
 	);
 }
 
-static int compare(const void *fst_ptr, const void *snd_ptr) {
+/* 	This function similar to "strcasecmp" except for "tolower" changed to "toupper",
+	because "sort -f" folds lower case to upper case characters */
+static int case_compare(const char *str1, const char *str2) {
```
oniich: I figured out that `strcasecmp` still can't do the job well with `-f` option, when there're characters between 'Z' and 'a' (like '[', '\\', ']' and so on). GNU `sort -f` treats `'z' < '['` and `'A' < '['`, while for `strcasecmp` `'[' < 'z'`  and `'[' < 'A'`. It happens because `strcasecmp` folds chars to lower case, while GNU sort folds chars to upper case. So I've rewrited it with just changing `tolower` to `toupper`.

src/cmds/fs/sort.c
```diff
@@ -0,0 +1,173 @@
...
+		"\n*** WARNING ***\n"
+		"The locale specified by the environment affects sort order.\n"
+		"This implementation assumes that LC_ALL=C set to get the\n"
+		"traditional sort order that uses native byte values.\n"
+	);
+}
+
+/* 	This function similar to "strcasecmp" except for "tolower" changed to "toupper",
+	because "sort -f" folds lower case to upper case characters */
+static int case_compare(const char *str1, const char *str2) {
```
oniich: I figured out that `strcasecmp` still can't do the job well with `-f` option, when there're characters between 'Z' and 'a' (like '[', '\\', ']' and so on). GNU `sort -f` treats `'z' < '['` and `'A' < '['`, while for `strcasecmp` `'[' < 'z'`  and `'[' < 'A'`. It happens because `strcasecmp` folds chars to lower case, while GNU sort folds chars to upper case. So I've rewrited it with just changing `tolower` to `toupper`.

alexkalmuk: Yes, it's more correct.

PR #1234, opened 2018-10-18
Fix -O2 builds

Deryugin:
* Fix `cmd/fs/wc`
* Fix a bunch of issues related to -O2 + GCC 8

### DIFF
--- a/src/cmds/net/arping.c
+++ b/src/cmds/net/arping.c
@@ -110,9 +110,9 @@ int main(int argc, char **argv) {
 		return -EINVAL;
 	}
 
-	strncpy(dst_b, inet_ntoa(dst), sizeof(dst_b));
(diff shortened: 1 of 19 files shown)


src/cmds/net/arping.c
```diff
@@ -110,9 +110,9 @@ int main(int argc, char **argv) {
 		return -EINVAL;
 	}
 
-	strncpy(dst_b, inet_ntoa(dst), sizeof(dst_b));
+	strncpy(dst_b, inet_ntoa(dst), sizeof(dst_b) - 1);
```
anton-bondarev: it should set '\0' at the last byte of the buffer after calling strncpy
`dst_b[sizeof(dst_b) - 1] = '\0';`

PR #1237, opened 2018-10-24
ramfs

Deryugin:
Simple implementation of ramfs for new vfs.

It can use any block device as backend, but file metadata is always stored in RAM.

### DIFF
--- a/src/fs/driver/ramfs/ramfs_dvfs.c
+++ b/src/fs/driver/ramfs/ramfs_dvfs.c
@@ -1,0 +1,401 @@
+/**
+ * @file
+ * @brief Tmp file system
+ *
+ * @date 12.11.12
+ * @author Andrey Gazukin
+ */
+
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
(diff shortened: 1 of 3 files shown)


src/fs/driver/ramfs/ramfs_dvfs.c
```diff
@@ -0,0 +1,401 @@
...
+
+#include <embox/unit.h>
+#include <drivers/block_dev.h>
+
+/* define sizes in 4096 blocks */
+#define MAX_FILE_SIZE   OPTION_GET(NUMBER, ramfs_file_size)
+#define RAMFS_FILES     OPTION_GET(NUMBER, inode_quantity)
+#define FILESYSTEM_SIZE (MAX_FILE_SIZE * RAM_FILES)
+
+#define RAMFS_NAME_LEN	32
```
anton-bondarev: May be use NAME_MAX from limits.h
[https://serverfault.com/questions/9546/filename-length-limits-on-linux](https://serverfault.com/questions/9546/filename-length-limits-on-linux)

PR #1240, opened 2018-11-01
posix: Break cyclic deps for module exit

anton-bondarev:


src/compat/posix/proc/exit/Mybuild
```diff
@@ -0,0 +1,9 @@
+
+package embox.compat.posix.proc
+
```
alexkalmuk: Extra empty line.

Issue #1185, opened 2018-07-27
pgd_allocator free more than available space


alexkalmuk:
Template: arm/qemu

include embox.cmd.mem.infomem

Then run infomem command.

"Also there is a somthing strange with PGD tables: number of free tables is more than number of tables at all. (2048 / 2049 - all / free)."

PR #1244, opened 2018-11-07
Fix of pgd_allocator frees more than available space.

oniich:
The problem was that in the `src/mem/mmap/mmap_mmu.c:132` called `mmap_clear(mmap);` and `vmem_free_context(mmap->ctx);`. 
This functions leads to the call of `src/mem/vmem/vmem_alloc.c:vmem_free_pgd_table` 2 times, which causes to free 2 pages instead of 1.
Fix was to check if current page already freed or not.


oniich:
Also it fixes issue https://github.com/embox/embox/issues/1185

PR #1250, opened 2018-11-21
Update template for x86/qemu

oniich:
Increased stack size


Deryugin:
Please change commit message, it's too ambiguous.

Add template name as a prefix.

PR #1253, opened 2018-11-23
Stm32 fix static irq

alexkalmuk:
Fix `irq_static` to make test `embox.test.kernel.thread.thread_systimer_irq_test` work for both static and non-static vector table. Nevertheless, the old version of irq_static remained, since it is useful when we do not use threads, but only light threads. Now, this module is renamed to `irq_static_light`.

### DIFF
--- a/src/drivers/audio/stm32/stm32_audio_f7.c
+++ b/src/drivers/audio/stm32/stm32_audio_f7.c
@@ -233,7 +233,7 @@ static irq_return_t stm32_audio_in_dma_interrupt(unsigned int irq_num,
 }
 
 static_assert(86 == STM32_AUDIO_IN_DMA_IRQ);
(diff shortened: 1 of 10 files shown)


src/drivers/audio/stm32/stm32_audio_f7.c
```diff
@@ -233,7 +233,7 @@ static irq_return_t stm32_audio_in_dma_interrupt(unsigned int irq_num,
 }
 
 static_assert(86 == STM32_AUDIO_IN_DMA_IRQ);
-STATIC_IRQ_ATTACH(STM32_AUDIO_IN_DMA_IRQ, stm32_audio_in_dma_interrupt, NULL);
+STATIC_IRQ_ATTACH(86, stm32_audio_in_dma_interrupt, NULL);
```
anton-bondarev: STM32_AUDIO_IN_DMA_IRQ looks better, in my mind

PR #1256, opened 2018-12-04
Stm32 improve irq

alexkalmuk:
Move cortexm interrupt controller to a separate folder and remove redundant memcpy


src/drivers/interrupt/cortexm/cortexm_nvic.c
```diff
@@ -178,18 +177,14 @@ extern void __pending_handle(void);
...
-		memcpy(&state->ctx, ctx, BASE_CTX_SIZE);
-		state->misc[0] = regs->lr;
-		state->misc[1] = regs->sp;
-	} else {
-		memcpy(&state->ctx, ctx, FPU_CTX_SIZE);
-		state->misc[18] = regs->lr;
-		state->misc[19] = regs->sp;
-	}
+	/* Save LR and SP below the context. */
+	offset = interrupted_from_fpu_mode(regs->lr) ? FPU_CTX_SIZE : 0;
```
Deryugin: There's no `memcpy()` now, is it correct?

alexkalmuk: Yes, since we do not need registers values that were before the interrupt happened.

alexkalmuk: We do not need them inside interrupt handling

alexkalmuk:
@Deryugin @anton-bondarev 

ping

src/drivers/interrupt/cortexm/cortexm_nvic.c
Deryugin: Did you check it on STM32 board? If it works, I think it's ok

alexkalmuk: Yes, I have checked that before and checked right now after `git rebase`

PR #1257, opened 2018-12-05
SPI subsystem

Deryugin:
**Add SPI subsystem**
imx6_ecspi is compatible with this interface.
Old test for i.Wave G15S SPI is rewritten for new interface.

**Rework char_dev for new VFS**
* Remove boilerplate for devfs-related init of char dev (no more files ops for char devices, there's no need in this)
* Improve devfs by using more common code for char/block devices. Remove `struct device_module`, use `struct dev_module` instead (the same structure is used for block devices)
* Fix read/write for char devices (previously this could not be done as vfs tried to truncate them)
* Minor chages (rename some variables, clean unused code etc)


src/drivers/char/char_dev_dvfs/char_dev_dvfs.c
```diff
@@ -10,20 +10,76 @@
...
+	cdev->dev_id = cdev_id;
+
+	if (cdev->dev_file.f_ops == NULL) {
+		cdev->dev_file.f_ops = &char_dev_fops;
+	}
 
 	return 0;
 }
+
+int char_dev_idesc_fstat(struct idesc *idesc, void *buff) {
```
anton-bondarev: This is global?

src/drivers/char/char_dev_dvfs/char_dev_dvfs.c
```diff
@@ -10,20 +10,76 @@
...
+			log_error("Failed to open %s", devmod->name);
+			return NULL;
+		}
+	}
+
+	return &devmod->dev_file.f_idesc;
+}
+
+struct file_operations char_dev_fops = {
+	.open = char_dev_open,
```
anton-bondarev: There is '.fstat' member?

src/drivers/common/device.c
```diff
@@ -35,6 +35,10 @@ struct dev_module *dev_module_create(struct device *dev, const char *name,
 	assert(privdata); /* No known devices without privdata. Currently
 			   *  it's up to legacy of the old FS */
 
+	while (*name == '/') {
```
anton-bondarev: *name++ ?

src/drivers/char/char_dev_dvfs/char_dev_dvfs.c
Deryugin: Yes.

src/drivers/char/char_dev_dvfs/char_dev_dvfs.c
Deryugin: No, `.fstat` is used in `idesc_ops`, not `file_operations`.

src/fs/dvfs/dvfs.h
```diff
@@ -28,6 +28,7 @@
 #define DVFS_DIR_VIRTUAL   0x01000000
 #define DVFS_CHILD_VIRTUAL 0x02000000
 #define DVFS_MOUNT_POINT   0x04000000
+#define DVFS_NO_LSEEK      0x08000000
```
anton-bondarev: It looks not very good

src/drivers/common/device.c
Deryugin: This way it will cause unwanted side-effect when `*name != '/'`.

src/fs/dvfs/dvfs.h
Deryugin: Why? Macro name looks pretty self-explanatory.

src/drivers/char/char_dev_dvfs/char_dev_dvfs.c
Deryugin: I added some comments to explain how this functions is supposed to be used.

src/drivers/spi/Mybuild
```diff
@@ -1,6 +1,15 @@
 
 package embox.driver.spi
 
+module core {
+	option number log_level=0
+
+	@IncludeExport(path="drivers")
+	source "spi.h"
```
alexkalmuk: Where are `spi.h` and `spi.c`? I do not see these file in the PR

src/drivers/char/char_dev_dvfs/char_dev_dvfs.c
```diff
@@ -10,20 +10,78 @@
...
-DLIST_DEFINE(cdev_repo_list);
+#define MAX_DEV_QUANTITY OPTION_GET(NUMBER, dev_quantity)
 
-ARRAY_SPREAD_DEF(const struct device_module, __char_device_registry);
+static struct dev_module *devtab[MAX_DEV_QUANTITY];
+INDEX_DEF(char_dev_idx, 0, MAX_DEV_QUANTITY);
+
+ARRAY_SPREAD_DEF(const struct dev_module, __char_device_registry);
+
+struct dev_module **get_cdev_tab() {
```
alexkalmuk: get_cdev_tab(void)?

src/drivers/spi/Mybuild
Deryugin: Forgot to add it, check last commits.

src/drivers/char/char_dev_dvfs/char_dev.h
```diff
@@ -9,25 +9,45 @@
...
-
-extern struct dlist_head cdev_repo_list;
+#include <util/dlist.h>
+#include <util/macro.h>
+#include <util/member.h>
+
+extern struct file_operations char_dev_fops;
+
+#define CHAR_DEV_DEF(chname, idesc_op, open_fn, priv) \
+	struct dev_operations MACRO_GUARD(dev_ops) = { \
```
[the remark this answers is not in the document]
Deryugin: Fixed.

src/drivers/common/device.c
alexkalmuk: I don't understand why. 
```
    while (*name++ != '/')
        ;
```
works well when there are '/' in name at any place.

By the way, in any case, what if the name do not contain '/'? Or this function assumes it does anyway?

src/drivers/char/char_dev_dvfs/char_dev_dvfs.c
Deryugin: Fixed

src/fs/dvfs/dvfs.h
anton-bondarev: It's not a node flag, better if it's be a driver specific, in my mind

anton-bondarev: It's only for the future

src/drivers/spi/spi.h
```diff
@@ -0,0 +1,64 @@
...
+
+struct spi_ops {
+	int (*init)(struct spi_device *dev);
+	int (*select)(struct spi_device *dev, int cs);
+	int (*transfer)(struct spi_device *dev, uint8_t *in, uint8_t *out, int cnt);
+};
+
+extern struct spi_device *spi_dev_by_id(int id);
+extern int spi_transfer(struct spi_device *dev, uint8_t *in, uint8_t *out, int cnt);
+extern int spi_transfer(struct spi_device *dev, uint8_t *in, uint8_t *out, int cnt);
```
alexkalmuk: Two identical declarations of `spi_transfer`

src/drivers/common/device.c
Deryugin: Code you provided doesn't work well, that's exactly why another way is used :)

Just compile following code and look at the output
```
#include <stdio.h>

char *trim1(char *name) {
	while (*name == '/') { name++; }
	return name;
}

char *trim2(char *name) {
	while (*name++ == '/');
	return name;
}

int main() {
	printf("%s\n%s\n%s\n%s\n",
		trim1("/123"),
		trim2("/123"),
		trim1("123"),
		trim2("123"));
	return 0;
}
```

Deryugin: @alexkalmuk it's `while (*name++ == '/')`, not `while (*name++ != '/')`

alexkalmuk: My mistake

src/drivers/spi/spi.c
```diff
@@ -0,0 +1,175 @@
...
+ARRAY_SPREAD_DECLARE(const struct dev_module, __char_device_registry);
+
+static int spi_init(void) {
+	struct dev_module *dev_module;
+	struct spi_device *sdev;
+
+	/* Note: this currently is neccessary only because
+	 * common probe() is not implemented for serial devices */
+	array_spread_foreach_ptr(dev_module, __char_device_registry) {
+		if (dev_module->device->dev_iops != &spi_iops) {
```
alexkalmuk: So SPI now depends on dvfs in any way?

src/fs/dvfs/dvfs.h
Deryugin: When this flag is used in vfs layer, there's no need to implement stubs for `truncate` calls when this calls have no sense (for example, for char devices).

src/drivers/spi/spi.h
Deryugin: Fixed.

src/drivers/spi/spi.c
Deryugin: Yes, it depends on devfs.

Probably, separate modules can be used for devfs-related stuff and general SPI functions. But this leads SPI devices be **not** a specific case for char devices (i.e. we should duplicate char_dev pool with spi_dev pool and so on).

Look at the last commit, I splitted source into two files. If it will be just two modules, then for non-devfs module we will have no interface to get a device in any way. I'm not sure if it's ok.

Deryugin: But here's another question: are there any configs which do not use FS at all?

alexkalmuk: Since we didn't use SPI well yet (only 1-2 platforms), it's hard to say... But anyway, after your last commit it looks better to me.

Deryugin: Check out new version for SPI module, please

src/drivers/spi/spi.h
```diff
@@ -0,0 +1,63 @@
...
+#include <util/array.h>
+#include <util/macro.h>
+
+struct spi_ops;
+
+struct spi_device {
+	int    id;
+	int    cs;
+	int    flags;
+	struct dev_module *dev;
```
alexkalmuk: In case if we do not use fs, `dev_module` will be presented anyway?

src/drivers/spi/spi.c
alexkalmuk: Now it looks good to me :)

src/drivers/spi/spi.c
```diff
@@ -0,0 +1,81 @@
...
+			log_warning("SPI%d has no init funcion", (*dev)->id);
+		}
+	}
+
+	return 0;
+}
+
+EMBOX_UNIT_INIT(spi_init);
+
+struct spi_device *spi_dev_by_id(int id) {
```
alexkalmuk: `spi_dev_by_id` searches for device without comparing with its spi_device->id, as I can see. But what if I have two spi devices with id's 0 and 1, but they were added to spi_device_registry in the reverse order somehow (SPI_DEF(spi1), then SPI_DEF(spi0))? Then when I write spi_dev_by_id(0), it's possible situation when I can get second device, not the first one?

alexkalmuk: But how I can make sure `spi_dev_by_id` returned _right_ spi controller, since the enumeration within ARRAY_SPREAD depends on linker?

anton-bondarev: @alexkalmuk  It's not so significant now
@Deryugin In i2c I set number of i2c controller (bus) statically, and during i2c module init check are the numbers not cross between each other. It's rather useful to understand which real controller used

Deryugin: Check out last commits pls

PR #1261, opened 2018-12-07
telnet: Fix to ignore commands

oniich:
Fix to avoid junk symbols to be executed by tish after telnet connection.
This code assume that telnet commands get sent by separate packets (without any data) and drops them.


src/cmds/net/telnetd/telnetd.c
```diff
@@ -23,6 +23,7 @@
 
 #include <util/math.h>
 
+#include <kernel/printk.h>
```
alexkalmuk: Where `printk` is required?

src/cmds/net/telnetd/telnetd.c
```diff
@@ -331,6 +332,12 @@ static void *telnetd_client_handler(void* args) {
 				MD(printf("read on sock: %d %d\n", sock_data_len, errno));
 			}
 			sock_data_len = telnet_fix_crnul(s, sock_data_len);
+
+			/* TODO: Check if there's T_IAC in string and delete it and
+					 following commands, instead of dropping whole message */
+			if (strchr((const char *)s, T_IAC) != NULL)
```
alexkalmuk: Probably, it would be useful to add at least `log_debug` here, to notify about options skipping.

src/cmds/net/telnetd/telnetd.c
oniich: In 173 row:
`printk("system return error: %d\n", ret);`

src/cmds/net/telnetd/telnetd.c
alexkalmuk: I am not sure any raw data from socket can be interpreted as a string since maybe it's not a null-terminated string..

src/cmds/net/telnetd/telnetd.c
alexkalmuk: Ah, you just moved it to another place, now I see

src/cmds/net/telnetd/telnetd.c
oniich: Yes, I've changed it.

alexkalmuk: Thank you, if telnet works, then it looks to me.

PR #1266, opened 2018-12-17
Unify char_dev interfaces

anton-bondarev:
Make char_dev_register and CHAR_DEV_DEF similar  interface
In ideal, keep single implementation for oldfs and dvfs and char_dev_stub (char_dev without fs)


src/drivers/tty/serial/ttys_dvfs.c
```diff
@@ -58,7 +62,7 @@ int ttys_register(const char *name, void *dev_info) {
 	memset(cdev, 0, sizeof(*cdev));
 	memcpy(cdev->name, name, sizeof(cdev->name));
 	cdev->dev_priv = dev_info;
-	cdev->dev_file.f_ops = &ttys_fops;
+	// cdev->dev_file.f_ops = &ttys_fops;
```
alexkalmuk: I think this probably doesn't work. You have to assign something for `open`.

src/fs/syslib/kfile.c
```diff
@@ -47,12 +47,13 @@ struct file_desc *kopen(struct node *node, int flag) {
...
 
-	if (node_is_file(node)) {
+	if (!(node_is_directory(node)) && !(node_is_file(node))) {
 		if (NULL == nas->fs->drv) {
 			SET_ERRNO(ENOSUPP);
 			return NULL;
 		}
-		ops = (struct kfile_operations *) nas->fs->drv->file_op;
+		extern struct idesc *char_dev_open(struct node *node, int flags);
+		return char_dev_open(node, flag);
```
alexkalmuk: I'm not sure but shouldn't we create here `file_desc` firstly and only then return idesc? I think so because in functions below (`kwrite` etc.) `file_desc` will be passed as an argument.

alexkalmuk: Probably it's ok, since we can guarantee `kwrite` is only used for usual files, right?. But it's not obvious because of the similar naming of `kopen`/`kwrite`. Assuming I called `kopen` I'm hoping to see `kwrite` somewhere.

And direct calling of `char_dev_open` here looks a bit weird..

oniich: `idesc` used instead of `file_desc` need some consideration.
Direct call of `char_dev_open` is needed because there's no other way to interact with `char_dev`

src/drivers/tty/serial/ttys_dvfs.c
oniich: This is new FS, that currently not used in x86/qemu build. I'm working on making old one as close as possible to new, so I could combine them in the end. This will be fixed eventually.

src/drivers/block_dev/block_dev_idesc_dvfs.c
```diff
@@ -237,3 +242,4 @@ int bdev_read_blocks(struct dev_module *devmod, void *buf, int blk, int count) {
 
 	return count * bdev->block_size;;
 }
+#endif
```
anton-bondarev: please add a new line

oniich: yep, sure

src/drivers/char/char_dev.c
```diff
@@ -15,27 +15,73 @@
...
+	struct dev_module *cdev = node->nas->fi->privdata;
+	struct idesc *idesc;
+
+	if (!cdev) {
+		log_error("Can't open char device");
+		return NULL;
+	}
+
+	if (cdev->open != NULL) {
+		idesc = cdev->open(cdev, cdev->dev_priv);
```
Deryugin: You can reduce it to `return cdev->open(cdev, cdev->dev_priv);`

src/drivers/char/char_dev.c
```diff
@@ -46,18 +92,18 @@ int char_dev_register(const char *name, const struct kfile_operations *ops) {
...
 	dev_nas = node.node->nas;
-	dev_nas->fs = filesystem_create("empty");
+	dev_nas->fs = filesystem_create("devfs");
 	if (dev_nas->fs == NULL) {
 		return -ENOMEM;
 	}
-
-	dev_nas->fs->file_op = ops;
+	
+	node.node->nas->fi->privdata = (void *) cdev;
```
Deryugin: No need to cast `cdev` to `void *` (for `void *` pointer of any type is ok, no warnings here).

src/drivers/char/char_dev_dvfs/char_dev.h
```diff
@@ -12,42 +12,26 @@
...
+extern int char_dev_init_all(void);
 extern int char_dev_register(struct dev_module *cdev);
+extern int char_dev_idesc_fstat(struct idesc *idesc, void *buff);
+extern struct idesc *char_dev_idesc_create(struct dev_module *cdev);
 
 static inline struct dev_module *idesc_to_dev_module(struct idesc *desc) {
 	struct file *f = mcast_out(desc, struct file, f_idesc);
 
-	return mcast_out(f, struct dev_module, dev_file);
+	return (struct dev_module *) f->f_inode->i_data;
```
Deryugin: No need to cast result to `(struct dev_module *)` (it's `void *`, so no warnings here).

src/drivers/common/device.h
```diff
@@ -22,34 +21,28 @@
...
 
-extern struct dev_module *dev_module_create(struct device *dev,
-	const char *name, void *privdata);
+extern struct dev_module *dev_module_create(
+	const char *name,
+	struct idesc * (*open)  (struct dev_module *, void *),
+	int 		   (*close) (struct idesc *),
+	const struct idesc_ops *dev_iops,
+	void *privdata
+);
```
Deryugin: I think it looks better if `);` is left on the same line with the last argument.

src/drivers/net/tun/tun.c
```diff
@@ -115,30 +116,39 @@ static inline struct net_device *tun_netdev_by_name(const char *name) {
...
-
-static inline int tun_netdev_by_desc(const struct file_desc *fdesc,
+static inline int tun_netdev_by_idesc(struct idesc *idesc,
 		struct net_device **netdev, struct tun **tun) {
-	*netdev = fdesc->file_info;
+	struct dev_module *cdev;
+
+	cdev = idesc_to_dev_module(idesc);
+
+	*netdev = (struct net_device *) cdev->dev_priv;
```
Deryugin: No need to cast `void *` to `(struct net_device *)`

src/drivers/serial/uart_dev_dvfs.c
```diff
@@ -130,23 +130,13 @@ static int uart_fill_name(struct uart *dev) {
 }
 
 extern const struct idesc_ops idesc_serial_ops;
-static struct device tty_device = {
-	.dev_iops = &idesc_serial_ops,
-};
 
-extern struct file_operations ttys_fops;
+extern int ttys_register(const char*name, void *dev_info);
```
Deryugin: `char*name` -> `char *name`

src/drivers/tty/serial/idesc_serial.h
```diff
@@ -14,4 +14,6 @@ struct uart;
 
 extern struct idesc *idesc_serial_create(struct uart *uart, mode_t mod);
 
+extern const struct idesc_ops *idesc_serial_get_ops();
```
Deryugin: If argument list is empty, function can get any number of arguments, so it's better to write `idesc_serial_get_opts(void)` explicitly.

src/drivers/char/char_dev.c
anton-bondarev: If it's used 'const' than it's required

PR #1269, opened 2018-12-18
Fix arm/stm32_vl_light

anton-bondarev:
Add thread_none to enable single task without threads and include it to arm/stm32_vl_light


src/kernel/thread/Mybuild
```diff
@@ -1,5 +1,11 @@
 package embox.kernel.thread
 
+
```
Deryugin: Double blank line

PR #1271, opened 2018-12-20
Fix ci fs

alexkalmuk:
* Enable interactive tests for `x86/test/fs`
* Add new bunch of tests like running unit tests multiple times, and shell commands tests

### DIFF
--- a/templates/x86/test/fs/mods.config
+++ b/templates/x86/test/fs/mods.config
@@ -49,7 +49,7 @@ configuration conf {
 	include embox.kernel.thread.signal.siginfoq
 
 	@Runlevel(1) include embox.test.block_dev.ramdisk_test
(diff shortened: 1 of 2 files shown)


src/tests/fs/concurrency/concurrency_test.c
```diff
@@ -0,0 +1,157 @@
...
+#include <drivers/block_dev/ramdisk/ramdisk.h>
+#include <drivers/block_dev.h>
+
+#include <pthread.h>
+
+#include <embox/test.h>
+
+EMBOX_TEST_SUITE("File system concurrency test suite");
+
+#define TEST_FS_NAME    "ramfs"
```
anton-bondarev: OPTION_GET and ramfs as default?

PR #1274, opened 2018-12-27
Malloc tests

alexkalmuk:
Add two tests for malloc. Both are currently failing.

```
    @Runlevel(1) include embox.test.mem.heap_extended_test
    @Runlevel(1) include embox.test.mem.heap_threads_test
```

* heap_extended_test tries to allocate all possible space within the kernel heap, counting how many malloc() call succeed. Then it free() all allocated objects and checks whether it can repeat the same operation once more - i.e. malloc the same number of objects. This test is currently failed.

* heap_threads_test checks whether two threads can malloc() simultaneously. This test currently finished with an assertion failure.


anton-bondarev:
I think It's not fully correct check the heap on kernel_task because kernel_task using if test run during starting the system, but if you use test command for running test, it will not use kernel_task. May be it is better to separate this test for 2 for sysmalloc and malloc + new_task (vfork) ?

alexkalmuk:
@anton-bondarev Thanks, I'm agree about the separation of the testing kernel_task/non kernel_task

PR #1277, opened 2019-01-15
Qt automatic testing

alexkalmuk:
It's an automatic Qt test. It works in the following way. In the beginning, a virtual display (`xvfb`) is created. Then `xvncviewer` is executed inside this display, and display is grabbing by `ffmpeg` at the same time. Then, we have a set of predefined gold frames, and we compare the grabbed video of `moveblocks` Qt application execution and compare it frame by frame with each of the gold frames. In case if _all_ gold frames were found in the grabbed video, we claim the test passed successfully.


alexkalmuk:
@AntonKozlov Hi! I would like to add qt testing and some additional packages like `xvnc4viewer`, `xvfb` and `ffmpeg` are required. As you can see, I'm using my own docker image which is mostly based on emdocker and ubuntu 16.04 instead of 14.04  - https://github.com/embox/embox/pull/1277/commits/35b05b9ae2eba0035cba07a31b8ee0e8a96d607b.
Can I update emdocker and emdocker-test with the corresponding changes?

.travis.yml
```diff
@@ -52,7 +53,10 @@ script:
   - set -e # make this script fail as soon as any individual command fail
   - . ./scripts/docker/docker_rc.sh
   - dr make confload-"$TEMPLATE"
-  - run dr make
+  - if [ "$TEMPLATE" == "x86/test/qt-vnc" ] || [ "$TEMPLATE" == "x86/test/fs" ];
+    then run dr 'make > /dev/null';
```
AntonKozlov: Are you sure we want drop build logs? I suspect build error will not be uncommon, changes to break these templates are even higher than others, considering number of modules included

scripts/continuous/run.sh
```diff
@@ -102,6 +103,13 @@ run_check() {
 			ret=0
 		fi
 	done
+
+	if [ "x$CONTINIOUS_SUCCESS_START_PATTERN" != "x" ] ; then
```
AntonKozlov: I'm not sure about this part: you could just include `QVNCServer created` into the list. Or refactor a line above to 
```
SUCESS_PATTERNS=(
^run: success auto poweroff' 'embox>' '[a-z]\+@embox'
)
if [ "x$CONTINIOUS_SUCCESS_START_PATTERN" != "x" ]
    SUCESS_PATTERNS += '$CONTINIOUS_SUCCESS_START_PATTERN'
fi
for success_pattern in "$CONTINIOUS_SUCCESS_START_PATTERN[@]"; do 
```
(code excerpt is untested)

AntonKozlov:
> I would like to add qt testing and some additional packages like `xvnc4viewer`, `xvfb` and `ffmpeg` are required. As you can see, I'm using my own docker image which is mostly based on emdocker and ubuntu 16.04 instead of 14.04 - [35b05b9](https://github.com/embox/embox/commit/35b05b9ae2eba0035cba07a31b8ee0e8a96d607b).
> Can I update emdocker and emdocker-test with the corresponding changes?

Of course, you can. Those images are built from https://github.com/embox/emdocker and https://github.com/embox/emdocker-test repos. To update the image, you just have to push changes to those repos (into master, directly or with PR, whatever you like). Docker Hub cloud will get notification and will re-build the image(s). This is the correct way to update images. The other option is to `docker push`, but I see no reasons to use it, so please do not until discussed. Sorry for being restrictive, but I see no Dockerfile's for your docker images (the image is totally opaque), so I'm just assertive :)

What are the reasons to change emdocker? emdocker-test is big pile of various stuff for testing, it's OK to bring anything there. But emdocker suggested to hold only stuff required for build and be used by humans, not only travis. Is something missing there, that is required to build qt-vnc template?

PS I'll add you into embox docker org admins

alexkalmuk:
@AntonKozlov 

> Sorry for being restrictive, but I see no Dockerfile's for your docker images (the image is totally opaque), so I'm just assertive :)

Yes, it's OK, I will create PR and you can review and then push by yourself :)

> What are the reasons to change emdocker?

The only reason that there are no `ffmpeg` in 14.04 repos.. So I decided it's not bad idea to use more new version of ubuntu.

.travis.yml
alexkalmuk: It's due to Travis log size limitations. If you just do `make`, then Travis failed because of log size exceeded.

scripts/continuous/qt/test_run.sh
```diff
@@ -0,0 +1,47 @@
...
+# to CAPTURED_VIDEO
+timeout $TIMEOUT ffmpeg -draw_mouse 0 -video_size 1024x768 -framerate 25 \
+	-f x11grab -i :$QT_VNC_DISPLAY_NR $CAPTURED_VIDEO &
+# Start xvncviewer on display :QT_VNC_DISPLAY_NR
+timeout $TIMEOUT xvncviewer -geometry 1024x768 10.0.2.16:5900
+
+# Now search for the gold frames within captured video.
+# If some of the gold frames are not found then exit with failure.
+final_ret=0
+for gold_frame in $(ls $GOLD_FRAMES_DIR); do
```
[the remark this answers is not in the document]
alexkalmuk: Yes, of course, I will fix it soon.

AntonKozlov:
> > Sorry for being restrictive, but I see no Dockerfile's for your docker images (the image is totally opaque), so I'm just assertive :)
> Yes, it's OK, I will create PR and you can review and then push by yourself :)
I should placed more emphasis it's a joke. I pushed without a review before, but now it's two of us collaboratting, so I'll have to send PRs too... (sigh) 
OK:)

> > What are the reasons to change emdocker?
> The only reason that there are no `ffmpeg` in 14.04 repos.. So I decided it's not bad idea to use more new version of ubuntu.

OK, having all other tests are passes on 16.04.

.travis.yml
alexkalmuk: The log size exceded initially for `qt-vnc` only. But after updating 14.04 up to 16.04 test/fs log becames larger too.

scripts/continuous/run.sh
alexkalmuk: Initially, I included `QVNCServer created` in the list, but then decided to use a bit more flexible way, since it's could be a common workaround for all applications running in kernel task.

But probably it's better to solve it simpler - just to include `QVNCServer created` in the list back like you propose.

AntonKozlov:
How about to store whole log and `tail` after make finished? (successfully
or not). We can output significant number of lines without bothering
travlis log limit then.

>

.travis.yml
AntonKozlov: (posting to correct thread)
How about to store whole log and `tail` after make finished? (successfully or not). We can output significant number of lines without bothering travlis log limit then.

scripts/continuous/run.sh
AntonKozlov: The current version is just feels like a workaround. Probably the whole success_pattern thing is wrong, as it require simple runner script to know all positive and negative outcomes. 
Can we ignore the error from qemu_bg and check for `QVNCServer created` in qt-vnc test script?

.travis.yml
alexkalmuk: Or we can echoing something like "Building..." every minute or smth. for example during the build (for Travis only).

AntonKozlov: Seems like we assume something is regularly printed to stderr?
The suggestion may work. It like it way more than dropping the output. I suppose, can be implemented for all templates? There is a quirk there, it would hide any parallel buld issues that may be resolved by second make. Then we'll have to make sure exit code of whole expression is the one from first make

alexkalmuk: I found [here](https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received) "If you have a command that doesn’t produce output for more than 10 minutes, you can prefix it with travis_wait, a function that’s exported by our build environment".
So probably we could use this, and the output last 3Mb bytes. I will check this a bit later.

alexkalmuk: I tried this in https://github.com/embox/embox/pull/1277/commits/d25066a155208e17b17d1be0251a2f3ef21b4b70. Seems to be working.

scripts/continuous/qt/test_run.sh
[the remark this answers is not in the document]
alexkalmuk: Fixed, thanks.

scripts/continuous/run.sh
alexkalmuk: I made it somehow by [this](https://github.com/embox/embox/pull/1277/commits/cdb95401bb1a9af21e0f5ea7f91713046150a6dd) and [this](https://github.com/embox/embox/pull/1277/commits/cbb035c53f4f122a7e6075fb8d9b8527a9dc3d4f).

alexkalmuk:
@AntonKozlov Thank you for reviewing!

> Docker Hub built new emdocker(-test), please update scripts

I updated scripts and now will check if the things work well.

PR #1278, opened 2019-02-04
Check if inetdev is registered in net_rx

alexkalmuk:
Sometimes when a network driver enables interrupts before inetdev_register_dev(), interrupt can happen and we go right to `net_rx` without any inet device registered yet.

### DIFF
--- a/src/net/l0/net_rx.c
+++ b/src/net/l0/net_rx.c
@@ -16,6 +16,7 @@
 #include <net/netdevice.h>
 #include <net/skbuff.h>
 #include <net/socket/packet.h>
+#include <net/inetdevice.h>
 #include <util/log.h>
 
 #define LOG_LEVEL OPTION_GET(NUMBER, log_level)
@@ -53,6 +54,16 @@ int net_rx(struct sk_buff *skb) {
 	case PACKET_BROADCAST:
 	case PACKET_MULTICAST:
 		break;
+	}
+
+	/* Sometimes when a network driver enables interrupts before
+	* inetdev_register_dev(), interrupt can happen and we go
+	* right here without any inet device registered yet. */
+	if (!inetdev_get_by_dev(skb->dev)) {
+		log_debug("inet device is not registered yet for %s\n",
+			skb->dev->name);
+		skb_free(skb);
+		return 0;
 	}
 
 	/* setup L3 header */


AntonKozlov:
Shouldn't it be considered as a bug in a driver that does such thing? It then assert looks more appropriate and the driver should be fixed

anton-bondarev:
@AntonKozlov No it's a correct way, If we want to use ethernet without IP settings.
There is a direct (point to point) line for example. Net driver has to pass all correct packets to net stack, and it must resolve collisions

alexkalmuk:
@AntonKozlov So I am a bit confused about what to choose - claim the hard interrupt enabling / device registering order for all network drivers, or add this `if` in net_rx..

AntonKozlov:
> @AntonKozlov You are right that it's a bug in the driver, but we have such bugs in all our network drivers. 

It would be great to estimate how much efforts will it take to fix drivers and is it possible at all. Unexpected interrupt may be an indicator of more deep problem of the driver. It may not only have not registered the device in net subsystem, the device itself may have not configured properly on that point

alexkalmuk:
In case if we would have separate `probe` and `open` for each network device we could to register our inetdevice in `probe` and then enable interrupts in `open`.

alexkalmuk:
> It may not only have not registered the device in net subsystem, the device itself may have not configured properly on that point

Good point, as I understand the logic of `probe` and `open` in Linux kernel, they separate these two logics from each other. `probe` only resets and _prepares_ device for further usage.

AntonKozlov:
Isn't `netdriver`'s `start` function serves for start receiving packets?
and what should `probe` do? By it's name it look like "detect if device is there". But how can be device be registered and but is not usable? It look like as such probe should be impelemented in a bus communicating to net device, and it's presented, i.e. in pci

alexkalmuk:
> Isn't `netdriver`'s `start` function serves for start receiving packets?

Ah, I see now, we already have that one

alexkalmuk:
@AntonKozlov So it's just a bug of mipsnet driver since interrupts are enabled in `init`, not `start`.

alexkalmuk:
Sorry, it looks like my mistake. I will create a new fix for `mipsnet` driver.

PR #1281, opened 2019-02-10
General changes from xen18 branch

AntonKozlov:
During work on xen port in branch `xen18` a bunch of changes was made. They are not related to xen itself. So making a separate PR for them.

### DIFF
--- a/src/cmds/sys/date.c
+++ b/src/cmds/sys/date.c
@@ -79,42 +79,52 @@ static void set_date(char *new_date) {
 	settimeofday(&tv, NULL);
 }
 
-static char *get_next_key(char *string) {
-	while(*string++) {
-		if('%' == *string) {
-			return ++string;
-		}
-	}
-	return NULL;
-}
-#define MAX_NANOSECONDS 999999999
-static int check_format(char *string) {
-	char *str;
-	if(string[0] != '+') {
-		printf("invalid format string use '+%%N'\n");
+static int show_fdate(char *fmt) {
+
+	if (fmt[0] != '+') {
+		printf("invalid format string, use '+FORMAT'\n");
 		return -1;
 	}
-	if(NULL != (str = get_next_key(string))) {
-		switch(str[0]) {
-		case 'N': {
-			struct timeval tv;
-			time64_t ns;
 
-			ktime_get_timeval(&tv);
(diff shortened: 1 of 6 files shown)


src/cmds/sys/date.c
```diff
@@ -79,42 +79,52 @@ static void set_date(char *new_date) {
...
+	char *c = fmt + 1;
+	while (*c) {
+		if (*c != '%') {
+			printf("%c", *c);
+			++c;
+			continue;
+		}
+
+		++c;
+		if (!c) {
```
Deryugin: Should it be `!*c`?

src/kernel/time/timer.c
```diff
@@ -31,6 +33,7 @@ void clock_tick_handler(int irq_num, void *dev_id) {
 	if (++cs->jiffies_cnt == (1 << SLOWDOWN_SHIFT)) {
 		cs->jiffies_cnt = 0;
 		cs->jiffies++;
+		timer_sched_cnt++;
```
alexkalmuk: Probably, it would be better to do `timer_sched_cnt++` right before `lthread_launch(&clock_handler_lt);` since we want to handle only these clock ticks (i.e. from cs_jiffies)?

alexkalmuk: By the way, these changes mean lthreads are working slower than interrupts somehow and do not count all clock interrupts properly, aren't them?

src/cmds/sys/date.c
AntonKozlov: It should. That's the real value of code-review. Thank you!

src/kernel/time/timer.c
AntonKozlov: > Probably, it would be better to do timer_sched_cnt++ right before lthread_launch(&clock_handler_lt) since we want to handle only these clock ticks (i.e. from cs_jiffies)?

Thanks, correct, that should be there.

> By the way, these changes mean lthreads are working slower than interrupts somehow and do not count all clock interrupts properly, aren't them?

Not exactly. Xen timer generates interrupts based on virtual time, i.e. CPU time consumed by particular OS. This requires us to catch up for passed real-time. More, tick is hardcoded for 10ms. So in Xen driver we'll call clock_tick_handler as many times as required to align with real time.

PR #1282, opened 2019-02-12
Xen port improvements made in xen18 branch

AntonKozlov:


src/arch/xen/include/COPYING
```diff
@@ -0,0 +1,23 @@
+Source files taken from the Xen distribution are governed by their original license.
```
anton-bondarev: This is unusual file or place for it

src/arch/xen/include/barrier.h
```diff
@@ -0,0 +1,24 @@
+/*
```
anton-bondarev: include guard?

src/arch/xen/include/xenstore.h
```diff
@@ -0,0 +1,17 @@
+#include <stdint.h>
```
anton-bondarev: include guard?

src/arch/xen/include/COPYING
AntonKozlov: It belong to Xen port at least. I have no strong opinion where would it belong. `src/arch/xen` (one level upper)?

anton-bondarev: Yes, one level upper looks better.
But it means than all code in xen folder is under this license?

src/arch/xen/include/xenstore.h
AntonKozlov: fixed as well, thanks!

src/arch/xen/include/barrier.h
AntonKozlov: thanks, fixed

src/arch/xen/include/COPYING
AntonKozlov: License specifies xenstore related files, it's xensotre.c and xenstore.h

AntonKozlov:
Just to make sure, COPYING moved to src/arch/xen/ is better, but is it good enough? Please duplicate LGTM after the move :)

anton-bondarev:
duplicate LGTM :)

PR #1285, opened 2019-02-13
travis: Get correct build status after travis_wait

alexkalmuk:
Fix the problem with `travis_wait`: it used status of the last executed command, that is of `tail`, not make.

### DIFF
--- a/scripts/continuous/build.sh
+++ b/scripts/continuous/build.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+# Since travis_wait doesn't work with docker well somewhy,
+# just use bell which prints something to stdout every 5 minutes.
+function bell() {
+	i=0
+	while true :
+	do
+		sleep 5m
+		i=$(($i+5))
+		echo "CI: Still building (minutes=$i)"
+	done
+}
+
+bell &
+bell_pid=$!
+
+make &> build.log
+RETVAL=$?
+tail -c 3M build.log
+
+# https://github.com/travis-ci/travis-ci/issues/6018 -- Prevert log truncation
+# on error. There are also many issues which refer to this one.
+sleep 5
+
+kill $bell_pid
+exit $RETVAL
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,7 +54,7 @@ script:
   - set -e # make this script fail as soon as any individual command fail
   - . ./scripts/docker/docker_rc.sh
   - dr make confload-"$TEMPLATE"
-  - travis_wait run dr 'make &> build.log; tail -c 3M build.log'
+  - run dr ./scripts/continuous/build.sh
   - run dr ./scripts/continuous/run.sh "$TEMPLATE"
 
 addons:


Deryugin:
I'm not familiar with travis configs, but for regular bash commands you can save first exit status as follows:

`make &> build.log; RETVAL=$?; tail -c 3M build.log; $(exit $RETVAL);`

alexkalmuk:
@Deryugin Yes, I know. But from what I understand `travis_wait` spawns the commands, backgrounds it with both stdout/stderr redirection to file, and then watching for its status. The problem here is that it's doing something like in your case:

```
make &> build.log; RETVAL=$?; tail -c 3M build.log; $(exit $RETVAL) &> out.log &
=== waiting for command finish and check result===
cat out.log
```

Probably, we should remove `set -e` at the beginning of `script:` and check build status after travis_wait after it finished.

Deryugin:
Why do we need to use `tail` here?

alexkalmuk:
> Why do we need to use `tail` here?

Because travis fails if the log size exceeds 4M. so we can't just call make.

AntonKozlov:
> `make &> build.log; RETVAL=$?; tail -c 3M build.log; $(exit $RETVAL);`

I'd prefer approach like this too.

> make &> build.log; RETVAL=$?; tail -c 3M build.log; $(exit $RETVAL) &> out.log &

what about `( make &> build.log; RETVAL=$?; tail -c 3M build.log; exit $RETVAL )` ?
Extra parentheses will create a subshell that travis could monitor

alexkalmuk:
@AntonKozlov Good tip, but it fails somehow https://travis-ci.org/embox/embox/jobs/493160521 
Config: https://travis-ci.org/embox/embox/jobs/493160521/config

AntonKozlov:
> $ dr make confload-"$TEMPLATE"
> Config complete
> The command "dr make confload-"$TEMPLATE"" exited with 0.

This is the last line. Magical travis_run may have rejected such form. What
about
`run dr '(make &> build.log; RETVAL=$?; tail -c 3M build.log; exit
$RETVAL)'` -- move subshell under docker_run?
If that fail, subshell can be replaced with honest build_make.sh with same
content

>

alexkalmuk:
@AntonKozlov What about something like this?

https://github.com/embox/embox/commit/83b7eb2e8385c21455c98359ed2e6b8122704e05

alexkalmuk:
@AntonKozlov 

> If that fail, subshell can be replaced with honest build_make.sh with same content

Yes, it failed https://travis-ci.org/embox/embox/jobs/493170907

AntonKozlov:
it failed in different way. Do we understand why? `dr` is `docker exec`,
for travis_wait it is a single process to wait. I see no deviations from
runnning `travis_make make`

>

alexkalmuk:
@AntonKozlov I do not understand why and wish to reduce the value of travis magic :) Just trying using build.sh like you proposed.

alexkalmuk:
What the hell

https://travis-ci.org/embox/embox/jobs/493190520

AntonKozlov:
I beleive it's one of you previoius tries:
https://travis-ci.org/embox/embox/jobs/493165932#L673

output of travis_wait cmd may be not available right after fail

alexkalmuk:
@AntonKozlov I pushed a new version with build.sh. Please, take a look

scripts/continuous/build.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+# Since travis_wait doesn't work with docker well somewhy,
+# just use bell which prints something to stdout every 5 minutes.
+function bell() {
+	for i in {5..20..5}; do
```
AntonKozlov: Why not forever?

scripts/continuous/build.sh
```diff
@@ -0,0 +1,21 @@
...
+		sleep 5m
+		echo "CI: Still building (minutes=$i)"
+	done
+}
+
+bell &
+bell_pid=$!
+
+make &> build.log; RETVAL=$?; tail -c 3M build.log
```
AntonKozlov: replace `;` with new lines may be? Just for readabillity.
Here and below

scripts/continuous/build.sh
```diff
@@ -0,0 +1,21 @@
...
+
+# https://github.com/travis-ci/travis-ci/issues/6018 -- Prevert log truncation
+# on error. There are also many issues which refer to this one.
+sleep 5;
```
AntonKozlov: redundant semicolon

scripts/continuous/build.sh
alexkalmuk: Hm, yes, since we assume build could not hand, infinite loop seems to be ok. Fixed

scripts/continuous/build.sh
alexkalmuk: Thanks, removed

scripts/continuous/build.sh
alexkalmuk: Thanks, I separated this with new lines

alexkalmuk:
@Deryugin I pushed a new version. If you think it's OK I will merge.

scripts/continuous/build.sh
alexkalmuk: btw, Thank you for review!

PR #1287, opened 2019-02-15
Telnet client exit

anton-bondarev:
Fix problem with zombie processes when "exit" sent from a client.
Rewrite telnetd. Use vfork() instead pthread_create() and handle SIGCHLD signal

### DIFF
--- a/src/cmds/net/telnetd/telnetd.c
+++ b/src/cmds/net/telnetd/telnetd.c
@@ -9,7 +9,7 @@
 #include <arpa/inet.h>
 #include <errno.h>
 #include <fcntl.h>
-#include <pthread.h>
+
 #include <stdio.h>
 #include <string.h>
 #include <signal.h>
@@ -23,7 +23,14 @@
 
 #include <util/math.h>
 
+#include <framework/mod/options.h>
+
 #define TELNETD_MAX_CONNECTIONS OPTION_GET(NUMBER,telnetd_max_connections)
+#define TELNETD_USE_PTHEAD      OPTION_GET(NUMBER,telnetd_use_pthread)
+
+#if TELNETD_USE_PTHEAD
+#include <pthread.h>
+#endif
 
 /* Telnetd address bind to */
 #define TELNETD_ADDR INADDR_ANY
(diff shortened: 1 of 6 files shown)


src/cmds/net/telnetd/telnetd.c
```diff
@@ -132,7 +140,7 @@ static int utmp_login(short ut_type, const char *host) {
 	return 0;
 }
 
-static void *shell_hnd(void* args) {
+static inline void *shell_hnd(void* args) {
```
alexkalmuk: Is `inline` really required here?

src/cmds/net/telnetd/telnetd.c
```diff
@@ -23,6 +23,14 @@
 
 #include <util/math.h>
 
+#define PTHEAD_USE 0
```
alexkalmuk: Maybe it's better to define this macro as module option?

src/cmds/net/telnetd/telnetd.c
```diff
@@ -197,54 +212,63 @@ static int telnet_fix_crnul(unsigned char *buf, int len) {
 	return bpo - buf;
 }
 
+struct client_args {
+	int sock;
+	int pptyfd[2];
+};
+
+#if !PTHEAD_USE
+void sigchild_handler(int sig) {
```
alexkalmuk: static?

src/cmds/net/telnetd/telnetd.c
```diff
@@ -368,10 +393,54 @@ int main(int argc, char **argv) {
...
 
 	if (argc > 1) {
-		client_descr = atoi(argv[1]);
-		MD(printf("telnetd: %d\n", client_descr));
-		telnetd_client_handler((void *)client_descr);
-		_exit(0);
+		struct client_args client_args;
+
+		if (0 == strcmp("connection", argv[1])) {
+#if !PTHEAD_USE
```
alexkalmuk: It's probably better to have two functions here - for pthreads and without pthreads, since wth these ifdef's it's hard to understand what happens.

src/cmds/net/telnetd/telnetd.c
```diff
@@ -23,7 +23,18 @@
 
 #include <util/math.h>
 
+
+
```
alexkalmuk: Empty lines

PR #1288, opened 2019-02-18
stm32 I2C driver

alexkalmuk:
* Add I2C driver for STM32
* Improve GY-30 i2c light sensor command - now it doesn't depend on stm32 
Add wiki page: https://github.com/embox/embox/wiki/GY-30-I2C-light-sensor-on-STM32
* Fix stm32 gpio driver
* Fix `sscanf` which ignored EOF


platform/stm32f4_multibots/cmds/robot/main.c
```diff
@@ -23,6 +23,8 @@
 
 #define NRF24_PLAYLOAD_LEN    16
 
+#define GY_30_I2C_BUS_NR      1
```
anton-bondarev: may be extract this as an option?

src/drivers/gpio/gpio.h
```diff
@@ -49,6 +49,10 @@ typedef volatile unsigned long gpio_mask_t;
 #define GPIO_MODE_OUT_ALTERNATE     (1 << 13)
 #endif
 
+/* Alternate function number. Bits: 17 ... 20 */
+#define GPIO_GET_ALTERNATE(mode) (((mode) & 0x1e0000) >> 17)
```
anton-bondarev: This is common part or STM32 depends

alexkalmuk: It doesn't depends on stm32. It's a general alternative function number. Each driver could implement this I think

alexkalmuk: So we set GPIO in alternative mode and set also alternative mode.

platform/stm32f4_multibots/cmds/robot/main.c
alexkalmuk: Thank you, fixed.

PR #1290, opened 2019-02-22
fix dlist_ndebug.h

oniich:
There's empty dlist_ndebug.h, which can not compile.


src/include/util/dlist.h
```diff
@@ -123,7 +123,12 @@ static inline void dlist_move(struct dlist_head *head, struct dlist_head *list)
  * @return 1 if list is not empty and zero if empty *
  */
 static inline int dlist_empty(const struct dlist_head *head) {
-	__dlist_debug_check(head);
+	/*
+	 * debug_check commented out because of util/tree. It somehow has:
```
oniich: Is this meant to be this way? IDK currently how to fix it

src/util/dlist/dlist_debug.h
```diff
@@ -136,6 +138,8 @@ static inline void dlist_add_prev(struct dlist_head *_new,
 	 * initialization */
 	// assert(__is_linked(list)); /* add to not initialized list */
 
+	// __dlist_debug_check(_new);
```
oniich: This 2 checks commented because otherwise system goes crazy during tests and hangs on infinite loop. Don't know why.

Deryugin: First check causes error because `_new` is not supposed to be initialized, so generally it's zeroes or garbage there. All fields are set later in `_dlist_add`

src/mem/mmap/mmap_kernel.c
```diff
@@ -41,7 +41,6 @@ int mmap_kernel_init(void) {
 	emmap = task_resource_mmap(task_kernel_task());
 
 	dlist_foreach_entry(marea, &early_emmap.marea_list, mmap_link) {
-		dlist_del_init(&marea->mmap_link);
```
anton-bondarev: This is required?

src/util/dlist/dlist_debug.c
```diff
@@ -10,22 +10,17 @@
...
-	/* XXX: not working in util/tree because of tree_delete_link func */
-#if 0
+
+#if ENABLE_DEBUG
+
 	const struct dlist_head *p = head->prev;
 	const struct dlist_head *n = head->next;
 	uintptr_t poison = head->poison;
 
 	assert((!poison || (void *) ~poison == head) &&
```
anton-bondarev: you can use assertf instead assert for this purpose.

src/mem/mmap/mmap_kernel.c
oniich: This has been deleted by mistake. I'll fix.

src/util/dlist/dlist_debug.c
oniich: Yes, didn't know about that.

src/util/dlist/dlist_debug.h
```diff
@@ -23,6 +23,8 @@
 #include <stdint.h>
 // #include <assert.h> /*it's required for assertion operation */
 
+#define ENABLE_DEBUG 1
```
Deryugin: Avoid ambiguous macro names: this header can be included in some source files which relies on the macro with the same name. It's better to use something like `DLIST_DEBUG`.

Also, I guess this macro can be defined by some module option.

oniich: Yep, gonna fix

src/include/util/dlist.h
```diff
@@ -144,6 +143,15 @@ static inline struct dlist_head *dlist_prev(const struct dlist_head *list) {
 	return list->prev;
 }
 
+/**
+ * @def dlist_head_init_entry()
+ *
+ * TODO
+ *
+ */
+#define dlist_head_init_entry(element, link_member) \
```
Deryugin: Is this macro used anywhere?

oniich: Was used, but now it's not. It needs to be removed, if not used anywhere, yes?

Deryugin:
Please, replace abstract module for `dlist` with a simple module (we have a single implementation anyway).

src/util/dlist/Mybuild
```diff
@@ -1,13 +1,12 @@
...
 
-static module DListDebug extends DList {
+static module dlist_debug extends dlist {
 	source "dlist_debug.h"
 	source "dlist_debug.c"
-}
-static module DListNDebug extends DList {
-	source "dlist_ndebug.h"
+
+	option number dlist_enable_debug=1
```
Deryugin: Options are defined localy for each module, so it's ok to use shorter name like `debug` or something like that.

oniich: OK

src/util/dlist/dlist_debug.h
Deryugin: Another unused macro. Delete it, please.

Issue #1292, opened 2019-02-22
x86/qt-app build error


sikmir:
Steps to reproduce:
```bash
$ make confload-x86/qt-app
$ make
...
src/drivers/console/fbcon.c:157:12: error: 'char_dev_idesc_fstat' undeclared here (not in a function)
  .fstat  = char_dev_idesc_fstat,
            ^~~~~~~~~~~~~~~~~~~~
make[4]: *** [mk/image_lib.mk:33: build/base/obj/src/drivers/console/fbcon.o] Error 1
...
```

AntonKozlov:
Do we have the template in CI?

Deryugin:
No, we don't.

anton-bondarev:
Fix it in #1293 Check please

sikmir:
It's buildable now.

PR #1293, opened 2019-02-23
Fix x86/qt-app template

anton-bondarev:
Fix issue #1292 and add x86/qt-app template to travis build


.travis.yml
```diff
@@ -17,6 +17,7 @@ env:
     - TEMPLATE=x86/qemu
     - TEMPLATE=x86/smp
     - TEMPLATE=x86/user_apps
+    - TEMPLATE=x86/qt_app
```
alexkalmuk: x86/qt-app

PR #1296, opened 2019-02-27
[WIP] Fix noreturn keyword

anton-bondarev:


src/cmds/exit.c
```diff
@@ -24,4 +24,6 @@ int main(int argc, char *argv[]) {
 	}
 
 	exit(ecode);
+
```
AntonKozlov: Shouldn't `exit` be marked as `noreturn` instead?

src/include/compiler.h
```diff
@@ -9,17 +9,17 @@
 #ifndef COMPILER_H_
 #define COMPILER_H_
 
-#ifdef __GNUC__
-
+#if defined(__GNUC__) && !defined(__cplusplus)
 #define _NORETURN __attribute__ ((noreturn))
+#else /* __GNUC__ */
+#define _NORETURN
```
AntonKozlov: Commit message specifies `c++11`, but defines are not distinguish c++ version. Older ones will have no `noreturn`
C++11 do have `noreturn`: https://en.cppreference.com/w/cpp/language/attributes/noreturn

src/cmds/exit.c
anton-bondarev: Remove this commit

src/include/compiler.h
anton-bondarev: yes, of cause you right  C++11 have 'noreturn' attribute. But this fix for compilation c++ code by gcc8

AntonKozlov: Right, but isn't it because g++8 is c++11 or higher by default? Why should we change for older compilers with another default standart?
```
anton@zephyr:~$ g++ --version
g++ (GCC) 8.2.1 20181127
anton@zephyr:~$ g++ -E -dM -x c++ - < /dev/null | grep cplusplus
#define __cplusplus 201402L
```
My version is even c++14

anton-bondarev: @AntonKozlov 
Yes, I'm agree. it made only to easier ifdef switch
correct way is 

```
 #if __cplusplus<=201103L 
...
#endif

```

but I check it on some version of compiler and it works correctly

PR #1298, opened 2019-02-28
Fix some extbld compiler warnings

anton-bondarev:


src/compat/libc/include/ctype.h
```diff
@@ -27,6 +27,9 @@ extern const unsigned char _ctype[];
 #include <stdint.h>
 __BEGIN_DECLS
 
+#pragma GCC diagnostic push
```
Deryugin: Please, add `#ifdef __GNUC__` to ensure compatibility with non-gcc compilers.

Also it seems that gcc didn't support diagnostic push/pop for version 4.5.4 and before, I'm not sure if we have to keep support it (this check can be with `#if __GNUC__ > N`)

PR #1303, opened 2019-03-12
Add simple library to print FPS on screen

Deryugin:


platform/mesa/cmds/osdemo_fb/osdemo_fb.c
```diff
@@ -63,38 +64,19 @@ void init_buffers(void) {
...
-			time_prev = time;
-			fps = counter > 99 ? 99 : counter;
-			counter = 0;
-		}
-
-		fps_str[4] = '0' + fps / 10;
-		fps_str[5] = '0' + fps % 10;
-		fb_overlay_put_string(0, 0, "Embox MESA demo v0.2");
-		fb_overlay_put_string(0, 1, fps_str);
+		fps_set_format("Embox MESA demo v0.2\nFPS=%2d");
```
alexkalmuk: I think an argument for `%2d` is expected?

src/lib/fps/fps.c
```diff
@@ -0,0 +1,85 @@
...
+
+	assert(fb);
+
+	memset(msg_buf, 0, sizeof(msg_buf));
+
+	clock_gettime(CLOCK_REALTIME, &time);
+
+	counter++;
+
+	if (time.tv_sec != time_prev.tv_sec) {
```
alexkalmuk: Shouldn't we check only one second expired (I mean exactly one, not 2 or 3...)?

platform/mesa/cmds/osdemo_fb/osdemo_fb.c
Deryugin: Fixed this anyway :)

src/lib/fps/fps.c
Deryugin: Fixed

PR #1307, opened 2019-03-16
Cursor improve

jackbogdanov:


alexkalmuk:
Just another way to think about: Can we try QT standard cursor class like this https://doc.qt.io/archives/qt-4.8/qcursor.html? I cannot say right now whether this way better or not, but may be it deserves some investgation.

jackbogdanov:
Thanks, I think standard cursor would better, I will try to implement it!

AntonKozlov:
This PR deserves a screenshot!

image

The code seems OK for me and the result is so much better. So I don't really understand WIP label. 
Is there something that I missed, beside QCursor suggested later?
BTW, do we expect it will fix minor rendering issues?

AntonKozlov:
> PR is cool.

> the way for Eugene to go more in depths with graphics.

Oh, good then. Thanks for clarifying. Just making sure Eugene efforts appreciated :)

alexkalmuk:
@jackbogdanov I see you removed WIP. Does it mean you finished this task?

jackbogdanov:
Yes, this PR can be merged to master and I can later create new one, with QCursor implementation.

alexkalmuk:
OK. Thank you for your contribution! :)

P.S. Minor remark: Next time, please, add your commit message starting with a capital letter. For example, "qt: Add cursor image".

PR #1310, opened 2019-03-21
e2k context switch

alexkalmuk:
Implement context_switch for e2k (elbrus).

The following tests on context_switch and threads passed (included in e2k/template):
```
@Runlevel(1) include embox.test.hal.context_switch_test(stack_sz = 0x4000)
@Runlevel(1) include embox.test.kernel.sched.running_threads_test
@Runlevel(1) include embox.test.kernel.thread.thread_priority_test(threads_quantity=10)
@Runlevel(1) include embox.test.kernel.thread.thread_systimer_irq_test
@Runlevel(1) include embox.test.kernel.thread.thread_test
```

### DIFF
--- a/src/arch/e2k/context_switch.S
+++ b/src/arch/e2k/context_switch.S
@@ -1,0 +1,89 @@
+/**
+ * @file
+ * @brief Context switch for E2K
+ *
+ * @date 18.03.2019
+ * @author Alexander Kalmuk
+ */
+
+#include <hal/context.h>
+#include <e2k_api.h>
+
+.section ".text", "ax"
+
+/* TODO Save and restore Global Registers. */
+
+.global $context_switch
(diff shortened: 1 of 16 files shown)


src/arch/e2k/context_switch.S
```diff
@@ -0,0 +1,89 @@
...
+ */
+
+#include <hal/context.h>
+#include <e2k_api.h>
+
+.section ".text", "ax"
+
+/* TODO Save and restore Global Registers. */
+
+.global $context_switch
```
anton-bondarev: may be create macro C_ENTRY?

alexkalmuk: Thanks, I defined such macro. Probably, it should be placed in the common asm/linkage.h in the future.

PR #1311, opened 2019-03-22
Fix some MMU issues

Deryugin:
* Add implementation for `task.resource.mmap` to use single context for all tasks
* Add simple utility to measure memory performance
* Minor ARM-related fixes

### DIFF
--- a/scripts/continuous/fs/run.sh
+++ b/scripts/continuous/fs/run.sh
@@ -127,7 +127,6 @@ interactive_tests() {
 	rm $PID_FILE
 }
 
-interactive_tests $EXPECT_TESTS_BASE/x86_fs_unit_tests.config
 interactive_tests $EXPECT_TESTS_BASE/x86_fs_shell_commands.config
 
 for f in $FS_TEST_RO; do
--- a/src/lib/exec/Mybuild
+++ b/src/lib/exec/Mybuild
@@ -3,7 +3,7 @@ package embox.lib
 static module LibExec {
 	source "exec.c"
 
-	depends embox.kernel.task.resource.mmap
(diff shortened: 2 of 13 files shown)


src/lib/exec/Mybuild
```diff
@@ -3,7 +3,7 @@ package embox.lib
 static module LibExec {
 	source "exec.c"
 
-	depends embox.kernel.task.resource.mmap
```
anton-bondarev: are you sure that exec can be used only with full mmap implementation?

scripts/continuous/fs/run.sh
```diff
@@ -127,7 +127,6 @@ interactive_tests() {
 	rm $PID_FILE
 }
 
-interactive_tests $EXPECT_TESTS_BASE/x86_fs_unit_tests.config
```
anton-bondarev: may be just comment it?

alexkalmuk: It was my proposition, should be reverted in the future. Not master seems to be broken anyway. I am going to disable all fs autotests temporary to figure out what's going on.

anton-bondarev: @alexkalmuk Ok, thanks for the explanation

src/lib/exec/Mybuild
Deryugin: Yes, I'm sure, it breaks applications because LibExec explicitly operates with mareas/mmaps (i.e. loads section to specific vitrual addresses), so it goes wrong with single vmem context.

anton-bondarev: @Deryugin ok, thanks

PR #1315, opened 2019-03-25
Stlport to extbld

anton-bondarev:
- Fix openlibm c++ compilation
- Fix usb_whitelist_dev after char_dev rework
- Use extbld script in STLport
- Add simple demo for STL using

### DIFF
--- a/src/drivers/usb/usb_whitelist_dev.c
+++ b/src/drivers/usb/usb_whitelist_dev.c
@@ -9,6 +9,9 @@
 #include <errno.h>
 #include <string.h>
 #include <stdlib.h>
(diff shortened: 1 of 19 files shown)


src/drivers/usb/usb_whitelist_dev.c
```diff
@@ -9,6 +9,10 @@
 #include <errno.h>
 #include <string.h>
 #include <stdlib.h>
+#include <sys/ioctl.h>
+#include <sys/stat.h>
+#include <sys/uio.h>
+
```
Deryugin: Please, remove this blank line.

PR #1316, opened 2019-03-26
Fix etna-related examples

Deryugin:
Use `@App` signature to reset MESA .bss and .data sections.

### DIFF
--- a/src/arch/arm/armlib/exceptions/Mybuild
+++ b/src/arch/arm/armlib/exceptions/Mybuild
@@ -2,6 +2,7 @@ package embox.arch.arm.armlib
 
 module exceptions {
 	option number log_level=0
+	option boolean keep_going=true /* Don't stall on undefined exception */
 
 	source "entry.S"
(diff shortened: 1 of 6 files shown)


src/arch/arm/armlib/exceptions/Mybuild
```diff
@@ -2,6 +2,7 @@ package embox.arch.arm.armlib
 
 module exceptions {
 	option number log_level=0
+	option boolean keep_going=true /* Don't stall on undefined exception */
```
anton-bondarev: may be default make false?

Deryugin: Before these changes OS will not stop on exception, so I decided to keep the same behavior (for some examples it even works well and application continues to execute normally instead of infinite exception loop).

But I'm not sure what to do here.

anton-bondarev: Ok

PR #1318, opened 2019-03-28
add functionality to xen_net driver

CherepanovAleksei:


AntonKozlov:
It looks like I just merged the PR, but it's not :) There is also commit with changes made on-site (https://github.com/embox/embox/commit/d1eacb6e15b323dbbec180a70903fde655fe471f), and now the driver doesn't compile

PR #1322, opened 2019-04-04
heap_simple: Rework

Deryugin:
This feature fixes problems with heap_simple on ARM, where memory access should be 4-byte aligned.

Also extend `memtest` command for heap stress test.

### DIFF
--- a/src/cmds/testing/memtest/memtest.c
+++ b/src/cmds/testing/memtest/memtest.c
@@ -12,27 +12,143 @@
 #include <stdlib.h>
 #include <string.h>
 #include <stdint.h>
+#include <stdbool.h>
 #include <unistd.h>
 #include <time.h>
 
+#include <framework/mod/options.h>
 #include <mem/vmem.h>
-
-static const size_t default_region_len = 16 * 1024 * 1024;
 
 static void print_help(char **argv) {
 	assert(argv);
 	assert(argv[0]);
 	printf("Usage: %s [-nh] [-l LENGHT]\n", argv[0]);
 	printf("\tOptions: -n Test non-cached memory\n");
-	printf("\t         -l LENGTH Number of bytes to be copied (default 0x1000000)\n");
-	printf("\t         -r REPEAT Number of repeats (default 64)\n");
+	printf("\t         -l LENGTH Number of bytes to be operated "
+			"(default 0x1000000 for memcpy() and 0x2000 for heap)\n");
+	printf("\t         -r REPEAT Number of repeats (default 64 for memcpy() and 2048 for heap)\n");
+	printf("\t         -p Run test for heap (a.k.a. pyramid)\n");
+	printf("\t         -q Maximum number of heap regions\n");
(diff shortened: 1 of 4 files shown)


src/cmds/testing/memtest/memtest.c
```diff
@@ -12,27 +12,143 @@
...
+	printf("\t         -l LENGTH Number of bytes to be operated "
+			"(default 0x1000000 for memcpy() and 0x2000 for heap)\n");
+	printf("\t         -r REPEAT Number of repeats (default 64 for memcpy() and 2048 for heap)\n");
+	printf("\t         -p Run test for heap (a.k.a. pyramid)\n");
+	printf("\t         -q Maximum number of heap regions\n");
+	printf("\t         -s Random seed\n");
+}
+
+/* Heap-related stuff */
+#define MAX_HEAP_REGIONS 4096
```
anton-bondarev: may be as an option?

src/cmds/testing/memtest/memtest.c
```diff
@@ -12,27 +12,143 @@
...
+
+static int max_region_len;
+static int random_seed;
+static int heap_regions;
+
+#define SCREEN_WIDTH 80
```
anton-bondarev: in ideal world should be 
```struct winsize w;
ioctl(STDOUT_FILENO, TIOCGWINSZ, &w);
```

Deryugin: Is it actually implemented in Embox?

src/cmds/testing/memtest/memtest.c
```diff
@@ -12,27 +12,143 @@
...
+	printf("\t         -r REPEAT Number of repeats (default 64 for memcpy() and 2048 for heap)\n");
+	printf("\t         -p Run test for heap (a.k.a. pyramid)\n");
+	printf("\t         -q Maximum number of heap regions\n");
+	printf("\t         -s Random seed\n");
+}
+
+/* Heap-related stuff */
+#define MAX_HEAP_REGIONS 4096
+
+static int max_region_len;
```
anton-bondarev: may be create a structure with global fields and pass it in routines as argument?

src/mem/heap/heap_simple.c
```diff
@@ -1,116 +1,132 @@
...
-	if(NULL == mem_pool) {
+	if (NULL == mem_pool) {
 		return -1;
 	}
 
-	managed_memory_start = (void*) mem_pool;
-	last_valid_address   = managed_memory_start;
-	has_initialized      = 1;
+	heap_simple_start =
+		(void*) binalign_bound((uintptr_t) mem_pool, HEAP_ALIGN);
```
alexkalmuk: I think we should assert heap_simple_start is already aligned to HEAP_ALIGN bound after `page_alloc` above since if it's not so, we will align get out of bounds below 

heap_simple_end   = heap_simple_start + HEAP_SIZE;

PR #1324, opened 2019-04-08
Ping domU

KutuevVladimir:


AntonKozlov:
Thanks! Will look shortly

scripts/xen/gold_linux/run_alpine
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+sudo ln $(dirname $0)/alpine.cfg /etc/xen/alpine.cfg
+
+# Getting alpine linux image
+sudo mkdir /data && cd /data
+sudo wget dl-cdn.alpinelinux.org/alpine/v3.9/releases/x86_64/alpine-virt-3.9.2-x86_64.iso
```
AntonKozlov: `wget -c` or download only if the file is absent, to avoid transferring an extra bytes

oniich:
So now we're supposed to run `up_bridge.sh` from dom0 to configure bridge? If yes, may be consider to add this into README.
Anyway, LGTM.

AntonKozlov:
> So now we're supposed to run `up_bridge.sh` from dom0 to configure bridge? If yes, may be consider to add this into README.

This can be added even in `run_xen` (embox) or `run_alpine` scripts. But I'd prefer do not block the PR by this

> Anyway, LGTM.

Thanks!

PR #1329, opened 2019-04-10
Rework flash subsystem

Deryugin:
* Simplify interface
* Rework flash emulator
* Fix DumbFS
* Fix block devices idesc ops for new VFS


src/cmds/fs/mkflashemu.c
```diff
@@ -0,0 +1,49 @@
...
+static void print_usage(void) {
+	printf("Usage: mkflashemu [-h] [[-n BLOCKS] [-b BLOCK_SIZE] dev_name]\n");
+}
+
+int main(int argc, char **argv) {
+	int opt;
+	size_t block_size = 512;
+	int blocks = 1024;
+
+	getopt_init();
```
anton-bondarev: it's not necessary

src/drivers/flash/flash.c
```diff
@@ -1,328 +1,132 @@
...
-
-static int flashbdev_write(struct block_dev *bdev,
-		char *buffer, size_t count, blkno_t blkno) {
-	struct flash_dev *flash;
-	uint32_t startpos, endpos;
-
-	flash = (struct flash_dev *)bdev->privdata;
-	startpos = flash->start + (blkno * flash->block_info.block_size);
-	endpos = startpos + count - 1;
+static int flash_initialized = 0;
```
anton-bondarev: May be use UNIT_INIT?

src/cmds/fs/mkflashemu.c
Deryugin: fixed

src/drivers/flash/flash.c
Deryugin: This is necessary due to broken RunLevel sequence: sometimes we have to init devices before init of this module.

PR #1336, opened 2019-04-16
E2k clean build

anton-bondarev:
* Clean e2k compilation flags
* Clean some e2k code
* Fix some code where pointers casting to integer types (It's incorrect for not 32 bits platforms перевод)

### DIFF
--- a/src/arch/e2k/e2k_syscall_handler.c
+++ b/src/arch/e2k/e2k_syscall_handler.c
@@ -7,13 +7,13 @@
  */
 
 #include <stdint.h>
-
+#include <inttypes.h>
 #include <kernel/printk.h>
 
 unsigned long cpuSyscall(int number, uint64_t arg1, uint64_t arg2,
 			 uint64_t arg3, uint64_t arg4)
 {
-	printk("syscall handler %d (%x, %x, %x, %x)\n", number, arg1, arg2, arg3, arg4);
+	printk("syscall handler %d (%"PRIu64", %"PRIu64", %"PRIu64", %"PRIu64")\n", number, arg1, arg2, arg3, arg4);
 
 	return 0;
 }
@@ -22,7 +22,7 @@ unsigned long cpuSyscall(int number, uint64_t arg1, uint64_t arg2,
 unsigned long cpuJmpSyscall(int number, uint64_t arg1, uint64_t arg2,
 			 uint64_t arg3, uint64_t arg4)
 {
-	printk("jmpsyscall handler %d (%x, %x, %x, %x)\n", number, arg1, arg2, arg3, arg4);
+	printk("jmpsyscall handler %d (%"PRIu64", %"PRIu64", %"PRIu64", %"PRIu64")\n", number, arg1, arg2, arg3, arg4);
 
 	return 0;
 }
--- a/src/arch/e2k/include/va-e2k.h
+++ b/src/arch/e2k/include/va-e2k.h
@@ -1,0 +1,43 @@
+/* Internals of varargs for e2k */
+#ifndef VA_E2K_H
+#define VA_E2K_H
+
+#if (!defined __STDARG_H) && (!defined __VARARGS_H)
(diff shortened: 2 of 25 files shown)


src/drivers/serial/am85c30/am85c30.c
```diff
@@ -112,13 +110,12 @@ static int am85c30_getc(struct uart *dev) {
 	while (am85c30_has_symbol(NULL) == 0) {
 	}
 
-	r1 = am85c30_com_inb_command(port, AM85C30_RR1);
-	ch = am85c30_com_inb(port+1);
+	r1 = am85c30_com_inb_command((void*)(uintptr_t)port, AM85C30_RR1);
```
alexkalmuk: Probably, declare port as `uintptr_t ` in that case? By the way, you already use just `(void*) port` in am85c30_has_symbol

src/arch/e2k/e2k_syscall_handler.c
```diff
@@ -22,7 +22,7 @@ unsigned long cpuSyscall(int number, uint64_t arg1, uint64_t arg2,
 unsigned long cpuJmpSyscall(int number, uint64_t arg1, uint64_t arg2,
 			 uint64_t arg3, uint64_t arg4)
 {
-	printk("jmpsyscall handler %d (%x, %x, %x, %x)\n", number, arg1, arg2, arg3, arg4);
+	printk("jmpsyscall handler %d (%"PRIu64", %"PRIu64", %"PRIu64", %"PRIu64")\n", number, arg1, arg2, arg3, arg4);
```
Deryugin: Why not just `%llx`? We usually don't use this "PRI.." stuff.

src/drivers/serial/am85c30/am85c30.c
```diff
@@ -8,32 +6,34 @@
...
-#define UART_BASE   0x83200040
+#define UART_BASE   (uintptr_t)0x83200040
 #define	IRQ_NUM     3 /* ZILOG_IRQ_DEFAULT 3 */ /* default IRQ # */
 
-
 static inline void
 am85c30_com_outb(void *iomem_addr, uint8_t data)
 {
-	e2k_write8(data, UART_BASE + 1);
+	e2k_write8(data, (void*)(UART_BASE + 1));
```
Deryugin: May be it's a better idea to change `e2k_readX/e2k_writeX` functions to use second argument as uintptr_t? Otherwise we'll have to convert it to (void*) everywhere.

src/drivers/serial/am85c30/am85c30.c
```diff
@@ -8,32 +6,34 @@
  * @date 18.12.2017
  */
 
+#include <stdint.h>
+
+#include "../am85c30/am85c30.h"
```
Deryugin: `#include "am85c30.h"`?

src/arch/e2k/e2k_syscall_handler.c
anton-bondarev: It's more correctly using PRIu64 in printf() for uint64_t.
I think we should replace another than PRI for 'stdint' types in future

src/arch/e2k/include/va-e2k.h
```diff
@@ -0,0 +1,39 @@
+/* Internals of varargs for e2k */
```
Deryugin: Macro guard?

src/drivers/serial/am85c30/am85c30.c
anton-bondarev: Ops, Thanks a lot!

src/drivers/serial/am85c30/am85c30.c
anton-bondarev: Done

src/arch/e2k/include/va-e2k.h
anton-bondarev: done

PR #1340, opened 2019-04-17
Sending XenStore information about rings

oniich:


AntonKozlov:
There was some issues with the code, that we resolved all together when we''ve met. I forget to commit the merge first, so the fixes in merge commit itself. Sorry for that.

PR #1343, opened 2019-04-23
[WIP] FB/devfs

Deryugin:
* Use non-vfs-specific interface for FB/devfs access
* Work on unifying chardev interfaces


src/drivers/char/char_dev_stub.c
```diff
@@ -42,3 +42,11 @@ struct idesc *char_dev_open(struct node *node, int flags) {
 int char_dev_idesc_fstat(struct idesc *idesc, void *buff) {
 	return 0;
 }
+
+struct idesc *char_dev_default_open(struct dev_module *cdev, void *priv) {
+	return NULL;
```
anton-bondarev: May be also use log_error?

src/drivers/video/fb.c
```diff
@@ -71,34 +71,33 @@ static void fb_ops_fixup(struct fb_ops *ops) {
...
 		fb_ops_fixup(&info->ops);
-
-
-		/* FIXME probably should be in open/start */
 		fb_update_current_var(info);
+		fb_devfs_create(info, map_base, map_size);
 	}
-	fb_devfs_create(ops, map_base, map_size);
+out:
+	mutex_unlock(&fb_static);
```
anton-bondarev: mutex_lock/mutex_unlock may be  used safely during the memory allocation because further used only the pointer to  the allocated memory 
`info = &dev->info;`

Deryugin: Problem may occur if `fb_lookup()` is called before intialization is complete (for example, if `dlist_add_next` is not finished yet or `info->ops` are not initialized yet).

Same problems for calling `fb_create()` in another thread at the same time.

PR #1345, opened 2019-04-23
posix: Add missing getgrXXX functions

wietmann:
Also:
* Fix typos with incorrect buffer size passed to reentrant functions in getpwuid and getmaxuid
* getgrgid now returns pointer to correct internal data structure
* All getgrXXX functions return (or set errno to) ENOENT if no entry is found and corresponding errno value on other errors

### DIFF
--- a/src/compat/posix/pwd_grp/pwd_grp.c
+++ b/src/compat/posix/pwd_grp/pwd_grp.c
@@ -291,6 +291,20 @@ int fgetgrent_r(FILE *fp, struct group *gbuf, char *tbuf,
 	return 0;
 }
 
+struct group *fgetgrent(FILE *stream) {
+	static struct group grp_buf;
+	static char buf[64];
+	struct group *res;
+	int ret;
+
+	if (0 != (ret = fgetgrent_r(stream, &grp_buf, buf, sizeof(buf), &res))) {
+		SET_ERRNO(ret);
+		return NULL;
+	}
+
+	return res;
+}
+
 int getgrnam_r(const char *name, struct group *grp,
 	char *buf, size_t buflen, struct group **result) {
 	int res;
--- a/src/compat/posix/pwd_grp/pwd_grp.c
+++ b/src/compat/posix/pwd_grp/pwd_grp.c
@@ -39,6 +39,8 @@ static int read_field(FILE *stream, char **buf, size_t *buflen, char **field,
 
 	while (ch != delim) {
 		if (ch == EOF) {
+			if (ferror(stream))
+				return EIO;
 			return EOF;
 		}
 
@@ -57,7 +59,7 @@ static int read_field(FILE *stream, char **buf, size_t *buflen, char **field,
 	}
 
 	if (*buflen <= 0) {
-		return -ERANGE;
+		return ERANGE;
 	}
 
 	*((*buf)++) = '\0';
@@ -74,7 +76,7 @@ static int read_int_field(FILE *stream, const char *format, void *field, int del
 
 	ret = fscanf(stream, format, field);
 	if (0 > ret && ferror(stream)) {
-		return -ret;
+		return EIO;
 	}
 
 	if (delim != (val = fgetc(stream))) {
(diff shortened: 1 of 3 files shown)


src/compat/posix/pwd_grp/pwd_grp.c
```diff
@@ -219,7 +223,7 @@ struct passwd *getpwuid(uid_t uid) {
 	struct passwd *res;
 	int ret;
 
-	if (0 != (ret = getpwuid_r(uid, &getpwuid_buffer, buff, 80, &res))) {
+	if (0 != (ret = getpwuid_r(uid, &getpwuid_buffer, buff, 0x80, &res))) {
```
anton-bondarev: sizeof(buff) looks better then 0x80

src/compat/posix/pwd_grp/pwd_grp.c
```diff
@@ -356,7 +444,7 @@ int getmaxuid() {
 		return res;
 	}
 
-	while (0 == (res = fgetpwent_r(file, &pwd, buff, 80, &result))) {
+	while (0 == (res = fgetpwent_r(file, &pwd, buff, 0x80, &result))) {
```
anton-bondarev: Use sizeof(buff) instead of 0x80

wietmann: Done

src/compat/posix/pwd_grp/pwd_grp.c
wietmann: Done

wietmann:
Also, as I understand, open_db is a part of an attempt to separate file access from getpw stuff implementation. Would it be a good idea to expand it to some common db open/read/write/close interface, which in turn can be implemented as access to passwd file or something else in separate modules?

anton-bondarev:
> Also, as I understand, open_db is a part of an attempt to separate file access from getpw stuff implementation. Would it be a good idea to expand it to some common db open/read/write/close interface, which in turn can be implemented as access to passwd file or something else in separate modules?

@wietmann Yes, it may be useful for some cases.

anton-bondarev:
@wietmann Please remove follow code
```static inline
void setgrent(void) {
	DPRINT();
}

static inline
struct group *getgrent(void) {
	DPRINT();
	errno = EPERM;
	return 0;
}

static inline
void endgrent(void) {
	DPRINT();
}
```
from third-party/samba/samba_embox_compat.h
It broke [template x86/test/fs on travis]( https://travis-ci.org/embox/embox/builds/523576858?utm_source=github_status&utm_medium=notification)

src/compat/posix/pwd_grp/pwd_grp.c
```diff
@@ -73,8 +75,10 @@ static int read_int_field(FILE *stream, const char *format, void *field, int del
 	sscanf("0", format, field);
 
 	ret = fscanf(stream, format, field);
-	if (0 > ret && ferror(stream)) {
-		return -ret;
+	if (0 > ret) {
+		if (ferror(stream))
+			return EIO;
+		return EOF;
```
alexkalmuk: Thanks for your contribution. We don't merge since ssh test is broken on Travis. First thing is this line - when `fscanf` returns something lower than 0, it probably means it's empty field , that is `::`. So we can't just return `EOF` here, it breaks `fgetspent` for example.

src/compat/posix/pwd_grp/pwd_grp.c
```diff
@@ -297,35 +325,46 @@ int getgrnam_r(const char *name, struct group *grp,
...
 
 	fclose(file);
-
-	if (0 != res) {
-		result = NULL;
-		return res == ENOENT ? 0 : -1;
-	}
-	return 0;
-
+	return res;
```
alexkalmuk: http://pubs.opengroup.org/onlinepubs/9699919799/functions/getgrnam.html

`The getgrnam_r() function shall return zero on success or if the requested entry was not found`

So we shall return 0 if the requested entry was not found.

It also breaks ssh.

alexkalmuk:
Here is a patch which fixes ssh.

[patch.txt](https://github.com/embox/embox/files/3121030/patch.txt)

src/compat/posix/pwd_grp/pwd_grp.c
alexkalmuk: Probably, we shall delete `return EOF` and either leave a comment about empty fields `::` and fall through below. As I can see next lines below are checking something about field emptiness already.

src/compat/posix/pwd_grp/pwd_grp.c
wietmann: Done

src/compat/posix/pwd_grp/pwd_grp.c
wietmann: Done

wietmann:
What will be a reasonable value for `_SC_GETGR_R_SIZE_MAX`? I can see that `_SC_GETPW_R_SIZE_MAX` was set to 512 bytes and as I understand both may be connected to some `LOGIN_NAME_MAX`, `NGROUP_MAX`, etc, which are not defined for now.

anton-bondarev:
> What will be a reasonable value for `_SC_GETGR_R_SIZE_MAX`? I can see that `_SC_GETPW_R_SIZE_MAX` was set to 512 bytes and as I understand both may be connected to some `LOGIN_NAME_MAX`, `NGROUPS_MAX`, etc, which are not defined for now.

There was no strong reason to set up _SC_GETPW_R_SIZE_MAX 512 bytes. I think we should create header local_limits.h for example and define some macro in it NSS_BUFLEN_GROUP for _SC_GETPW_R_SIZE_MAX & _SC_GETGR_R_SIZE_MAX (512 bytes will enough for it) and LOGIN_NAME_MAX   (128 or 256 bytes will enough)

anton-bondarev:
Thanks a lot.
I've merged it to master

PR #1348, opened 2019-04-26
e2k: clean: Clean some e2k specific headers

anton-bondarev:


src/arch/e2k/include/asm/io.h
```diff
@@ -1,3 +1,4 @@
+
```
Deryugin: Empty line

anton-bondarev: Done

alexkalmuk:
Cool, finally, we get rid of some large e2k headers :)

PR #1358, opened 2019-05-08
Fix builds

Deryugin:
* Fix arm/overo
* Fix arm/bpi
* Fix arm/raspi


src/arch/arm/mmu/mmu_small_page.c
```diff
@@ -124,7 +126,9 @@ uint32_t vmem_info(uint32_t vaddr)
 
     vaddr &= ~0xFFF;
 
-    printk("vmem info: base %p; vaddr %p", (void*) vmem_current_context(), (void *) vaddr);
+    printk("vmem info: base %p; vaddr %p",
```
anton-bondarev: log_xxx?

Deryugin: This function is not used directly right now, so I don't think it should me modified this way.
May be it will be used in some command, so `printf()` will be more appropriate.

PR #1381, opened 2019-05-23
Added guide for start nginx on domU

KutuevVladimir:
Modified httpd Makefile (start static build by specifying STATIC variable)
Modified VM (Vagrantfile) and domU (alpine.cfg) configs: allocate more memory

### DIFF
--- a/scripts/xen/gold_linux/README
+++ b/scripts/xen/gold_linux/README
@@ -15,7 +15,9 @@ echo "http://dl-3.alpinelinux.org/alpine/v3.9/main" >> /etc/apk/repositories
 # add stable community repository (required for docker installation)
 echo "http://dl-cdn.alpinelinux.org/alpine/latest-stable/community" >> /etc/apk/repositories
 
-# intall docker (now doesn't work, it is necessary to extend disk space)
+# Run nginx with index.html from hwproj.me in docker container
 apk update
 apk add docker
-
+rc-update add docker boot
+service docker start
+docker run --name nginx -d -p 80:80 vladimirkutuev/nginx:v1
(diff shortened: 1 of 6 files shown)


scripts/xen/gold_linux/README
```diff
@@ -15,7 +15,9 @@ echo "http://dl-3.alpinelinux.org/alpine/v3.9/main" >> /etc/apk/repositories
 # add stable community repository (required for docker installation)
 echo "http://dl-cdn.alpinelinux.org/alpine/latest-stable/community" >> /etc/apk/repositories
```
AntonKozlov: At the time of writing https://mirrors.alpinelinux.org/ propoes `dl-cdn` as first choice, I think we'll stop at this.
I've change `dl-3` above to this.

But `latest-stable` looks like moved further. I get strange errors while trying add docker until I replaced this with `v3.9` as above.  

Just wanted to eliminate any surprises https://github.com/embox/embox/commit/d915e6a6f1235874858d45558eab1414642355a9

scripts/xen/gold_linux/README
```diff
@@ -15,7 +15,9 @@ echo "http://dl-3.alpinelinux.org/alpine/v3.9/main" >> /etc/apk/repositories
...
 
-# intall docker (now doesn't work, it is necessary to extend disk space)
+# Run nginx with index.html from hwproj.me in docker container
 apk update
 apk add docker
-
+rc-update add docker boot
+service docker start
+docker run --name nginx -d -p 80:80 vladimirkutuev/nginx:v1
```
AntonKozlov: This is a good example. It may be even more useful if Dockerfile get added along the instruction. I haven't found one linked in docker hub

AntonKozlov: No urgency, not a request. Anyone can build desired one and use that anyway

PR #1401, opened 2019-06-07
Mmap stub

anton-bondarev:


src/compat/posix/sys/mman/mmap_stub.c
```diff
@@ -12,7 +12,14 @@
 
 #include <util/log.h>
 
+#include <module/embox/fs/syslib/idesc_mmap_api.h>
+
 void *mmap(void *addr, size_t len, int prot, int flags, int fd, off_t off) {
+	if(fd > 0){
```
Deryugin: `if(fd > 0){` -> `if (fd > 0) {`

anton-bondarev: Done

PR #1405, opened 2019-06-15
Fix timer init sequence

anton-bondarev:


src/kernel/time/jiffies.c
```diff
@@ -11,11 +11,14 @@
 #include <kernel/time/clock_source.h>
 #include <kernel/time/time.h>
 
-EMBOX_UNIT_INIT(module_init);
+//EMBOX_UNIT_INIT(module_init);
```
alexkalmuk: Probably just remove?

PR #1408, opened 2019-06-18
Third party mirrors

alexkalmuk:
* Remove multiple PKG_SOURCES/PKG_MD5. Now PKG_SOURCES is just a list of mirrors for a single package
* Get rid of PKG_ARCHIVE_NAME. Now each downloaded package will be saved to download/ folder with name combined of $(PKG_NAME)-$(PKG_VER).extension. Extension will be derived from url.
* When package downloading is failed for some reason remove partially downloaded package.

### DIFF
--- a/third-party/phoneme/Makefile
+++ b/third-party/phoneme/Makefile
@@ -3,7 +3,6 @@ PKG_NAME := phoneme
 PKG_VER  := trunk
 
 PKG_SOURCES := https://www.dropbox.com/s/13ihqov40oud914/phoneme-trunk.tar.gz
-PKG_ARCHIVE_NAME :=phoneme-trunk.tar.gz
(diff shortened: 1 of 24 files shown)


third-party/phoneme/Makefile
```diff
@@ -3,7 +3,6 @@ PKG_NAME := phoneme
 PKG_VER  := trunk
```
Deryugin: May be name `phoneme-trunk` will be more meaningful?

alexkalmuk: Hm, I see here

```
PKG_NAME := phoneme
PKG_VER := trunk
```

Do you propose just PKG_NAME := phoneme-trunk?

Deryugin: nvm, my mistake

PR #1412, opened 2019-06-20
Task foreach single

anton-bondarev:
Simplify task_foreach macro

### DIFF
--- a/src/include/kernel/task.h
+++ b/src/include/kernel/task.h
@@ -107,9 +107,9 @@ __END_DECLS
 
 #define task_foreach(tsk) \
 	for (int tid = 1; \
-			(tid = task_table_get_first(tid)) >= 0 \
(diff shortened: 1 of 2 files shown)


src/include/kernel/task.h
```diff
@@ -107,9 +107,9 @@ __END_DECLS
 
 #define task_foreach(tsk) \
 	for (int tid = 1; \
-			(tid = task_table_get_first(tid)) >= 0 \
-				&& (tsk = task_table_get(tid), assert(tsk != NULL), 1); \
-			tid = task_get_id(tsk) + 1)
+			(tid = task_table_get_first(tid)) >= 0 && \
+			NULL != (tsk = task_table_get(tid)); \
```
alexkalmuk: As far as we assume tid is correct, `NULL !=` is redundant.

anton-bondarev: Thanks, fixed

PR #1415, opened 2019-06-25
cmds/hardware/mem rework

Deryugin:
This command requires `-a` flag, which doesn't make much sense.

Now this command works without this flag, i.e.: `mem 0x90000000`, `mem 0xdeadbeef -n12` and so on.

### DIFF
--- a/src/cmds/hardware/mem.c
+++ b/src/cmds/hardware/mem.c
@@ -65,7 +65,6 @@ int main(int argc, char **argv) {
 	bool length_passed = false;
 	enum access_type at = MEM_AT_LONG;
 
-	getopt_init();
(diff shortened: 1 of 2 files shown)


src/cmds/hardware/mem.c
```diff
@@ -58,25 +58,18 @@ static int parse_option(char *optarg, int opt, long unsigned int *number) {
...
 int main(int argc, char **argv) {
-	bool a_flag = false;
 	int opt, i, ret;
 	void *address;
 	unsigned long val;
 	size_t length = DEFAULT_LENGTH;
+	bool length_passed = false;
 	enum access_type at = MEM_AT_LONG;
 
 	getopt_init();
```
anton-bondarev: rm it's deprecated now

PR #1419, opened 2019-07-02
Ld regions

alexkalmuk:
* Add ability to declare new memory regions and new linker sections in lds.conf
* Add ability to put .text, .rodata, .data and. bss sections to specified memory regions

I added brief [description](https://github.com/embox/embox/wiki/Mybuild%3A-LDS-user-defined-sections) to our wiki.

### DIFF
--- a/mk/gen_ld_sections.sh
+++ b/mk/gen_ld_sections.sh
@@ -1,0 +1,21 @@
+#!/bin/sh
+
+# Input
+CONFIG_LDS_H=$1
+# Output
+SECTIONS_LDS_H=$2
+
+rm -f $SECTIONS_LDS_H && touch $SECTIONS_LDS_H
+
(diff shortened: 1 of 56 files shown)


mk/gen_ld_sections.sh
```diff
@@ -0,0 +1,21 @@
...
+
+# Input
+CONFIG_LDS_H=$1
+# Output
+SECTIONS_LDS_H=$2
+
+rm -f $SECTIONS_LDS_H && touch $SECTIONS_LDS_H
+
+SECTIONS=`grep -o 'LDS_SECTION_VMA_[0-9a-zA-Z_]*' $CONFIG_LDS_H | \
+			sed 's/LDS_SECTION_VMA_//g' | uniq`
```
Deryugin: I guess `sort -u` is more appropriate than `uniq`

alexkalmuk: Sort would change the sections order.

PR #1425, opened 2019-07-07
Qt stm32 optimization

alexkalmuk:
* Enable Qt to run faster on stm32f7discovery
* Optionally put framebuffer into a separate linker section, not at the specified fixed address.


templates/arm/qt-stm32f7discovery/start_script.inc
```diff
@@ -1,4 +1,4 @@
 "export PWD=/",
 "export HOME=/",
 "export QT_QPA_FONTDIR=/fonts",
-"moveblocks -platform emboxfb",
+//"moveblocks -platform emboxfb"
```
anton-bondarev: Don't You  want to start automatically?

PR #1429, opened 2019-07-08
tests: add fgets_test

nastya-nizharadze:


src/tests/stdio/fgets_test.c
```diff
@@ -0,0 +1,49 @@
...
+/* File's content is
+* qwer\n
+* asdf\n
+* \n
+* zxcv
+*/
+#define TEST_FILE_NAME       "/test_fgets.txt"
+#define TEST_FILE_NAME_EMPTY "/test_fgets_empty.txt"
+
+#define STR_SIZE_TEST (6 * sizeof(char))
```
anton-bondarev: What is 6. A magic number?

src/tests/stdio/fgets_test.c
```diff
@@ -0,0 +1,49 @@
...
+
+TEST_CASE("fgets should read lines with \\n at the end") {
+	FILE *fp = fopen(TEST_FILE_NAME, "r");
```
anton-bondarev: Please separate variable declarations and initializations

src/tests/stdio/fgets_test.c
```diff
@@ -0,0 +1,49 @@
...
+	char *line = (char *) malloc(STR_SIZE_TEST);
```
anton-bondarev: where is free() for this memory?
Declare array instead of malloc looks better in this place

src/tests/stdio/fgets_test.c
```diff
@@ -0,0 +1,49 @@
...
+	test_assert_str_equal(line, "asdf\n");
+
+	line = fgets(line, STR_SIZE_TEST, fp);
+	test_assert_str_equal(line, "\n");
+
+	line = fgets(line, STR_SIZE_TEST, fp);
+	test_assert_str_equal(line, "zxcv");
+}
+
+TEST_CASE("fgets should return NULL") {
```
anton-bondarev: There is no file with this name?

src/tests/stdio/fgets_test.c
nastya-nizharadze: in file "test_fgets.txt" all lines is shorter or equals 4. +1 for \n and +1 because fgets manual says that fgets() reads in at most one less characters than the input size from file.

src/tests/stdio/fgets_test.c
nastya-nizharadze: Done.

src/tests/stdio/fgets_test.c
nastya-nizharadze: Done for first test case. For the second one i should use malloc. Otherwise there is an error in ```suggestion
	test_assert_null(line)
```: the address of ‘line’ will always evaluate as ‘true’

src/tests/stdio/fgets_test.c
nastya-nizharadze: There is a file "test_fgets_empty.txt"

PR #1432, opened 2019-07-09
stm32f769i port

Deryugin:
Split stmf7cube into two modules to support STM32F769i board


src/cmds/testing/fb_direct_access/fb_direct_access.c
```diff
@@ -51,7 +51,7 @@ int main() {
 				uint32_t t = ((r & 0xFF) << 16) |
 						((g & 0xFF) << 8) | (b & 0xFF);
 
-				((uint32_t *)fbp)[idx + x] = t;
+				((uint32_t *)fbp)[idx + x] = t | 0xff000000;
```
anton-bondarev: It's with alpha?

PR #1440, opened 2019-07-11
Libs: strxfrm: add implementation and test

nastya-nizharadze:


src/compat/libc/string/Mybuild
```diff
@@ -32,6 +33,7 @@ static module str {
 	source "strtok.c"
 	source "strlcpy.c"
 	source "strnlen.c"
+	depends embox.compat.libc.all
```
anton-bondarev: Why do you need this dependence? It has wider scope than 'string' package

PR #1443, opened 2019-07-15
opencv/edges: Pass file as an argument

Deryugin:
Also make a bunch of fixes (mostrly for file system)


platform/opencv/cmds/edges.cpp
```diff
@@ -20,11 +20,13 @@ using namespace std;
 static void help() {
```
anton-bondarev: help(void)

PR #1470, opened 2019-08-06
cmds: Fix telnetd race condition

anton-bondarev:


src/cmds/net/telnetd/telnetd.c
```diff
@@ -476,9 +477,9 @@ int main(int argc, char **argv) {
 						MD(printf("cannot execv process err=%d\n", errno));
 					}
 				}
+				sched_yield();
```
alexkalmuk: Please, do not forget to add the corresponding comment about this workaround :)

PR #1478, opened 2019-08-08
fixed uninitialized flag is_group_set which is compared after in some…

678098:
… execution branch


Deryugin:
Please, change commit message to fit our [git commit message aggrements](https://github.com/embox/embox/wiki/Contributing%3A-Git-commit-message-agreements), i.e. it should be something like
`cmds: chown: Fix uninitialized flag is_group_set`

Deryugin:
Merged in #1479

PR #1479, opened 2019-08-08
Minor fixes

678098:


Deryugin:
Please, change commit message to fit our [git commit message aggrements](https://github.com/embox/embox/wiki/Contributing%3A-Git-commit-message-agreements), i.e. it should be something like
`drivers: bluetooth/lego/blue_core4: Initialize sum by 0 in calc_chksumm`

PR #1484, opened 2019-08-13
X86 vfork stackframe

nastya-nizharadze:
backtrace now work everywhere except a few lines in `ptregs_jmp`, because in this function we jump on previous frame and change values of stack and registers.

### DIFF
--- a/src/arch/x86/include/asm/entry.h
+++ b/src/arch/x86/include/asm/entry.h
@@ -54,16 +54,22 @@
 	iret;
 
 #define CALL_WPTREGS \
+	push    %ebp;               \
+	mov     %esp, %ebp;         \
 	subl 	$28, %esp;          \
 	SAVE_ALL_REGS;              \
-	movl	PT_END(%esp), %ecx; \
+	movl    %esp, %eax;         \
+	addl    $PT_EBP, %eax;      \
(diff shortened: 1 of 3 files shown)


src/arch/x86/include/asm/entry.h
```diff
@@ -54,16 +54,22 @@
 	iret;
 
 #define CALL_WPTREGS \
+	push    %ebp;               \
+	mov     %esp, %ebp;         \
 	subl 	$28, %esp;          \
 	SAVE_ALL_REGS;              \
+	movl    %esp, %eax;         \
+	addl    $PT_EBP, %eax;      \
+	push    PT_OLD_EBP(%esp);   \
```
alexkalmuk: What does `PTL_OLD_EBP` means? Here `PT_<reg_name>` corresponds to offset within `struct pt_regs` for other registers. As I can see PTL_OLD_EBP is beyond pt_regs

nastya-nizharadze: for backtrace to iterate over frames we need to save ebp on top of frame. Also we need to move esp to ebp to save ebp in the next frame. But we need old ebp to be in pt_regs, so i use macros to hide magic offset number.

PR #1500, opened 2019-09-06
Usb

anton-bondarev:
* Add parsing USB options in auto_qemu
* Add auto_qemu_with_usb_disc script
* Add ehci_pci host controller driver

To check use [https://github.com/embox/embox/wiki/Using-USB-on-qemu](https://github.com/embox/embox/wiki/Using-USB-on-qemu)


templates/x86/qemu/system_start.inc
```diff
@@ -1,7 +1,7 @@
 "export PWD=/",
 "export HOME=/",
-"mkdir -v /bin",
-"mount -t binfs / /bin",
+//"mkdir -v /bin",
```
alexkalmuk: Why it is commented out?

templates/x86/qemu/system_start.inc
```diff
@@ -1,7 +1,7 @@
...
+//"mount -t binfs / /bin",
```
alexkalmuk: Why it is commented out?

scripts/qemu/auto_qemu_with_usb_storage
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+image_file_name="usbdisk.img"
+
+if [ -f "$file" ]
+then
+	echo "$image_file_name found."
+else
+	echo "$image_file_name not found."
+	qemu-img create -f qcow2 usbdisk.img 512M
```
alexkalmuk: I'd suggest to add message about device creation here.

```
echo "$image_file_name not found. A new one will be created now..."
```

scripts/qemu/auto_qemu_with_usb_storage
```diff
@@ -0,0 +1,15 @@
...
+fi
+
+sudo mkfs -t vfat $image_file_name
```
alexkalmuk: It's not good to have `sudo` in shell scripts. It's better to use sudo when calling script.

src/drivers/block_dev/scsi/scsi_disc_create.c
```diff
@@ -56,6 +56,10 @@ static void* scsi_create_thread(void *arg) {
 	}
 
 	bdev = block_dev_create(path, (void *) &bdev_driver_scsi, NULL);
+	if (!bdev) {
+		/* log_error("can't create block device"); */
```
alexkalmuk: Why is it commented? Probably use this log_error?

scripts/qemu/auto_qemu_with_usb_storage
alexkalmuk: Hm, if I run your script multiple times I got: `usbdisk.img not found` every time.

scripts/qemu/auto_qemu_with_usb_storage
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+image_file_name="usbdisk.img"
+
+if [ -f "$file" ]
```
alexkalmuk: `if [ -f $image_file_name ]`

scripts/qemu/auto_qemu_with_usb_storage
alexkalmuk: Add `if [ -f $image_file_name ]`, see above. It fixes the problem.

PR #1505, opened 2019-09-16
Improve FS usability

Deryugin:
Number of minor fixes for FS commands and modules

### DIFF
--- a/src/cmds/fs/mkfs.c
+++ b/src/cmds/fs/mkfs.c
@@ -110,6 +110,11 @@ int main(int argc, char **argv) {
 	fs_type = DEFAULT_FS_TYPE;
 	blocks = DEFAULT_BLOCK_QTTY;
 
+	if (argc < 2) {
+		print_usage();
+		return 0;
+	}
+
 	getopt_init();
 	while (-1 != (opt = getopt(argc, argv, "ht:q:F:"))) {
 		switch (opt) {
--- a/src/cmds/fs/mkfs.c
+++ b/src/cmds/fs/mkfs.c
@@ -118,7 +118,6 @@ int main(int argc, char **argv) {
 		return 0;
 	}
 
-	getopt_init();
 	while (-1 != (opt = getopt(argc, argv, "ht:q:F:"))) {
 		switch (opt) {
 		case 'F':
@@ -143,7 +142,6 @@ int main(int argc, char **argv) {
 			}
 
 			return mkfs_create_ramdisk(path, blocks);
-		case '?':
 		case 'h':
 			print_usage();
 			return 0;
--- a/src/cmds/fs/mkfs.c
+++ b/src/cmds/fs/mkfs.c
@@ -70,7 +70,10 @@ static int mkfs_do_operation(size_t blocks, char *path, const char *fs_name,
 			printf("File %s not found.\n", path);
 			return 0;
 		}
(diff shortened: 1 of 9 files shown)


src/cmds/fs/mount.c
```diff
@@ -83,53 +84,46 @@ static void show_mount_list(void) {
...
-
-	if (argc > 2) {
-		dev = argv[argc - 2];
-		dir = argv[argc - 1];
-		if (argc > 3) {
-			if (opt_cnt) {
-				fs_type = argv[argc - 3];
-			} else {
+			case 't':
+				fs_type_passed = true;
```
alexkalmuk: Why do not set `fs_type = optarg` here? And use `!fs_type` instead of `!fs_type_passed` below?

src/cmds/hardware/usb_whitelist.c
```diff
@@ -258,8 +259,14 @@ int main(int argc, char *argv[]) {
 		return 0;
 	}
 
-	strncpy(usb_wl_dev_path, "/dev/", USB_WL_DEV_PATH_LEN);
-	strncat(usb_wl_dev_path, USB_WHITELIST_DEV_NAME, USB_WL_DEV_PATH_LEN);
+	strncpy(usb_wl_dev_path,
```
alexkalmuk: I think we can just use `strcpy` instead, and then do `strncat` to the result which appends null byte. There is no need to use `strncpy` here since we explicitly assume the size of `"/dev/"` is lower than `sizeof(usb_wl_dev_path)`.

src/cmds/fs/mkfs.c
```diff
@@ -110,6 +113,11 @@ int main(int argc, char **argv) {
 	fs_type = DEFAULT_FS_TYPE;
 	blocks = DEFAULT_BLOCK_QTTY;
 
+	if (argc < 2) {
+		print_usage();
+		return 0;
+	}
+
 	getopt_init();
```
anton-bondarev: Rm getopt_init() now it's depricated

src/cmds/fs/mount.c
```diff
@@ -83,53 +84,46 @@ static void show_mount_list(void) {
...
+				break;
+			case '?':
+				break;
```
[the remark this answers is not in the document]
anton-bondarev: I think just rm break in this place

src/cmds/fs/mount.c
Deryugin: Done

src/cmds/fs/mount.c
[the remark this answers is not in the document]
Deryugin: Done

src/cmds/fs/mkfs.c
Deryugin: Done

src/cmds/hardware/usb_whitelist.c
Deryugin: I have no idea why GitHub shows this commit, it's already merged to master.

src/cmds/fs/mount.c
alexkalmuk: Thanks

src/cmds/hardware/usb_whitelist.c
alexkalmuk: Ah, ok

PR #1510, opened 2019-09-20
DE0 Nano SoC porting

Deryugin:
Add support for Terasic  DE0_Nano_SoC Version-A Board.


src/drivers/ambapp/ambapp_id.h
```diff
@@ -10,8 +10,6 @@
 #ifndef AMBAPP_H_
 #define AMBAPP_H_
 
-#include <module/embox/driver/ambapp_api.h>
```
anton-bondarev: Why is this required?
If it rm it can break sparc templates

Deryugin: fixed

PR #1542, opened 2019-11-01
x86: Multiboot header uses address fields

alexkalmuk:
It is introduced to overcome Qemu's slow BSS loading.

See the comment in code.


alexkalmuk:
I checked with grub2.

PR #1544, opened 2019-11-06
Fix live flash for qemu

alexkalmuk:
Fix ./scripts/live-flash to try bootloaders with Embox on QEMU:

For example, on x86/qemu:

Grub2:
```
$ ./scripts/live-flash grub2 grub2.img
$ sudo qemu-system-i386 -hda grub2.img -m 256 -nographic
```

Syslinux:
```
$ ./scripts/live-flash syslinux syslinux.img
$ sudo qemu-system-i386 -hda syslinux.img -m 256 -nographic
```

x86/debug seems to be broken, so I only checked in non-graphic mode currently.

### DIFF
--- a/scripts/live-flash
+++ b/scripts/live-flash
@@ -20,8 +20,15 @@ usage() {
 	echo "<type> is bootloader type: grub (for grub v1.0), grub2 or syslinux"
 	echo "<device> is dev file of flash."
 	echo
-	echo "Example:"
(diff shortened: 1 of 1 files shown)


alexkalmuk:
I added a small description to our wiki - https://github.com/embox/embox/wiki/QEMU-with-GRUB2-and-Syslinux

PR #1565, opened 2019-11-25
pty: Fix synchronization with tty

alexkalmuk:
Fix `Execute "help" multiple times` autotest test case, which was failing from time to time due to pty <-> tty synchronization.


src/compat/posix/idx/pipe_pty.c
```diff
@@ -275,6 +276,8 @@ static int pty_master_status(struct idesc *idesc, int mask) {
 		return 1;
 	}
 
+	mutex_lock(&t->lock);
+
 	switch (mask) {
 	case POLLIN:
 		res = ring_can_read(&pty_to_tty(pty)->o_ring, TTY_IO_BUFF_SZ, 1);
```
anton-bondarev: &t->o_ring looks better than &pty_to_tty(pty)->o_ring

alexkalmuk: Thank you, fixed.

PR #1581, opened 2019-12-09
Ramfs

anton-bondarev:


src/fs/driver/ramfs/ramfs_oldfs.c
```diff
@@ -512,3 +404,47 @@ static struct fs_driver ramfs_driver = {
 
 DECLARE_FILE_SYSTEM_DRIVER(ramfs_driver);
 
+/****************/
+EMBOX_UNIT_INIT(ramfs_ramdisk_fs_init); /*TODO*/
+
+#define RAMFS_DEV  "/dev/ram#"
+static char ramfs_dev[] = RAMFS_DEV;
+
+static int ramfs_init(void * par) {
```
Deryugin: I think it's a good idea just to remove this function, because it creates block device, which probably should be done somewhere else.

anton-bondarev: I agree, but I thought to do it in future PR

PR #1585, opened 2019-12-10
Rework oldfs file_operations structure

anton-bondarev:
Make identical struct file_operations for both fs (dvfs and oldfs)


Deryugin:
One more `struct node` left:
```
./src/tests/fs/vfs.c:69:static struct inode *vfs_lookup_nr(struct node *node, const char *str_path) {
```

PR #1586, opened 2019-12-10
drivers/block_dev: Avoid memory corruption for unaligned read/write ops

### LINKED ISSUE #623: harddisk_pio seems to misfunction with nsects greater than one
##### What steps will reproduce the problem?
1. e2e9d8e, x86/nonvga_debug, include embox.fs.driver.ntfs, run on qemu with hda as ntfs-formatted image
2. ntfs tries to read 19456 position of size 2 blocks
##### What is the expected output? What do you see instead?

irq on that write not asserted, and status register holding 0x58, which could be considered as precached read. However, if irq waiting skipped on DRDY and DRQ on, half of sector read right, another one consists from zero, which is spotted by ntfs.

Reading by one sector everytime seems to be ok (made in 7d3133b)
Cc: @fsulima, @sikmir

Deryugin:
Block devices operate with blocks; so if someone wants to read or write
a single byte, they should either provide large buffer or this should be
done somewhere in driver. Currently there are no such support in
drivers, so it's better to restrict such calls.

Closes #623

### DIFF
--- a/src/drivers/block_dev/block_dev.c
+++ b/src/drivers/block_dev/block_dev.c
@@ -228,6 +228,10 @@ int block_dev_read(void *dev, char *buffer, size_t count, blkno_t blkno) {
 	}
 	bdev = block_dev(dev);
 
+	if (count % bdev->block_size != 0) {
+		return -EINVAL;
+	}
+
 	if (blkno >= bdev->size / bdev->block_size) {
 		return -EINVAL;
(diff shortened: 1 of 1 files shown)


Deryugin:
Sorry, this problem is actually handled by `block_dev_read_buffered()`/`block_dev_write_buffered()`..

PR #1591, opened 2019-12-11
drivers/net/dwc_gmac: Add 1gbit support

Deryugin:
Add 1gbit speed support and move speed-related stuff to `dwc_set_speed()`


anton-bondarev:
I've checked it. It's still worked with usb-ethernet net-card.
But now telnetd and httpd don't work

PR #1596, opened 2019-12-12
libc/assert: Fix compilation for NDEBUG

Deryugin:


src/drivers/usb/class/usb_class_cdc.c
```diff
@@ -96,7 +98,9 @@ static void cdc_get_endpoints(struct usb_dev *dev, struct usb_class_cdc *cdc) {
 
 		for (j = 0; j < iface->b_num_endpoints; j++) {
 			ep = (struct usb_desc_endpoint *) endpoints + j;
-			assert(usb_endp_alloc(dev, ep));
+			if (0 == usb_endp_alloc(dev, ep)) {
+				log_error("usb_enp_alloc() failed");
```
anton-bondarev: panic() or assert(0);

PR #1600, opened 2019-12-16
tcp: Fix tcp_write to enable send msg with several iov

anton-bondarev:


src/net/socket/tcp_sock.c
```diff
@@ -374,39 +374,75 @@ static int tcp_accept(struct sock *sk, struct sockaddr *addr,
...
+			tcp_set_ack_field(skb->h.th, tcp_sk->rem.seq);
+			send_seq_from_sock(tcp_sk, skb);
+			cp_off = 0;
+		}
+	}
 
-		memcpy(skb->h.th + 1, pb, bytes);
-		pb += bytes;
-		len -= bytes;
+	if (0 != cp_off) {
```
alexkalmuk: Is it normal behavior that cp_off is not zero? I added `assert(0)` inside this if branch and didn't get it under telnet and httpd.

src/net/socket/tcp_sock.c
```diff
@@ -374,39 +374,69 @@ static int tcp_accept(struct sock *sk, struct sockaddr *addr,
...
-		 * We must pass no more than 64k bytes to underlaying IP level */
-		size_t bytes = min(len, IP_MAX_PACKET_LEN - MAX_HEADER_SIZE);
-		skb = NULL; /* alloc new pkg */
+	iov_len = 0;
+	skb_len = 0;
 
-		ret = alloc_prep_skb(tcp_sk, 0, &bytes, &skb);
-		if (ret != 0) {
-			break;
+	for (i = 0; (i < msg->msg_iovlen) || (0 != skb_len) || (0 == iov_len); ) {
```
alexkalmuk: Is it possible to use `while (tran_len < full_len) ` here instead? I think this condition should be equal to `(i < msg->msg_iovlen) || (0 != skb_len) || (0 == iov_len)`.

anton-bondarev: Thank a lot! You're right!

PR #1608, opened 2019-12-18
adxl345

Deryugin:
Add driver for adxl345 and simple command for continuously print sensor output

### DIFF
--- a/src/cmds/hardware/sensors/adxl345/adxl345.c
+++ b/src/cmds/hardware/sensors/adxl345/adxl345.c
@@ -1,0 +1,29 @@
+/**
+ * @file adxl345.c
+ * @brief Print adxl345 data in the endless loop
+ * @author Denis Deryugin <deryugin.denis@gmail.com>
+ * @version
+ * @date 18.12.2019
+ */
+
+#include <stdint.h>
+#include <stdio.h>
+
+#include <drivers/sensors/adxl345.h>
+
+int main(int argc, char **argv) {
+	printf("Testing adxl345 accelerometer output\n");
+
+	if (0 != adxl345_init()) {
+		printf("Failed to init ADXL345!\n");
+		return -1;
+	}
+
+	while (1) {
+		int x = adxl345_get_x();
+		int y = adxl345_get_y();
+		int z = adxl345_get_z();
+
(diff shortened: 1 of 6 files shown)


src/cmds/hardware/sensors/adxl345/adxl345.c
```diff
@@ -0,0 +1,18 @@
+#include <stdint.h>
```
alexkalmuk: File header?

src/cmds/hardware/sensors/adxl345/adxl345.c
```diff
@@ -0,0 +1,18 @@
...
+#include <stdio.h>
+
+#include <drivers/sensors/adxl345.h>
+
+int main(int argc, char **argv) {
+	printf("Testing adxl345 accelerometer output\n");
+
+	adxl345_init();
```
alexkalmuk: Shouldn't we check for result?

src/drivers/sensors/adxl345/adxl345.c
```diff
@@ -0,0 +1,91 @@
...
+#define ADXL345_DATA_AXIS(n)    (ADXL345_DATAX0 + (n))
+
+#define ADXL345_DEVID_VALUE     0xE5
+
+#define ADXL345_BUS    OPTION_GET(NUMBER, bus)
+#define ADXL345_OFFSET OPTION_GET(NUMBER, offset)
+
+struct adxl345_dev {
+	int i2c_bus;
+	int i2c_offset;
```
alexkalmuk: Why do not name this field `i2c_addr`? I think addr is more common for i2s devices than offset.

src/cmds/hardware/sensors/adxl345/adxl345.c
Deryugin: Added

src/cmds/hardware/sensors/adxl345/adxl345.c
Deryugin: Added check for successfull init

src/drivers/sensors/adxl345/adxl345.c
Deryugin: You're right. Changed it to `i2c_addr`

PR #1621, opened 2020-01-02
Samba extbld

anton-bondarev:


anton-bondarev:
@alexkalmuk Thanks, I'm agree. I've removed old version.

PR #1627, opened 2020-01-10
Clean ip stack

anton-bondarev:


src/include/net/l3/icmpv4.h
```diff
@@ -200,7 +200,7 @@ typedef struct icmphdr {
 		struct icmpbody_param_prob param_prob;
 		struct icmpbody_timestamp timestamp;
 		struct icmpbody_info info;
-	} __attribute__((packed)) body[];
+	} __attribute__((packed)) body;
 } __attribute__((packed)) icmphdr_t;
 
 #define ICMP_MIN_HEADER_SIZE (sizeof(struct icmphdr))
```
Deryugin: I think this value will be changed now, as I see `ping` command and `udp` module rely on this value.

anton-bondarev: thanks.
fixed ICMP_MIN_HEADER_SIZE to (4)

PR #1639, opened 2020-01-17
Use EMBOX_VERSION from make

anton-bondarev:


src/cmds/sys/uname.c
```diff
@@ -61,10 +60,14 @@ int main(int argc, char *argv[]) {
 		}
 	}
 
-	if (!uname_args_not_empty(&args))
+	if (!uname_args_not_empty(&args)) {
 		args.with_s = 1;
+	}
+
+	if (-1 == uname(&info)) {
+		printf("uname failed with code error %d", errno);
```
Deryugin: Should function `return 1;` here to avoid access to `info`?

anton-bondarev: thanks, done

Issue #1047, opened 2017-03-03
add RISC-V architecture support


frantony:
[RISC-V](https://www.riscv.org) is a new and very promising  open instruction set architecture.

kandeshvari:
Are there public MCs or CPUs available?

frantony:
There is a RISCV32 [HiFive1 board](https://www.sifive.com/boards/hifive1).
For RISCV64 please see [HiFive Unleashed board](https://www.crowdsupply.com/sifive/hifive-unleashed).

You can use mainline qemu for HiFive board emulation.

kandeshvari:
1. What arch is better to get as a base for porting? Is `generic` a stub for any new arch? 
2. What other code I have to write besides `src/arch`?

kandeshvari:
I'm trying to build minimal kernel as it described [here](http://web.archive.org/web/20160105032147/https://code.google.com/p/embox/wiki/EmboxArchPorting)

my template:
```
package genconfig

configuration conf {
	@Runlevel(2) include embox.arch.riscv.kernel.boot
	@Runlevel(2) include embox.arch.riscv.kernel.arch
	@Runlevel(2) include embox.arch.system(core_freq=3200000000)

	@Runlevel(2) include embox.kernel.spinlock(spin_debug=false)
	@Runlevel(2) include embox.driver.serial.sifive_uart(base_addr=0x10013000,baud_rate=115200,clock_freq=32000000)
	@Runlevel(2) include embox.driver.diag(impl="embox__driver__serial__sifive_uart")
	@Runlevel(2) include embox.driver.interrupt.no_interrupts
}
```
next:
```
make confload-riscv/qemu-sifive-e
make
```

but gain an error: 
```
In file included from src/s_fma.c:31:0:
/home/dk/devel/other/embox/build/extbld/third_party/lib/OpenLibm/openlibm-2252cbd02c176a31f74fa10f8fcdf926c3124987/include/openlibm_fenv.h:16:2: error: #error "Unsupported platform"
```

Do I really need libm for minimal kernel? How to switch it's build off?

alexkalmuk:
@kandeshvari Probably, you can try `include embox.compat.libc.math_libm_none`

anton-bondarev:
@kandeshvari And use add line
	`include embox.compat.libc.stdio.print(support_floating=0)`

for switch off  floating point support in 'print' routines

anton-bondarev:
@kandeshvari 

[...]

PR #1648, opened 2020-01-22
Riscv

nastya-nizharadze:
Add context switch

### DIFF
--- a/src/arch/riscv/kernel/context.c
+++ b/src/arch/riscv/kernel/context.c
@@ -1,11 +1,18 @@
-#include <kernel/panic.h>
+#include <stdint.h>
 #include <hal/context.h>
+#include <hal/ipl.h>
+#include <asm/regs.h>
+#include <string.h>
 
 void context_init(struct context *ctx, unsigned int flags,
 		void (*routine_fn)(void), void *sp) {
+	memset(ctx, 0, sizeof(*ctx));
 
-}
+	ctx->sp = (uint32_t) sp;
+	ctx->ra = (uint32_t) routine_fn;
+	ctx->mstatus = read_csr(mstatus);
 
-void context_switch(struct context *prev, struct context *next) {
-	panic("%s", __func__);
+	if (flags & CONTEXT_IRQDISABLE) {
+		ctx->mstatus &= ~(MSTATUS_MIE);
+	}
 }
--- a/src/arch/riscv/kernel/context.h
+++ b/src/arch/riscv/kernel/context.h
@@ -1,7 +1,35 @@
 #ifndef ARCH_RISCV_CONTEXT_H_
 #define ARCH_RISCV_CONTEXT_H_
 
+#include <stdint.h>
+
+#ifndef __ASSEMBLER__
+
 struct context {
+	uint32_t ra;
+	uint32_t sp;
+	uint32_t s[12];
+	uint32_t mstatus;
 };
+
+#endif
+
+#define CTX_RA      0x00
+#define CTX_SP      0x04
+#define CTX_S0      0x08
+#define CTX_S1      0x0C
+#define CTX_S2      0x10
+#define CTX_S3      0x14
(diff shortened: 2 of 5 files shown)


src/arch/riscv/kernel/context.h
```diff
@@ -1,7 +1,35 @@
...
+#ifndef __ASSEMBLER__
+
 struct context {
+	unsigned int ra;
+	unsigned int sp;
+	unsigned int s[12];
+	unsigned int mstatus;
 };
 
+#else
```
anton-bondarev: #endif ?
Do you want to hide definitions in none assembly code?

templates/riscv/qemu-sifive-u/mods.config
```diff
@@ -21,5 +22,9 @@ configuration conf {
 	include embox.compat.libc.math_builtins
 	include embox.kernel.spinlock(spin_debug = false)
 
+	include embox.kernel.task.resource.errno
+
 	include embox.test.stdlib.setjmp_test
+	include embox.test.kernel.timer_test
+	include embox.test.hal.context_switch_test(stack_sz=0x40)
```
anton-bondarev: Maybe you should switch on more multi-threading features?

src/arch/riscv/kernel/context.c
```diff
@@ -1,11 +1,15 @@
-#include <kernel/panic.h>
+#include <stdint.h>
 #include <hal/context.h>
+#include <hal/ipl.h>
+#include <asm/regs.h>
 
 void context_init(struct context *ctx, unsigned int flags,
 		void (*routine_fn)(void), void *sp) {
+	ctx->sp = (uint32_t) sp;
```
Deryugin: `ctx->sp` is `unsigned int`, but you cast `sp` to `(uint32_t)`, I recommend to use `uint32_t` explicitly for `ctx` fields (if they are 32 bit wide).

src/arch/riscv/kernel/context.c
```diff
@@ -1,11 +1,15 @@
-#include <kernel/panic.h>
+#include <stdint.h>
 #include <hal/context.h>
+#include <hal/ipl.h>
+#include <asm/regs.h>
 
 void context_init(struct context *ctx, unsigned int flags,
 		void (*routine_fn)(void), void *sp) {
```
Deryugin: I think that unused fields of `ctx` should be set to zero in this function to make sure there are no "garbage" in the register.
You can use `memset(ctx, 0, sizeof(*ctx))` at the beginning of the function.

src/arch/riscv/kernel/context_switch.S
```diff
@@ -0,0 +1,55 @@
...
+	sw      s7, CTX_S7(a0)
+	sw      s8, CTX_S8(a0)
+	sw      s9, CTX_S9(a0)
+	sw      s10, CTX_S10(a0)
+	sw      s11, CTX_S11(a0)
+	csrr    t6, mstatus
+	sw      t6, CTX_MSTATUS(a0)
+	
+	lw      sp, CTX_SP(a1)
+	nop
```
Deryugin: Are these `nop`s really required for correct context switch?

src/arch/riscv/kernel/context.h
nastya-nizharadze: Not really. Fixup done.

templates/riscv/qemu-sifive-u/mods.config
nastya-nizharadze: I think I'll do it later, after implementing vfork.

src/arch/riscv/kernel/context_switch.S
nastya-nizharadze: As it turned out, in RISC-V there is no slot delay, nor can the data hazard occur, so no.

anton-bondarev:
Work on #1047

PR #1657, opened 2020-01-30
Unify devfs

Deryugin:
* Use same handlers for `devfs_iterate()`
* Unify char_dev modules
* Unift serial/tty modules


src/drivers/serial/uart_dev.c
```diff
@@ -78,27 +65,40 @@ static int uart_setup(struct uart *uart) {
...
-	uart->tty = NULL;
-
 	if (uart_defparams) {
 		memcpy(&uart->params, uart_defparams, sizeof(struct uart_params));
 	} else {
 		memset(&uart->params, 0, sizeof(struct uart_params));
 	}
 
-	dlist_add_next(&uart->lnk, &uart_list);
+	ttys_register(uart->dev_name, uart);
```
anton-bondarev: May be check result and use log_error() if it failed

PR #1663, opened 2020-02-05
fs: Use common code to find fs_driver

Deryugin:
Also rename fs_driver structure to have same names in both VFS versions


src/fs/dvfs/fs_driver.h
```diff
@@ -0,0 +1,24 @@
...
+ * @author Denis Deryugin <deryugin.denis@gmail.com>
+ * @version
+ * @date 05.02.2020
+ */
+
+#ifndef FS_DRV_H_
+#define FS_DRV_H_
+
+#define FS_DRV_NAME_LEN   16
+
```
anton-bondarev: You should add using structure declarations:
```
struct block_dev;
struct super_block;
```

PR #1664, opened 2020-02-05
templates: Rename arm template stm32f764g->stm32f746g

anton-bondarev:


Deryugin:
Thanks for finding and fixing this typo!

We also have the same issue with some modules
```
> grep "STM32F764G" -rnil src
src/drivers/block_dev/stm32_sd/Mybuild
src/drivers/block_dev/stm32_sd/stm32f764g_discovery_sd.h
src/drivers/audio/stm32/stm32f764g_discovery_audio.h
src/drivers/audio/stm32/Mybuild
```

Also template
`./platform/pjsip/templates/stm32746g/`

PR #1665, opened 2020-02-05
fs/dvfs: Get rid of bdev_file from super_block

Deryugin:
This field was never actually used in any FS and also doesn't seem to be
useful at all. It's used just to bass block device pointer, so getting
rid of this field simplifies code.

Also remove related unused functions.

### DIFF
--- a/src/cmds/fs/mount.c
+++ b/src/cmds/fs/mount.c
@@ -66,9 +66,9 @@ extern struct dlist_head dentry_dlist;
 
 static void show_mount_list(void) {
 	struct dentry *d;
+	static const char *devfs_path = "/dev/";
 	char mount_path[DVFS_MAX_PATH_LEN];
 	char bdev_path[DVFS_MAX_PATH_LEN];
-	struct file_desc *bdev_file;
 
 	dlist_foreach_entry(d, &dentry_dlist, d_lnk) {
 		if (d->flags & DVFS_MOUNT_POINT) {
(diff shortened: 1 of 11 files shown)


src/fs/driver/ramfs/ramfs_dvfs.c
```diff
@@ -202,8 +199,8 @@ static int ramfs_fill_sb(struct super_block *sb, struct file_desc *bdev_file) {
 	memset(fsi, 0, sizeof(struct ramfs_fs_info));
 	fsi->block_per_file = MAX_FILE_SIZE / PAGE_SIZE();
 	fsi->block_size = PAGE_SIZE();
-	fsi->numblocks = block_dev(bdev_file->f_inode->i_data)->size / PAGE_SIZE();
-	fsi->bdev = block_dev(bdev_file->f_inode->i_data);
+	fsi->numblocks = bdev>size / PAGE_SIZE();
```
alexkalmuk: `bdev->size`?

src/cmds/fs/mount.c
```diff
@@ -68,21 +68,15 @@ static void show_mount_list(void) {
...
 
-			bdev_file = d->d_sb->bdev_file;
-
-			if (bdev_file && bdev_file->f_dentry) {
-				dentry_full_path(bdev_file->f_dentry, bdev_path);
-			} else {
-				strcpy(bdev_path, d->d_sb->fs_drv->name);
-			}
+			strcpy(bdev_path, "/dev");
+			strcat(bdev_path, d->d_sb->fs_drv->name);
```
alexkalmuk: Probably it's better to use `strncat` here?

src/fs/driver/initfs/initfs_dvfs.c
```diff
@@ -210,13 +210,13 @@ struct inode_operations initfs_iops = {
...
+static int initfs_fill_sb(struct super_block *sb, struct block_dev *bdev) {
 	sb->sb_iops = &initfs_iops;
 	sb->sb_fops = &initfs_fops;
 	sb->sb_ops  = &initfs_sbops;
 	sb->bdev = NULL;
 
-	if (bdev_file) {
+	if (bdev) {
 		return -1;
 	}
```
alexkalmuk: Does it mean that initfs accepts only `NULL` as bdev pointer? If so, can we return here immediately without filling super block, because block device is incorrect?

Deryugin: Yes, if block dev is passed to initfs, then caller is doing something wrong.

I'll place this check upper in the function

src/fs/driver/ramfs/ramfs_dvfs.c
Deryugin: Fixed

PR #1673, opened 2020-02-07
fs: Change mount handlers signature

Deryugin:
Use `char *` instead of `struct block_dev *` as a source of FS data as
different FS require different parameters which makes common VFS part a
bit complex. Instead, driver should decide how to interpret mount
argument.

Also `umount` handlers now should work with `struct inode *` argument instead of `void *` which all the time was interpreted as `struct nas *`

### DIFF
--- a/src/fs/syslib/kfsop.c
+++ b/src/fs/syslib/kfsop.c
@@ -347,8 +347,28 @@ int kformat(const char *pathname, const char *fs_type) {
 	return 0;
 }
 
-int kmount(const char *dev, const char *dir, const char *fs_type) {
-	struct path dev_node, dir_node, root_path;
+struct block_dev *bdev_by_path(const char *source) {
+	struct path dev_node;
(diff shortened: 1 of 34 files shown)


src/fs/syslib/kfsop.c
```diff
@@ -347,8 +347,29 @@ int kformat(const char *pathname, const char *fs_type) {
...
+		return NULL;
+	}
+
+	if (ENOERR != (res = fs_perm_lookup(source, &lastpath, &dev_node))) {
+		return NULL;
+	}
+
+	if (ENOERR != (res = fs_perm_check(dev_node.node, S_IROTH | S_IXOTH))) {
+		return NULL;
+	}
```
alexkalmuk: `res` is not used so we can remove this variable.

Deryugin: Done

PR #1674, opened 2020-02-07
fs: Split VFS mount handlers

Deryugin:
Now use VFS-agnostic `fill_sb` handler to initliaze super block from given source (this handler can be used in both VFS implementations) and `mount` itself which mounts all the files.

### DIFF
--- a/src/include/fs/super_block.h
+++ b/src/include/fs/super_block.h
@@ -15,6 +15,9 @@ struct fs_driver;
 struct inode_operations;
 struct super_block_operations;
 
+/* Stubs */
+struct super_block_operations { };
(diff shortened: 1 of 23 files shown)


src/include/fs/super_block.h
```diff
@@ -9,12 +9,17 @@
 #ifndef SUPER_BLOCK_H_
 #define SUPER_BLOCK_H_
 
+#include <fs/inode_operation.h>
```
anton-bondarev: Why is it necessary ?

Deryugin: Fixed

PR #1683, opened 2020-02-12
Mount rework

Deryugin:
A bunch of changes for VFS:
* Don't use dentries at all for FS drivers. It's achieved by:
  * Now each superblock has inode root pointer instead of dentry pointer. Previously it was used just to reference to inode date (e.g. `sb->root->d_inode->data = new_data;`)
  * `.lookup()` handlers now use inode as well
* Simplify `mount()`
  * Create and destroy root inode during super_block_alloc()/super_block_free(), so no need to do it in every FS driver
  * Get rid of `.mount_end()` handler. It turned out that all the stuff from this handler could be done in `.fill_sb()` if root inode is provided
* Finally, `struct super_block` is the same for both VFS versions
* Minor fixes for memory leaks and other bugs

### DIFF
--- a/src/fs/driver/ramfs/ramfs_oldfs.c
+++ b/src/fs/driver/ramfs/ramfs_oldfs.c
@@ -120,6 +120,7 @@ static int ramfs_format(struct block_dev *bdev, void *priv) {
 }
 
 struct inode_operations ramfs_iops;
+struct super_block_operations { };
 struct super_block_operations ramfs_sbops;
 static int ramfs_mount(struct super_block *sb, struct inode *dest) {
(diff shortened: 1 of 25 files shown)


src/fs/driver/ramfs/ramfs_oldfs.c
```diff
@@ -120,6 +120,7 @@ static int ramfs_format(struct block_dev *bdev, void *priv) {
 }
 
 struct inode_operations ramfs_iops;
+struct super_block_operations { };
```
anton-bondarev: Empty struct?

PR #1687, opened 2020-02-16
Add l3g4200d gyro (spi mode) to DE0 Nano SOC

alexkalmuk:
Fix DW SPI driver - in polling mode without DMA we have to fill tx fifo with data faster than fifo becomes empty, because this SPI controller automatically controls CS line state and if tx fifo becomes empty CS go high (inactive) state. Also, spi1 and i2c1 share the same GPIO pins, so I commented spi1 in mods.config and added comments. L3G4200D gyro currently works on SPI1 on DE0 Nano SOC.

### DIFF
--- a/src/drivers/spi/dw_spi/Mybuild
+++ b/src/drivers/spi/dw_spi/Mybuild
@@ -14,4 +14,5 @@ module dw_spi {
 
 	depends core
 	depends embox.driver.periph_memory
+	depends embox.driver.gpio.api
 }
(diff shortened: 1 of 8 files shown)


src/drivers/spi/dw_spi/Mybuild
```diff
@@ -6,12 +6,13 @@ module dw_spi {
 	option number base_addr2 = 0
 	option number base_addr3 = 0
 
-	option number log_level = 0
+	option boolean spi_de0_nano_soc = false
```
Deryugin: Do we really need this option? If so, then we need it for dw_i2c too, I think.

alexkalmuk: There are several possible ways as we discussed:

* Use SPI option (as currently)
* Use GPIO module option like `spi_switch=<something>`, `spi_switch_enable=true`.
* Add a new module for the specific board configuration, which enables SPI or I2C.

PR #1693, opened 2020-02-18
imx6 ehci

anton-bondarev:
Now usb flash drive works correctly on x86/qemu with ehci_pci host controller and iwave_imx6 board.
Implementation restrictions:
* EHCI works only with single QH
* QH is allocated for each endpoint (pipe) ones during creating.
* QH async list is activated/deactivate each time to adding QH
* Only async list using (not periodic)
* Each hub has only one device


alexkalmuk:
I checked this on imx6 Iwave - flash drive though usb hub.

PR #1700, opened 2020-02-22
Minor bug fixes

sunn-e:
- The <!DOCTYPE> declaration tells the web browser which (X)HTML version is being used on the page, and therefore how to interpret the various elements.
- The <html> element should provide the lang and/or xml:lang attribute in order to identify the default language of a document.

It enables assistive technologies, such as screen readers, to provide a comfortable reading experience by adapting the pronunciation and accent to the language. It also helps braille translation software, telling it to switch the control codes for accented characters for instance.


anton-bondarev:
Thanks!
But now there is one trouble in travis-ci.
It's a problem in our test (x86/test/net) in which we use different (not sync) index.html. The problem have to fix before this PR will be approved.

anton-bondarev:
@sunn-e I fix the CI problem in #1703. Please rebase under it.

PR #1703, opened 2020-02-22
ci: Use 'index.html' from test folder for page download test

anton-bondarev:


alexkalmuk:
It's related to https://github.com/embox/embox/pull/1700

AntonKozlov:
Why do you have to copy the file? It's "gold" file to compare with downloaded one. The file to download was in template/rootfs/, but moved in 4d615a4e92

Anyway, it looks like this `cp` do not work https://travis-ci.org/embox/embox/jobs/658848904#L5427

AntonKozlov:
Hey, this `cp` does nothing, index.html is not there. https://travis-ci.org/embox/embox/jobs/658848904#L5427
Tests run from embox root dir. 

I suppose the problem is that served and outdated gold files are different. You are trying to serve old gold file. 
Why not just refresh the gold file?

anton-bondarev:
> I suppose the problem is that served and outdated gold files are different. You are trying to serve old gold file.
> Why not just refresh the gold file?

@AntonKozlov What do you mean about refreshing the gold file?

The problem is

1. the test downloads a file from web server.
2. the test compares the downloaded file with a file from your directory (gold file)

if the file on web server is changed we must update test file too.

To avoid this we must synchronize them. We can or copy from web server (root) directory to test directory (gold file) or copy from test directory (gold file) to web server (root) directory.

if there are some problems lets copy from web server (root) directory to test directory (gold file)
Does It solve the problem?

AntonKozlov:
> To avoid this we must synchronize them. We can or copy from web server (root) directory to test directory (gold file) or copy from test directory (gold file) to web server (root) directory.

Do you mean copy when test runs? Or update the gold file once and commit?
I selected second option to avoid calculating where is the file in build directories of embox (does it come from conf/rootfs? @InitFS?)

> if there are some problems lets copy from web server (root) directory to test directory (gold file)

This will be better. It will save us make invocation.
If you want to copy during test run and you know a path to a file that embox will serve, I suggest to compare against that path directly without needless copy to test directory.

> Does It solve the problem?

The problem is that this PR did nothing. More precisely, now `cp` provides an error and does nothing. Served file is not updated.
I suggest to fix the problem and do a tiny optimization at once.

PR #1707, opened 2020-02-24
Move xen headers to third-party

lkm30:


anton-bondarev:
It looks very great. But It's really required to modify mk/extbld/lib.mk ?

third-party/xen/Mybuild
```diff
@@ -0,0 +1,4 @@
+package third_party
+
+@Build(stage=1,script="$(EXTERNAL_MAKE) BUILD_DIR=$(abspath $(EXTERNAL_BUILD_DIR))/third_party/xen download install")
```
AntonKozlov: Can you remove setting `BUILD_DIR`? 
I think you'll get `$(EXTERNAL_BUILD_DIR))/third_party/xen_headers` implicitly. Is something wrong with it?

third-party/xen/Makefile
```diff
@@ -0,0 +1,16 @@
...
+PKG_SOURCES := git://xenbits.xen.org/xen.git
+
+GIT_CLONE_ARGS := -b $(PKG_VER)
+
+PKG_MD5 :=
+
+include $(EXTBLD_LIB)
+
+$(INSTALL) :
+	test -d "$(INCUDE_INSTALL_DIR)/xen" && rm -rf "$(INCUDE_INSTALL_DIR)/xen"; \
```
AntonKozlov: Nice! It's more straightforward than I expected

third-party/xen/Makefile
```diff
@@ -0,0 +1,16 @@
+
+PKG_NAME := xen
+PKG_VER  := stable-4.13
+
+PKG_SOURCES := git://xenbits.xen.org/xen.git
```
AntonKozlov: Can we use source archive bundle, i.e. from https://xenproject.org/downloads/xen-project-archives/xen-project-4-13-series/xen-project-4-13-0/ (https://downloads.xenproject.org/release/xen/4.13.0/xen-4.13.0.tar.gz)

There is a problem with git sources, it requires a whole mini-language to specify how to clone. For example,  revision/branch `aaa` of project A and revision `bbb` of project B. And for one one of them you want `--depth 1`. I don't know even how could it be specified in the Makefile prettily.

So we tend to use source archive bundles if possible. It looks like it's possible here too.

lkm30:
It is necessary because the current check does not work for git sources.
Also, a .download file is created even if all sources fail.
I did it in wrong way at first, but now it should work.

src/arch/xen/event.c
```diff
@@ -29,42 +28,6 @@ void hypervisor_callback(void);
 void failsafe_callback(void);
 
 
-static evtchn_handler_t handlers[NUM_CHANNELS];
```
AntonKozlov: This is nice too. Having only this code made a feel of a lot of duplicates. There is still some commented code (`init_events` in kernel_init.c), redundant `xen_shared_info` in traps.c. But that's all that I can easily spot. This change is a big step forward.

third-party/xen/Makefile
lkm30: They don't share md5 sum. Is this correct to use self-computed sum?

AntonKozlov: Yes, of course, this is perfectly fine

lkm30: Done in #1721.

Issue #1696, opened 2020-02-21
Wrong backtrace on ARM


anton-bondarev:
To reproduce you can add `assert(0);` or simply add a breakpoint in gdb to critical_dispatch() and look at stacktrace.
for breakepoint:
```
(gdb) backtrace 
#0  critical_dispatch_pending () at src/kernel/critical.c:22
#1  0x00003dc4 in interrupt_handle ()
    at src/drivers/interrupt/integrator_pic/integrator_pic.c:81
#2  0x0005b7f0 in irq_handler () at src/arch/arm/armlib/exceptions/entry.S:61
#3  0x00000004 in arm_trap_table ()
    at src/arch/arm/armlib/exceptions/exception_table.S:37
(gdb) 
```

for assert(0):
```
void critical_dispatch_pending(void) {
	struct critical_dispatcher **pp = cpudata_ptr(&dispatch_queue);
	unsigned int count = critical_count();
	struct critical_dispatcher *d;
	unsigned int mask;
	ipl_t ipl;

	assert(0);
	ipl = ipl_save();

	while ((d = *pp) && !((mask = d->mask) & count)) {
		*pp = d->next;
		d->mask = ~mask;

		assert(d->dispatch != NULL);
		d->dispatch();
	}

	ipl_restore(ipl);
}```
```
 ASSERTION FAILED on CPU 0
	at src/kernel/critical.c:28
	in function critical_dispatch_pending

0
Data abort exception!
regs:
00090078 01c99c5c e59ff018 fffffff8
00124000 00000000 00000000 00000000
00000000 00000000 00000000 01c99c5c
0000001f 00002e3c
```

**Comment by bartekxk** [comment]
Hello, I would like to work on this issue but, what is exactly wrong with this backtrace?

Thanks.

**Comment by alexkalmuk** [comment]
I will try to explain this on another example:

I used the following patch to add `assert(0)` in src/cmds/net/ifconfig.c:

```diff
diff --git a/src/cmds/net/ifconfig.c b/src/cmds/net/ifconfig.c
index dbd9ea0..a2df5c9 100644
--- a/src/cmds/net/ifconfig.c
+++ b/src/cmds/net/ifconfig.c
@@ -285,6 +285,8 @@ int main(int argc, char *argv[]) {
        struct ifconfig_args args;
        void *addr;
 
+       assert(0);
+
        iface = NULL; /* XXX required for -O2 */
        memset(&args, 0, sizeof args);
```

[...]

PR #1711, opened 2020-02-24
tempates: Add -mapcs-frame flag to arm templates

alexkalmuk:
`-mapcs-frame` gcc option is used to make ARM stack frame in ACPS format, so we can do backtrace. It requires a bit of additional memory per each frame. So, I updated some of "big" ARM tempaletes so backtrace should work out of the box.


alexkalmuk:
I forgot to mention that it is related to https://github.com/embox/embox/issues/1696

Issue #1713, opened 2020-02-24
Implement memalign for heap_simple memory allocator


alexkalmuk:
There are several memory allocation algorithms in Embox, one of them is heap_simple - the very basic one, which is rarely used but is useful sometimes for debugging. However, the current heap_simple implementation misses `memalign()` function, so it's required to implement it.

bartekxk:
Hello, I would like to work on this issue.

alexkalmuk:
Hello, great!

bartekxk:
I created pull request. Can you review it?
 #1715 
Thanks.

bartekxk:
You can close this task.

alexkalmuk:
Yes, of course, thanks. btw you will use something like "Close #<issue>" or "Fix #<issue>" it will be closed automatically - https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

alexkalmuk:
Closed by https://github.com/embox/embox/pull/1715

PR #1715, opened 2020-02-24
[mem]Implement memalign for heap_simple memory allocator

bartekxk:
Implement memalign function for heap_simple.c. #1713 

The function works similar to malloc but it aligns the address. I check if the block size is big enough for alignment and if the alignment is power of 2 and after finding the area - I will align it.

### DIFF
--- a/src/mem/heap/heap_simple.c
+++ b/src/mem/heap/heap_simple.c
@@ -22,6 +22,7 @@ static void *heap_simple_end;
 
 #define HEAP_SIZE OPTION_MODULE_GET(embox__mem__heap_api,NUMBER,heap_size)
 #define HEAP_ALIGN OPTION_MODULE_GET(embox__mem__heap_api,NUMBER,align)
+#define ALIGN_PADDING sizeof(size_t)
 
 #define MEM_POOL_SIZE  (HEAP_SIZE / PAGE_SIZE())
 static void *mem_pool;
@@ -84,8 +85,10 @@ static void _mem_defrag(void) {
 
 void free(void *ptr) {
 	struct mem_control_block *mcb;
+	size_t shift = *(int*)(ptr - ALIGN_PADDING);
+	ptr -= MCB_SZ + shift + ALIGN_PADDING;
 
-	mcb = (void *)((char *) ptr - MCB_SZ);
+	mcb = ptr;
 	mcb->is_available = 1;
 
 	 _mem_defrag();
@@ -93,10 +96,15 @@ void free(void *ptr) {
 }
 
 void *malloc(size_t size) {
+	return memalign(sizeof(void*), size);
+}
+
+void *memalign(size_t boundary, size_t size) {
 	void *ptr;
 	struct mem_control_block *mcb, *next_mcb;
+	size_t shift, total_mem_size;
 
-	if (size == 0) {
+	if (size == 0 || (boundary & (boundary - 1)) != 0) {
 		return NULL;
 	}
 
@@ -112,20 +120,22 @@ void *malloc(size_t size) {
 
 	while (ptr + MCB_SZ + size < heap_simple_end) {
 		mcb = ptr;
+		shift = (boundary - ((int)(ptr + MCB_SZ + ALIGN_PADDING) % boundary)) % boundary;
+		total_mem_size = size + ALIGN_PADDING + shift + MCB_SZ;
 
-		if (mcb->is_available && mcb->size >= size) {
+		if (mcb->is_available && mcb->size >= total_mem_size - MCB_SZ) {
 			mcb->is_available = 0;
 
-			if (mcb->size > size + MCB_SZ) {
-				next_mcb = ptr + size + MCB_SZ;
+			if (mcb->size > total_mem_size) {
+				next_mcb = ptr + total_mem_size;
 
-				next_mcb->size = mcb->size - size - MCB_SZ;
+				next_mcb->size = mcb->size - total_mem_size;
 				next_mcb->is_available = 1;
 
-				mcb->size = size;
+				mcb->size = total_mem_size - MCB_SZ;
 			}
-
-			return ptr + MCB_SZ;
+			*(int*)(ptr + MCB_SZ + shift) = shift;
+			return ptr + total_mem_size - size;
 		}
 
 		ptr += mcb->size + MCB_SZ;


alexkalmuk:
Thanks @bartekxk 

First of all, how did you verified your changes do what expected?

alexkalmuk:
To test this algorithm you should include it in conf/mods.config instead of heap_bm I guess.

src/mem/heap/heap_simple.c
```diff
@@ -134,6 +134,48 @@ void *malloc(size_t size) {
 	return 0;
 }
 
+void *memalign(size_t boundary, size_t size) {
```
alexkalmuk: Now you have two functions - `memalign` and `malloc` which do almost the same things. We should use memalign in malloc instead.

src/mem/heap/heap_simple.c
```diff
@@ -134,6 +134,48 @@ void *malloc(size_t size) {
...
+	void *ptr;
+	struct mem_control_block *mcb, *next_mcb;
+
+	if (size == 0 || (boundary & (boundary - 1)) == 1) {
```
alexkalmuk: What does `(boundary & (boundary - 1)) == 1` mean? It just verifies that `boundary` is odd, isn't it?

alexkalmuk: But the boundary is required to be a power of two.

src/mem/heap/heap_simple.c
bartekxk: Done :).

src/mem/heap/heap_simple.c
bartekxk: It exactly checks if it is power of 2. I added one fix in my second commit and I check if it is different from 0.

bartekxk:
> To test this algorithm you should include it in conf/mods.config instead of heap_bm I guess.
@alexkalmuk 
Thanks. I tested it looks like the system is working properly.

src/mem/heap/heap_simple.c
alexkalmuk: >  I added one fix in my second commit and I check if it is different from 0.

Yes, it's right.

src/mem/heap/heap_simple.c
```diff
@@ -93,10 +93,14 @@ void free(void *ptr) {
 }
 
 void *malloc(size_t size) {
+	return memalign(1, size);
```
alexkalmuk: Great! But aligning memory to the 1-byte boundary is not a correct way, since 

```
The address of the allocated memory will be a multiple of alignment, which
       must be a power of two and a multiple of sizeof(void *)
```

you can read about this here https://linux.die.net/man/3/memalign

It's very usual behavior when the returned address is word aligned (word size can be 4 or 8 bytes).

src/mem/heap/heap_simple.c
```diff
@@ -125,7 +129,7 @@ void *malloc(size_t size) {
 				mcb->size = size;
 			}
 
-			return ptr + MCB_SZ;
+			return (void *)(((size_t)ptr + MCB_SZ + boundary -1) & ~(boundary-1));
```
alexkalmuk: Are you sure it's correct?

Look, what memory structure we have:
```
( MCB0 | object0 ) ( MCB1 | object1 ) ....
```

where MCB (Memory Control Block) is `struct mem_control_block`.

In more details:

```
( is_available0; size0 | object0 ) ( is_available1; size1 | object1 ) ...
```

Here, size0 is the size of object0, size1 is the size of object1, etc.

Ok, but what happens when you add `boundary` to ptr here?

```
( is_available0; size0 | <shift to boundary> object0 ) ( is_available1; size1 | object1 ) ...
```

So because now the actual size of object0 is _not_ size0 like it was before, but is `<shift to boundary> + size0`

src/mem/heap/heap_simple.c
alexkalmuk: Yes, I like how it looks now :) Only small comment about `memalign(1, ` below.

src/mem/heap/heap_simple.c
bartekxk: I think it's correct because we choose the memory block in which our offset fits and return the aligned address. Everything in this structure remains as before, we have only a few bytes of unused space before the address.

src/mem/heap/heap_simple.c
bartekxk: So why is this wrong? 1 is power of 2 and when we pass 1 (I added one more simple fix to check size condition) we do it like simple malloc.

alexkalmuk: But 1 is not a multiple of sizeof(void *), isn't it?

src/mem/heap/heap_simple.c
bartekxk: Ok, now I understand, you have right, I will fix it.

src/mem/heap/heap_simple.c
bartekxk: Sorry, I misunderstood this sentence. So we should pass sizeof (void *) in malloc? And additionally check this condition in memalign?

alexkalmuk: Yes, thank you! But it's required only for malloc() to be aligned.

src/mem/heap/heap_simple.c
```diff
@@ -93,10 +93,15 @@ void free(void *ptr) {
...
+	return memalign(sizeof(void*), size);
+}
+
+void *memalign(size_t boundary, size_t size) {
 	void *ptr;
 	struct mem_control_block *mcb, *next_mcb;
+	size_t shift;
 
-	if (size == 0) {
+	if (size == 0 || (boundary & (boundary - 1)) != 0 || boundary % sizeof(void*) != 0) {
```
alexkalmuk: I think we can remove `boundary % sizeof(void*) != 0` now, because this aligment reqired only for malloc()

src/mem/heap/heap_simple.c
alexkalmuk: So it was my mistake, I told you that is required by memalign, but it's required by malloc(). So we should pass this alignment (sizeof (void *) to memalign only to implement malloc).

bartekxk: Done :).

src/mem/heap/heap_simple.c
bartekxk: Done.

src/mem/heap/heap_simple.c
alexkalmuk: The amount of commits doesn't matter right now, do not worry. They can be squashed into a single one then.

I will take a look right now, this part of the code is not very simple, so we should be careful.

alexkalmuk: Yes, it is correct now.

alexkalmuk: But it seems like `free()` will broken now.

alexkalmuk: Because of that:

```
void free(void *ptr) {
	struct mem_control_block *mcb;

	mcb = (void *)((char *) ptr - MCB_SZ);
...
```

bartekxk: Maybe we can add a parameter to mcb "shifted" and then read it when we use free?

alexkalmuk: But how you will get the pointer to mcb only from the ptr passed into free() in this case? The shift is variable in size. We should save some information along with the allocated ptr to obtain mcb then.

bartekxk: At now we can do it like this
MCB_SZ shift pointer                  
and return ptr so I want do it like this
shift MCB_SZ pointer
We can read mcb and free aligned bytes. I will test it.

alexkalmuk: For example,

```
mcb <padding> object
```

So, `address of mcb = object - <padding> - MCB_SZ`. But the padding here depends on the shift value you calculated before in mmemalign. So probably you should do something like this:

```
mcb <padding> <shift value> object
```

So, you can get shift value first from ptr, and then get mcb..

alexkalmuk: Yes, or try to do it in the way you proposed. Any solution would be great.

alexkalmuk: Yes, please try your variant first.

alexkalmuk: But in your variant is another potential problem - with _mem_defrag() called from free(). Here we go over all free blocks and concatenate them.

```
shift0 mcb0 ...
```

how we can get the first mcb0?

bartekxk: I think, your solution is better and less complicated, I will use it :).

bartekxk: I added a fix, could you take a look? Are all tests from test cmd enough?

Thanks.

alexkalmuk: It looks great now!

alexkalmuk:
@bartekxk I tested your changes with

```
    @Runlevel(2) include embox.mem.static_heap(heap_size=0x8000000)
    include embox.mem.heap_simple(heap_size=0x4000000)

    @Runlevel(2) include embox.test.mem.heap   -- works fine
    @Runlevel(2) include embox.test.mem.heap_extended_test -- doesn't work, but may be it is another issue
```

I think since `embox.test.mem.heap` works fine and your changes look well, we can now prepare your PR to merge into master.

Can you please squash all your commits into a single one and rebase on the top of embox master?

You can also take a look at
https://github.com/embox/embox/wiki/Contributing%3A-Git-workflow

bartekxk:
> @bartekxk I tested your changes with
> 
> ```
>     @Runlevel(2) include embox.mem.static_heap(heap_size=0x8000000)
>     include embox.mem.heap_simple(heap_size=0x4000000)
> 
>     @Runlevel(2) include embox.test.mem.heap   -- works fine
>     @Runlevel(2) include embox.test.mem.heap_extended_test -- doesn't work, but may be it is another issue
> ```
> 
> I think since `embox.test.mem.heap` works fine and your changes look well, we can now prepare your PR to merge into master.
> 
> Can you please squash all your commits into a single one and rebase on the top of embox master?
> 
> You can also take a look at
> https://github.com/embox/embox/wiki/Contributing%3A-Git-workflow

Done :).

alexkalmuk:
Well, it looks good to me! I will merge in a bit later.

alexkalmuk:
I rebased on the top of embox master.

PR #1726, opened 2020-02-26
driver/clock: (minor) Adjust CLOCK_RATE in integrator_pit

### LINKED ISSUE #1708: Time flows too fast on arm/qemu
Template: `arm/qemu`
Include module `include embox.cmd.testing.ticker` in conf/mods.config and rebuild embox.

```
root@embox:/#ticker                                                             
1 sec
2 sec
3 sec
4 sec
5 sec
```

You can see time flows faster than in reality.

layman-n-ish:
The CLOCK_RATE is set to the cpu_core_freq which is supplied during the configuration of arm/qemu

Fixes #1708

### DIFF
--- a/src/drivers/clock/integrator_pit/integrator_pit.c
+++ b/src/drivers/clock/integrator_pit/integrator_pit.c
@@ -22,8 +22,8 @@
 /* Interrupt vector for timer (TMR1) */
 #define CLOCK_IRQ	5
 
-/* The clock rate per second - 1Mhz */
-#define CLOCK_RATE	10000000L
+/* The clock rate per second */
+#define CLOCK_RATE	48054841L
 
 /* The initial counter value */
 #define TIMER_COUNT	(CLOCK_RATE / HZ)


layman-n-ish:
Previously, the [CLOCK_RATE](https://github.com/embox/embox/blob/master/src/drivers/clock/integrator_pit/integrator_pit.c#L26) was set to 10000000 (10 MHz) as opposed to the current rate of 48054841 (~48 MHz). Ergo, due to the low rate earlier, the ticker clock expired rather quickly - since the initial counter value ([TIMER_COUNT](https://github.com/embox/embox/blob/master/src/drivers/clock/integrator_pit/integrator_pit.c#L29)) is set as:
```
/* The initial counter value */
#define TIMER_COUNT	(CLOCK_RATE / HZ)
```

This generated a premature timed interrupt for [usleep()](https://github.com/embox/embox/blob/master/src/cmds/testing/ticker/ticker.c#L64) in `ticker` which projected the behaviour of the clock ticking faster than reality.

anton-bondarev:
@layman-n-ish Thank you!

Issue #1702, opened 2020-02-22
service_run() does not consider result of starting service


anton-bondarev:
```
static int service_run(const char *path, char *const argv[]) {
	pid_t pid;
	pid = vfork();

	if (pid < 0) {
		int err = errno;
		printf("vfork() error(%d): %s", err, strerror(err));
		return -err;
	}

	if (pid == 0) {
		execv(path, argv);
		exit(1);
	}
	service_add(path, pid);

	return 0;
}
```

in spite of execv() can be broken service always adding to services list

It's blocked by #1335

worldinmyfist:
Hi can I work on this issue?
I need a starting point if you could help me.

Deryugin:
@worldinmyfist sure! Basicly, you should make sure that process is properly created with `execv()` and print something meaningful if something went wrong.

Codestyle and git workflow can be found on our wiki: https://github.com/embox/embox/wiki

worldinmyfist:
Hi it would be great if you could tell me in which file is this function located?

Deryugin:
@worldinmyfist `service_run()` is in src/cmds/service.c. I advise to use `grep` command to find functions quickly.
```
> grep service_run -rn ./src/
./src/cmds/service.c:92:static int service_run(const char *path, char *const argv[]) {
./src/cmds/service.c:133:	return service_run(command, &argv[1]);
```
`service_run` argument is used to pass the pattern you're looking for, `-r` (recousrive) flag makes it go into directories recoursively, `-n` (number) makes it print line number where pattern was matched

worldinmyfist:
Hey 
I went through the code and did not find any possible errors except in execv() function
Thus if issue #1335 is resolved, this issue should be solved.
I want you to assign me other issue so that I can work on that .
Thank you

worldinmyfist:
The command kill(parent, KILLSIG) runs when execv() fails. When this happens, service_add() will not get executed, and thus will not be added to the service list. Thus one of the solutions can be execv() returning to previous running environment after failing. Thus if #1335 is solved this issue should be solved.

alexkalmuk:
@worldinmyfist Ok, it seems it's actually blocked by #1335, you can choose another issue.

Ahish9009:
Hi!
Could I pls get some more info on how to test if this is working properly. I am testing whether my fix to #1335 is working, and thought I'd try it out on this too.

alexkalmuk:
[...]

PR #1734, opened 2020-02-26
service_run() can be fixed in this way if execv() behaves normally

nilCux:
service_run() does not consider result of starting service #1702

Based on POSIX execv(), the command `kill(parent, KILLSIG);`  will run only if execv() fails. When KILLSIG is sent, service_add() will not run, so unlanuched service will not be added to service list.

Should be solved in this way, only if execv() returns to previous running environment when it fails. I will work on execv() to realise that if you agree.

### DIFF
--- a/src/cmds/service.c
+++ b/src/cmds/service.c
@@ -89,7 +89,7 @@ static void listprint(void) {
 	}
 }
 
-static int service_run(const char *path, char *const argv[]) {
+static int service_run(const char *path, char *const argv[], pid_t father) {
 	pid_t pid;
 	pid = vfork();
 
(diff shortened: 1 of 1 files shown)


nilCux:
service_run() does not consider result of starting service #1702
Should be solved in this way, only if execv() returns to previous running environment when it fails. I will work on execv() to realise that if you agree.

anton-bondarev:
Thank you for your contribution. But  #1702 has already assigned with another student.

If you want to participate in Embox GSoC2020 you should send email to one of the mentors and chose other issue as described on the wiki page ([https://github.com/embox/embox/wiki/Google-Summer-of-Code-2020](https://github.com/embox/embox/wiki/Google-Summer-of-Code-2020))

nilCux:
Sorry for causing inconvenience.

Issue #1729, opened 2020-02-26
Implement 'du' command


anton-bondarev:
man du

Implement only for regular files and it's enough to follow flags:
```
      -B, --block-size=SIZE
              scale  sizes  by  SIZE  before printing them; e.g., '-BM' prints
              sizes in units of 1,048,576 bytes; see SIZE format below

       -b, --bytes
              equivalent to '--apparent-size --block-size=1'
```

bartekxk:
Hello, I would like to work on this issue.

alexkalmuk:
Hi, yes you can, of course :)

bartekxk:
Should the command work on a single file or recursively?

alexkalmuk:
Hm, why recursively? :) You should study a bit more what "regular file" from the issue description mean. For example, you can figure out how to check whether a file is a regular file on your Linux host.

But on the other hand, probably we can use `du` not only for regular files, but it's better to discuss it later.

alexkalmuk:
Anyway, you can start with regular files, without directories.

bartekxk:
I was thinking about recursive reading du documentation on another unix. Currently I did recursively by default with the option of choosing a file (regular file) / directory but I can change it if you want, for example with the -R parameter. I wrote the code, if you want you can see it here:
https://github.com/bartekxk/embox/commit/72ea9b329fc1caed01ae9d5596ed417742385124

This is almost the whole command but I have one problem. When I refer to st_blocks from the stat structure to check how many blocks file is using, I always get 0. Has reading of this parameter (by the stat function) not been implemented?

alexkalmuk:
Your version looks good!

>  When I refer to st_blocks from the stat structure to check how many blocks file is using, I always get 0. Has reading of this parameter (by the stat function) not been implemented?

I don't know if `st_blocks` is implemented correctly... Probably @anton-bondarev will answer.

But for `-b` it's better to refer to `st_size` rather than `st_blocks`.

alexkalmuk:
On Linux:
```
$ touch 1.txt
$ echo -n 0123456789 > 1.txt
$ du -b 1.txt 
10	1.txt
```

So it would be impossible to get this 10 bytes value from `st_blocks`.

anton-bondarev:
> This is almost the whole command but I have one problem. When I refer to st_blocks from the stat structure to check how many blocks file is using, I always get 0. Has reading of this parameter (by the stat function) not been implemented?

[...]

PR #1736, opened 2020-02-27
 Implement 'du' command

bartekxk:
Implement 'du' command for regular and directories. #1729

### DIFF
--- a/src/cmds/fs/du.c
+++ b/src/cmds/fs/du.c
@@ -1,0 +1,143 @@
+/**
+ * @file
+ * @brief
+ *
+ * @date 26.02.20
+ * @author Bartlomiej Kocot
+ */
+
+#include <errno.h>
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+#include <dirent.h>
+#include <limits.h>
+#include <stdlib.h>
+#include <sys/stat.h>
+
+#define BLK_SIZE 512
+
+static void print_usage(void) {
+	printf("Usage: du [-b BYTES] [-B BLOCK_SIZE] [FILE_OR_DIR] ...\n");
+}
+
+long long divide_with_round_up(long long size, size_t block_size) {
+	long long  result = size / block_size;
+	return result * block_size == size ? result : result + 1;
+}
+
+long long print_file_mem_usage(const char *path, struct stat *sb, size_t block_size) {
+	long long size = divide_with_round_up((long long)sb->st_size, block_size);
+	printf("%-21llu %s\n", size, path);
+	return size;
+}
+
+long long print(char *path, DIR *dir, size_t block_size) {
+	struct dirent *dent;
+	long long dir_size = 0;
+	char *line;
+
+	while (NULL != (dent = readdir(dir))) {
+		int pathlen = strlen(path);
+		int dent_namel = strlen(dent->d_name);
+		line = (char*)malloc(sizeof(char) * (pathlen + dent_namel + 3));
+		if (line == NULL) {
+			printf("Failed to allocate memory for buffer!\n");
+			return -1;
+		}
+		struct stat sb;
+
+		if (pathlen > 0) {
+			sprintf(line, "%s/%s", path, dent->d_name);
+		} else {
+			strcpy(line, dent->d_name);
+		}
+
+		if (-1 == stat(line, &sb)) {
+			printf("Cannot stat %s\n", line);
+			free(line);
+			continue;
+		}
+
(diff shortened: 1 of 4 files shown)


alexkalmuk:
It looks fine! 

Let's start with the following things to prepare your contribution:

1. Please fix setting of your github account - https://github.com/embox/embox/pull/1731#issuecomment-592389321

2. Fix code formatting (I mean tabs/spaces) please. You can find more here - https://github.com/embox/embox/wiki/Code-Style

3. Add your command only to `x86/qemu` (and maybe to another `*/qemu` templates, if you want). Configurations in Embox are very different, some of them do not contain FS support at all, so please  do not add your command to such a large list of configurations :)

bartekxk:
> 
> 
> It looks fine!
> 
> Let's start with the following things to prepare your contribution:
> 
>     1. Please fix setting of your github account - [#1729 (comment)](https://github.com/embox/embox/issues/1729#issuecomment-592397685)
> 
>     2. Fix code formatting (I mean tabs/spaces) please. You can find more here - https://github.com/embox/embox/wiki/Code-Style
> 
>     3. Add your command only to `x86/qemu` (and maybe to another `*/qemu` templates, if you want). Configurations in Embox are very different, some of them do not contain FS support at all, so please  do not add your command to such a large list of configurations :)

I think done.

alexkalmuk:
I tried `du`, it works fine for both regular files and directories! Can I now ask you to squash commits into a single one? Also, it's better to add this command only in `templates/x86/qemu/mods.config` for now.

bartekxk:
I squashed my commits but I have a problem with merge commits. Sorry, but I always have problems with forked repositories.

alexkalmuk:
> squashed my commits but I have a problem with merge commits. Sorry, but I always have problems with forked repositories.

It's normal, you can try to make git rebase like specified here https://github.com/embox/embox/wiki/Contributing%3A-Git-workflow#synchronize-with-the-upstream and just leave only your last commit https://github.com/embox/embox/pull/1736/commits/f264de673e2e0ab9aa33b59d9f8041e81eb4a1a0 on the top.

bartekxk:
Thanks, done.

src/cmds/fs/du.c
```diff
@@ -0,0 +1,128 @@
...
+}
+
+long long print(char *path, DIR *dir, size_t block_size) {
+	struct dirent *dent;
+	long long dir_size = 0;
+
+	while (NULL != (dent = readdir(dir))) {
+		int pathlen = strlen(path);
+		int dent_namel = strlen(dent->d_name);
+		char line[pathlen + dent_namel + 3];
```
alexkalmuk: We are trying to avoid using variable-length arrays because most compilers use on-stack allocations for VLA arrays, which may lead to unexpected stack overflows. The same troubles correspond to `alloca` too, and you can think about VLA arrays like about `alloca` btw. It's really harder to check whether on-stack allocation failed rather than just check if malloc() failed. Moreover, recursive calling the functions with `alloca` (or VLA arrays) make the program behavior more dangerous. You also can read more about why we should avoid using `alloca` here - https://stackoverflow.com/questions/1018853/why-is-the-use-of-alloca-not-considered-good-practice

How to overcome the issue with VLA arrays? There are probably several solutions, and I will point only two of them:

* Using the buffer of predefined size. You can declare `char line[64];` and then copy names into this buffer using `strncpy`. In your case since the buffer is only used to get a concatenation of several strings, you can define it once in `main` and then pass to your `print` function. Or for simplicity - just to declare `static char line[<something>];` and use it.
* Using `malloc/free`. I think all is clear about that.

alexkalmuk:
> Thanks, done.

Great! Let's fix VLA arrays usage and I think it could be merged.

src/cmds/fs/du.c
bartekxk: Sure, I took this fragment from the ls function so I corrected it there as well.

alexkalmuk: Ok, thanks.

alexkalmuk: btw, it would be better to check what malloc() returned to make sure it's not a NULL pointer.

bartekxk: Done but I'm not sure if in a good way.

alexkalmuk: I think it's ok :)

Two remarks:

* There is no need to write `sizeof(char)` in malloc
* It's much better to have separate commits for each module. For instance, you could create one commit for du, and one for ls.

But let's leave it as is so I could merge your contribution.

PR #1737, opened 2020-02-27
cmds/hardware/lsusb: Extending lsusb with -v flag

### LINKED ISSUE #1714: Extend lsusb command
Add the following option to lsusb command:

(from `man lsusb`)
```
       -v, --verbose
              Tells lsusb to be verbose and display detailed information about the devices shown.
              This  includes  configuration  descriptors  for  the device's current speed.  Class
              descriptors will be shown, when available, for USB device  classes  including  hub,
              audio, HID, communications, and chipcard.
```

We already have `lsusb` in Embox, but it only prints current devices, it would be great if we can see also device descriptors at least.

This issue requires to study USB, but only very basic high-level things, without going deeper into the USB subsystem. You have to study how to get a USB descriptor and show this descriptor to user in a nice way.

shivamanipatil:
Closes #1714 with printing device, configuration and interface descriptors for -v flag.

### DIFF
--- a/src/cmds/hardware/lsusb/lsusb.c
+++ b/src/cmds/hardware/lsusb/lsusb.c
@@ -12,8 +12,11 @@
 #include <drivers/usb/usb.h>
 
 static void print_usage(void) {
-	printf("Usage: lsusb [-h]\n");
+	printf("Usage: lsusb [-h] [-v]\n");
 	printf("\t[-h]      - print this help\n");
+	printf("\t[-v]      - prints verbose output with device, configuration" 
+		   " and interface descriptors\n");
+
 }
 
 static void print_error(void) {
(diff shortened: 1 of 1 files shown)


alexkalmuk:
Thank you, it looks good.

src/include/drivers/usb/usb.h
```diff
@@ -160,6 +160,7 @@ struct usb_dev {
 	void *config_buf;
 	struct usb_desc_device dev_desc;
 	struct usb_desc_interface iface_desc;
+	struct usb_desc_configuration conf_desc;
```
alexkalmuk: You can use the beginning of `config_buf` instead:
```
config = (struct usb_desc_configuration *) dev->config_buf;
```

alexkalmuk: Just because `conf_desc` is not filled anywhere

alexkalmuk:
Can I also ask you please to try `lsusb -v` with https://github.com/embox/embox/wiki/Using-USB-on-qemu?

src/include/drivers/usb/usb.h
shivamanipatil: I needed conf_desc in lsusb.c for displaying configuration descriptors

alexkalmuk: Yes, I understand, but who fills this `conf_desc`?

shivamanipatil: Sorry, but I didnt get what do you mean by filling the struct

alexkalmuk: You defined new field `conf_desc` in the structure. This field is uninitialized. I think you can try to use `lsusb -v` on https://github.com/embox/embox/wiki/Using-USB-on-qemu as I proposed below, and see what will happen.

shivamanipatil: Thank you, I used config_buf as you mentioned and removed conf_desc from struct usb_dev. I also put a check on config for null values as "Bus 000 Device 000 ID 0000:0000" in qemu returns null pointer. And  I tried `lsusb -v` on https://github.com/embox/embox/wiki/Using-USB-on-qemu it works.

alexkalmuk: Now it looks good, thank you.

src/cmds/hardware/lsusb/lsusb.c
```diff
@@ -28,18 +31,104 @@ static void show_usb_dev(struct usb_dev *usb_dev) {
...
+			config->i_configuration,
+			config->bm_attributes,
+			config->b_max_power);	
+}
+
+
 int main(int argc, char **argv) {
 	struct usb_dev *usb_dev = NULL;
-	int opt;
+	int opt, flag = 0;
```
alexkalmuk: btw the variable name `flag` is not very good, it's better to use `flag_verbose` at least..

Issue #1740, opened 2020-02-28
Incorrect output of more command on a short file


alexkalmuk:
```
root@embox:/#more hosts

*** many newlines goes here ***

127.0.0.1 localhost 
```

These new lines shouldn't appear on the screen.

MrSquanchee:
Hii, Alex kalmuk.
I would like to work on this issue.
And may you brief me on how to reproduce the bug and which files are the culprit.

anton-bondarev:
@MrSquanchee Ok, I've assigned it with you.

> And may you brief me on how to reproduce the bug and which files are the culprit.
It's described in issue
```
root@embox:/#more hosts

*** many newlines goes here ***

127.0.0.1 localhost 
```
When you start Embox you see command line and you have to create 'more' command which can be called from the command line.
You should start from our [wiki Quick-Start](https://github.com/embox/embox/wiki/Quick-Start)

layman-n-ish:
Hi, I had looked into this previously; maybe some of my observations could be useful to y'all. 

I think the `*** many newlines goes here ***` part is just `curses'` attempt to flush the screen (which is of size [80 * 24](https://github.com/embox/embox/blob/master/src/compat/posix/curses/curses.c#L29) - hence we see 24 `\n` chars) when `initscr()` is called - [here](https://github.com/embox/embox/blob/master/src/compat/posix/curses/curses.c#L112), the stdscr's buffer is being initialized with whitespaces and when `doupdate()` is [called](https://github.com/embox/embox/blob/master/src/compat/posix/curses/curses.c#L124) after this, the attempt to create a flushed screen starts which can be seen [here](https://github.com/embox/embox/blob/master/src/compat/posix/curses/curses.c#L246)! So, changing `' '` to `0` in [this](https://github.com/embox/embox/blob/master/src/compat/posix/curses/curses.c#L112) line, i.e. to initialize the buffer with zeros rather than whitespaces (just like the virtual screen's buffer is [initialized](https://github.com/embox/embox/blob/master/src/compat/posix/curses/curses.c#L114)), should remove those multiple newlines although I'm not sure if that is the behaviour required. 

Some useful references I found while reading more about `ncurses`:
[1] [man - newterm](https://linux.die.net/man/3/newterm)
[2] [man - doupdate](https://linux.die.net/man/3/doupdate)
[...]

PR #1760, opened 2020-03-02
Solved Issue #1740 Incorrect output of more command on short files

MrSquanchee:
Removed endwin call. Which was called before the screen write process.
Corrected the EOF detection which was faulty as random characters filled
the screen.
Replaced space with zero in function newterm in file curses.c
in the memset command.
more commands works correctly for short files.

### DIFF
--- a/src/cmds/fs/more.c
+++ b/src/cmds/fs/more.c
@@ -28,29 +28,36 @@ static void screen(FILE *fp) {
 		columns = MAX_SCREEN_WIDTH - 1;
 	}
 	lines = std->endy - std->begy;
-	endwin();
 
 	while (1) {
 		for (x = 0; x < lines - 1; x++) {
 
 			for (y = 0; y < columns; y++) {
-				buff[y] = getc(fp);
-				switch ((int) buff[y]) {
-				case EOF:
+				if (fread(&buff[y], 1, 1, fp) == 1) {
+
+					switch ((int) buff[y]) {
+
+						case '\n':
+							/*	End of the line, filling the rest of buffer */
+							memset(buff + y, ' ', columns - y);
+							y = columns;
+							break;
+
+						case '\t':
+							/*	Perform tab instert	*/
+							memset(buff + y, ' ', TAB_SIZE);
+							y += TAB_SIZE - 1;
+							break;
+
+					}
+
+				} else {
+
 					for (i = 0; i < y; i++) {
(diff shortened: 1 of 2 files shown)


Deryugin:
No need to create new pull request for new fixes :) Let's keep working in this PR, just push new commits to it.

src/cmds/fs/more.c
```diff
@@ -28,29 +28,37 @@ static void screen(FILE *fp) {
 		columns = MAX_SCREEN_WIDTH - 1;
 	}
 	lines = std->endy - std->begy;
-	endwin();
+	//endwin();
```
Deryugin: If this code is unused now you can safely delete it (that's another reason why we use VCS afterall).

src/cmds/fs/more.c
```diff
@@ -28,29 +28,37 @@ static void screen(FILE *fp) {
...
 	while (1) {
 		for (x = 0; x < lines - 1; x++) {
 
 			for (y = 0; y < columns; y++) {
-				buff[y] = getc(fp);
-				switch ((int) buff[y]) {
-				case EOF:
-					for (i = 0; i < y; i++) {
-						printf("%c", buff[i]);
+				if(fread(&buff[y], 1, 1, fp) == 1) {
```
Deryugin: Need to add 1 space here, `if(` -> `if (`. There's no need to hurry up, so take your time to format the code properly.

src/cmds/fs/more.c
```diff
@@ -28,29 +28,37 @@ static void screen(FILE *fp) {
...
+
+						case '\t':
+							/*	Perform tab instert	*/
+							memset(buff + y, ' ', TAB_SIZE);
+							y += TAB_SIZE - 1;
+							break;
+
 					}
+
+				}else {
```
Deryugin: Please add a space between `}` and `else`

src/cmds/fs/more.c
```diff
@@ -28,29 +28,37 @@ static void screen(FILE *fp) {
...
+
+					for (i = 0; i < y; i++) {
+							printf("%c", buff[i]);
```
Deryugin: Please remove extra tab on these two lines

Deryugin:
I've checked your changes and it works fine both for short and long files.

So you need to fix code style issues, then squash commits and it will be ready for merge.

PR #1762, opened 2020-03-02
mk: makefile: Fix stack-size error on Mac

Ahish9009:
Added code to change the stack size on a Mac without raising any error.

Code first checks if it is being run on a Mac. If so, it finds the maximum stack size limit allowed by the hardware and sets it with ```ulimit -s STACK_SIZE_LIMIT```.

```STACK_SIZE_LIMIT```is provided by ```ulimit -Hs```.


Deryugin:
Thank your for this patch!
It should work fine, but just for double check I'll try to build it on some Mac (don't have any nearby at the moment, so wait for day or two please).

PR #1763, opened 2020-03-03
command `ls` multiple arguments #1755

### LINKED ISSUE #1755: Multiple arguments for `ls` command
Looks like `ls` uses only first non-option argument, so it should be fixed to accept more files as arguments.

Now it works like this:
```
root@embox:/#ls -l shadow
-rw-------       root       root        398 shadow
root@embox:/#ls -l passwd
-rw-r--r--       root       root        335 passwd
root@embox:/#ls -l shadow passwd
-rw-------       root       root        398 shadow
```

How is it expected to work:
```
root@embox:/#ls -l shadow passwd
-rw-------       root       root        398 shadow
-rw-r--r--       root       root        335 passwd
```

This can be tested on `x86/qemu` or `arm/qemu` templates.

pdzhu:
command `ls` multiple arguments fix
Fixes #1755

### DIFF
--- a/src/cmds/fs/ls.c
+++ b/src/cmds/fs/ls.c
@@ -168,27 +168,37 @@ int main(int argc, char **argv) {
 	if (optind < argc) {
 		struct stat sb;
 
-		if (-1 == stat(argv[optind], &sb)) {
+		do {
+			if (-1 == stat(argv[optind], &sb)) {
+				return -errno;
+			}
+
+			if (!S_ISDIR(sb.st_mode)) {
+				printer(argv[optind], &sb);
+				continue;
+			}
+
+			snprintf(dir_name, NAME_MAX, "%s", argv[optind]);
+
+			if (NULL == (dir = opendir(dir_name))) {
+				return -errno;
+			}
+
+			print(dir_name, dir, recursive, printer);
+	
+			closedir(dir);
+		} while (optind++ < argc - 1);
+	} else {
+		strcpy(dir_name, ".");
+
+		if (NULL == (dir = opendir(dir_name))) {
 			return -errno;
 		}
 
-		if (!S_ISDIR(sb.st_mode)) {
-			printer(argv[optind], &sb);
-			return 0;
-		}
+		print(dir_name, dir, recursive, printer);
 
-		sprintf(dir_name, "%s", argv[optind]);
-	} else {
-		strcpy(dir_name, ".");
+		closedir(dir);
 	}
-
-	if (NULL == (dir = opendir(dir_name))) {
-		return -errno;
-	}
-
-	print(dir_name, dir, recursive, printer);
-
-	closedir(dir);
 
 	return 0;
 }


alexkalmuk:
The result looks fine!

src/cmds/fs/ls.c
```diff
@@ -168,27 +168,40 @@ int main(int argc, char **argv) {
...
+		while (optind++ < argc - 1) {
+
+		    if (-1 == stat(argv[optind], &sb)) {
+			    return -errno;
+            }
+
+            if (!S_ISDIR(sb.st_mode)) {
+                printer(argv[optind], &sb);
+                continue;
+		    }
```
alexkalmuk: Can I please ask you to follow our code style - https://github.com/embox/embox/wiki/Code-Style. In this part of code something wrong with tabs/spaces.

alexkalmuk: The same below.

pdzhu: yeah，I also found it but it looks right on my computer. Is it because of a mix of tabs and spaces?

src/cmds/fs/ls.c
```diff
@@ -168,27 +168,40 @@ int main(int argc, char **argv) {
...
-		if (-1 == stat(argv[optind], &sb)) {
-			return -errno;
-		}
+		optind--;
 
-		if (!S_ISDIR(sb.st_mode)) {
-			printer(argv[optind], &sb);
-			return 0;
-		}
+		while (optind++ < argc - 1) {
```
alexkalmuk: Why do we need `optind--` here? optind should point to the first file already. Probably we can use `do .. while` here?

src/cmds/fs/ls.c
alexkalmuk: yes, most likely because of that

src/cmds/fs/ls.c
pdzhu: I also feel strange but I didn't remember for a while. After I modified all of them， do I need make a another PR. Sorry,  I'm not familiar with this

alexkalmuk: No, you should push your new changes to this one

src/cmds/fs/ls.c
```diff
@@ -168,27 +168,40 @@ int main(int argc, char **argv) {
...
+		    if (-1 == stat(argv[optind], &sb)) {
+			    return -errno;
+            }
+
+            if (!S_ISDIR(sb.st_mode)) {
+                printer(argv[optind], &sb);
+                continue;
+		    }
+
+            sprintf(dir_name, "%s", argv[optind]);
```
alexkalmuk: It's not safe to do `sprintf` here, because your dir_name is only `NAME_MAX` in size, while the length of argv[optind] can be greater.

pdzhu: So do I need to add a judgement？

alexkalmuk: Sorry, I do not understand what judgement means in this context, but I was thinking about `snprintf`.

pdzhu: Is you mean this ?    snprintf(dirname,  NAME_MAX, "%s", argc[optind])
Here are two options， one is telling people the length of path name or file name is too long. The other one is  to avoid making mistake but people don't realize the results match the command.

pdzhu: I  am all right. Could you check it？Thanks for your great code review. It's my first time and I learned a lot. Thanks again.

alexkalmuk: It's a good question about how to handle long names, my point of view here is only to make long names handling to be safe. Your current variant with strncpy is good enough.

alexkalmuk:
@pdzhu I will run your changes a bit later, and if everything will be ok then we can merge PR.

Can you please squash your commits into a single one?

pdzhu:
I do 'git reset --soft HEAD~3 && git commit'. Is that right?

alexkalmuk:
> I do 'git reset --soft HEAD~2 && git commit'. Is that right?

It's a possible way in this particular case, but this one is far better - https://github.com/embox/embox/wiki/Contributing%3A-Git-cheatsheet#squash-commits-into-a-single-one

pdzhu:
> > I do 'git reset --soft HEAD~2 && git commit'. Is that right?
> 
> It's a possible way in this particular case, but this one is far better - https://github.com/embox/embox/wiki/Contributing%3A-Git-cheatsheet#squash-commits-into-a-single-one

I have done it. So what should I do now?

PR #1766, opened 2020-03-03
x86/qemu: multiple argumets for cat command

### LINKED ISSUE #1754: Multiple arguments for `cat` command
Looks like `cat` command simply prints content of the last argument (i.e. ignoring all the other arguments).

What do we have now:
```
root@embox:/#cat file1 file2
<file2 content>
```

What do we need:
```
root@embox:/#cat file1 file2
<file1 content>
<file2 content>
```

This can be tested on `x86/qemu` or `arm/qemu` templates.

sobhaniiest:
"Fixes #1754"
multiple_arguments_for_cat

### DIFF
--- a/src/cmds/fs/cat.c
+++ b/src/cmds/fs/cat.c
@@ -21,14 +21,8 @@ int main(int argc, char **argv) {
 	int number = 0, line = 0, new_line = 1;
 	FILE *fd;
 	char buff;
-	int no_of_arg;
-	char *file_names[argc];
-	size_t arg_len;
+	int i;
 
-	for (no_of_arg = 0; no_of_arg < argc; no_of_arg++) {
-		file_names[no_of_arg] = strdup (argv[no_of_arg]);
-	}
-	
 	if (argc < 2) {
 		print_usage();
 		return -EINVAL;
@@ -50,10 +44,10 @@ int main(int argc, char **argv) {
 		}
 	}
 
-	for (no_of_arg = 1; no_of_arg < argc; no_of_arg++) {
-		if (strcmp ("-n", file_names[no_of_arg]) != 0) {
-			if (NULL == (fd = fopen(file_names[no_of_arg], "r"))) {
-				printf("Can't open file %s\n", file_names[no_of_arg]);
+	for (i = 1; i < argc; i++) {
+		if (strcmp("-n", argv[i]) != 0) {
+			if (NULL == (fd = fopen(argv[i], "r"))) {
+				printf("Can't open file %s\n", argv[i]);
 				return -errno;
 			}
 
@@ -74,11 +68,6 @@ int main(int argc, char **argv) {
 
 			fclose(fd);
 		}
-	}
-
-	for (no_of_arg = 0; no_of_arg < argc; no_of_arg++) {
-		char *file_name = file_names[no_of_arg];
-		free (file_name);
 	}
 
 	return 0;
--- a/src/cmds/fs/cat.c
+++ b/src/cmds/fs/cat.c
@@ -8,9 +8,9 @@
 
 #include <unistd.h>
 #include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
 #include <errno.h>
-
 
 static void print_usage(void) {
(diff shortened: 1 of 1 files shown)


src/cmds/fs/cat.c
```diff
@@ -21,13 +21,20 @@ int main(int argc, char **argv) {
 	int number = 0, line = 0, new_line = 1;
 	FILE *fd;
 	char buff;
+	int no_of_arg;
+	char *file_names[argc];
+	size_t arg_len;
 
+	for (no_of_arg = 0; no_of_arg < argc; no_of_arg++) {
+		file_names[no_of_arg] = strdup (argv[no_of_arg]);
```
Deryugin: Do we really need to create a copy for those strings? As I can see, they are just passed to `strcmp()` and `fopen()`, both of them have `const` modifier for arguments, so they should not change this memory anyway.
```
int strcmp(const char *s1, const char *s2);
FILE *fopen(const char *pathname, const char *mode);
```

src/cmds/fs/cat.c
```diff
@@ -43,26 +50,36 @@ int main(int argc, char **argv) {
 		}
 	}
 
-	if (NULL == (fd = fopen(argv[argc - 1], "r"))) {
-		printf("Can't open file %s\n", argv[argc - 1]);
-		return -errno;
-	}
+	for (no_of_arg = 1; no_of_arg < argc; no_of_arg++) {
+		if (strcmp ("-n", file_names[no_of_arg]) != 0) {
```
Deryugin: Please follow our [code style](https://github.com/embox/embox/wiki/Code-Style), we don't place space between function name and opening parenthesis.

src/cmds/fs/cat.c
```diff
@@ -26,11 +26,9 @@ int main(int argc, char **argv) {
 	size_t arg_len;
 
 	for (no_of_arg = 0; no_of_arg < argc; no_of_arg++) {
-		arg_len = strlen(argv[no_of_arg]);
-		file_names[no_of_arg] = (char *)malloc (sizeof(char) * arg_len);
-		strcpy (file_names[no_of_arg], argv[no_of_arg]);
+		file_names[no_of_arg] = strdup (argv[no_of_arg]);
 	}
-
```
anton-bondarev: Please, keep code clear

src/cmds/fs/cat.c
```diff
@@ -84,4 +82,4 @@ int main(int argc, char **argv) {
 	}
 
 	return 0;
-}
```
anton-bondarev: It's mark wrong, may be you've removed a new line?

sobhaniiest:
Hi all, I have done the changes. Please have a look.

src/cmds/fs/cat.c
```diff
@@ -43,26 +44,31 @@ int main(int argc, char **argv) {
 		}
 	}
 
-	if (NULL == (fd = fopen(argv[argc - 1], "r"))) {
-		printf("Can't open file %s\n", argv[argc - 1]);
-		return -errno;
-	}
+	for (i = 1; i < argc; i++) {
+		if (strcmp("-n", argv[i]) != 0) {
```
anton-bondarev: Do you use getopt or manually parse arguments?

sobhaniiest: There may be scenarios like 'cat file1 file2 -n' or 'cat file1 -n file2'. To handle these kinds of scenarios I am parsing manually and ignoring '-n' in 'argv'.

Deryugin: It's not widely known, but this is actually an extension of GNU coreutils.
If you refer to POSIX (Utility Syntax Guidelines)[https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html#tag_12_02], then you can see
```
Guideline 9:
    All options should precede operands on the command line.
```
so POSIX-compatible `cat` utility would interpret `cat f1 -n` like two files.

But for me it's fine to deal with this option this way, so let's leave it as is.

sobhaniiest: Oh! Thank you very much for the information.

sobhaniiest:
Thanks for approving.

PR #1772, opened 2020-03-03
fixes #1745 math floating point tests

### LINKED ISSUE #1745: math floating point tests
There are no basic math tests from FP functions in Embox. Let's add something like `src/tests/math/math_test_fp.c` and test for the following functions:

```
extern double modf(double x, double *i_ptr);
extern double fmod(double x, double y);
extern int signbit(double x);
extern double fabs(double x);
extern float fabsf(float x);
extern long double fabsl(long double x);
```

(it's first several functions from src/compat/libc/math/math_builtins.h)

It's assumed the tests should be very simple, just to make sure the corresponding functions are not stubs and really doing something.

rushilkvs:
wrote tests for fp functions and also replaced embox.libc.math_builtins with embox.compat.libc.math_openlibm.

### DIFF
--- a/src/tests/math/math_test_fp.c
+++ b/src/tests/math/math_test_fp.c
@@ -1,0 +1,66 @@
+/**
+ * @math_test_fp
+ * @simple tests for fp functions in embox just to make sure the corresponding functions are not stubs
+ * @date 03.03.2020
+ * @author Rushil kvs
+ */
+
+#include <embox/test.h>
+#include <math.h>
+
+EMBOX_TEST_SUITE("Tests for fp functions");
+
+bool equal(double a, double b)
+{
+	double epsilon = 0.00000000000001;
+	if ((a-b) < epsilon && -(a-b) < epsilon) return true;
+  	return false;
+}
+
+TEST_CASE("Test for modf") {
+	double i,t,j;
+	i = modf(2.34, &t);
+	j = 0.34;
+	test_assert_true(equal(i, j));
+	test_assert_true(equal(t, 2.0));
+}
+
+TEST_CASE("Test for fmod") {
+	double i;
+	i = fmod(11, 2.5);
+	test_assert_true(equal(i, 1.0));
+}
+
+TEST_CASE("Test for signbit") {
+	size_t i;
+	const double values[] = { 0.0, 0.131321312312, 21.3, 2883321.0 };
+
+	for (i = 0; i < sizeof values / sizeof values[0]; ++i) {
+		test_assert_zero(signbit(values[i]));
+		test_assert_not_zero(signbit(-values[i]));
+	}
+}
+
+TEST_CASE("Test for fabs") {
+	double i,j;
+	i = 8.2;
+	j = -0.12;
+	test_assert_true(equal(i, fabs(i)));
+	test_assert_true(equal(-j, fabs(j)));
+}
+
+TEST_CASE("Test for fabsf") {
+	float i,j;
+	i = 4.6;
+	j = -0.56;
+	test_assert_true(equal(i, fabsf(i)));
+	test_assert_true(equal(-j, fabsf(j)));
+}
+
+TEST_CASE("Test for fabsl") {
+	long double i,j;
+	i = 5.6;
+	j = -0.34;
+	test_assert_true(equal(i, fabsl(i)));
+	test_assert_true(equal(-j, fabsl(j)));
+}
--- a/src/tests/math/Mybuild
+++ b/src/tests/math/Mybuild
@@ -1,5 +1,13 @@
 package embox.test.math
 
+module math_test_fp {
+	source "math_test_fp.c"
+
+	depends embox.compat.libc.all
+	depends embox.compat.libc.math
+	depends embox.framework.LibFramework
+}
+
 module math_test {
 	source "math_test.c"
 


src/tests/math/math_test_fp.c
```diff
@@ -0,0 +1,89 @@
...
+ * @author Rushil kvs
+ */
+
+#include <embox/test.h>
+#include <math.h>
+
+EMBOX_TEST_SUITE("Tests for fp functions");
+
+bool Equal(double a, double b)
+{
```
alexkalmuk: Can you please follow our code style? https://github.com/embox/embox/wiki/Code-Style

We are using lower case functions namings.

src/tests/math/math_test_fp.c
```diff
@@ -0,0 +1,89 @@
...
+	float epsilon = 0.00000000001;
+	if ((a-b) < epsilon && -(a-b) < epsilon) return true;
+  	return false;
+}
+
+long double Equall(long double a, long double b){
+	long double epsilon = 0.0000000000000000000001;
+	if ((a-b) < epsilon && -(a-b) < epsilon) return true;
+  	return false;
+}
```
alexkalmuk: As I understand, it is required because the results are not simple floating-point numbers like `1.5`, but `1.545463634634(8)`, right? Can we make such tests to only use simple floating-point numbers like `1.5`, `3.7` etc.? If os, then you can use simple `test_assert_equal` instead of these epsilons.

src/tests/math/math_test_fp.c
```diff
@@ -0,0 +1,89 @@
...
+	test_assert_true(Equal(t,8.0));
+	test_assert_false(Equal(t,-8.0));
+}
+
+TEST_CASE("Test for fmod") {
+	double i;
+	//bool b;
+	i = fmod(8.432,3.112);
+	test_assert_true(Equal(i,2.208));
+	test_assert_false(Equal(i,1.345));
```
alexkalmuk: There is no need to check `assert_false` since we already checked `assert_true`.

src/tests/math/math_test_fp.c
```diff
@@ -0,0 +1,89 @@
...
+	test_assert_true(Equal(i,j));
+	test_assert_false(Equal(i,-j));
+	test_assert_true(Equal(t,8.0));
+	test_assert_false(Equal(t,-8.0));
+}
+
+TEST_CASE("Test for fmod") {
+	double i;
+	//bool b;
+	i = fmod(8.432,3.112);
```
alexkalmuk: Speaking about of simplifications with epsilons I just mean that you can try to change this one

```
i = fmod(8.432,3.112);
test_assert_true(Equal(i,2.208));
```

to this one:

```
i = fmod(5.0,2.0);
test_assert_true(i, 2.5);
```

src/tests/math/math_test_fp.c
rushilkvs: Thanks for reviewing, I'll do the required changes

src/tests/math/math_test_fp.c
rushilkvs: I guess, we can remove Equalf and Equall. But we can't use simple test_assert_equal , as it directly checks a==b.

rushilkvs: So, we need to have atleast one Equal function

rushilkvs:
Done all the suggested changes

Deryugin:
Please provide more meaningful commit messages and PR name.

src/tests/math/math_test_fp.c
```diff
@@ -0,0 +1,66 @@
...
+	if ((a-b) < epsilon && -(a-b) < epsilon) return true;
+  	return false;
+}
+
+TEST_CASE("Test for modf") {
+	double i,t,j;
+	i = modf(2.34, &t);
+	j = 0.34;
+	test_assert_true(equal(i, j));
+	test_assert_true(equal(t, 2.0));
```
alexkalmuk: Wouldn't just the following code work?

```
	i = modf(2.34, &t);
	j = 0.34;
	test_assert_equal(i, j);
	test_assert_equal(t, 2.0);
```

src/tests/math/math_test_fp.c
alexkalmuk: Hmm, let's check this, I commented below for the `modf(2.34, &t)`.

alexkalmuk: Ok, let's leave this one `equal` function.

src/tests/math/math_test_fp.c
alexkalmuk: Ok, let's use your variant.

alexkalmuk:
@rushilkvs Can I ask you to squash your commits into a single one with a commit message like `tests: Add math floating-point tests`.

src/tests/math/math_test_fp.c
alexkalmuk: Here I just meant that we actually do not know the possible error, so my proposition was to choose such numbers to make it just `0.0`. But let's use the `epsilon` everywhere here.

alexkalmuk:
@rushilkvs Thank you for squashing! But what about the commit message?

rushilkvs:
> @rushilkvs Thank you for squashing! But what about the commit message?

Sorry, I forgot I'll change it rightaway

src/tests/math/math_test_fp.c
```diff
@@ -0,0 +1,66 @@
...
+	test_assert_true(equal(-j, fabsf(j)));
+}
+
+TEST_CASE("Test for fabsl") {
+	long double i,j;
+	i = 5.6;
+	j = -0.34;
+	test_assert_true(equal(i, fabsl(i)));
+	test_assert_true(equal(-j, fabsl(j)));
+}
```
anton-bondarev: add a new line please

alexkalmuk:
@rushilkvs Great! Now the last thing and we will merge - can you please squash your commits in the single one with a commit message like `tests: Add math_test_fp`, for example.

PR #1773, opened 2020-03-03
posix: semaphore: Add an adaptive layer to get POSIX semaphore value

### LINKED ISSUE #1750: Incompatible behavior of kernel sempahore with POSIX semaphore
Up to now, in Embox, the POSIX-compliance API of semaphore are implemented in `src/compat/posix/semaphore.c`:

```c
sem_init(sem_t *sem, int pshared, unsigned int value);
sem_destroy(sem_t *sem);
sem_getvalue(sem_t *restrict sem, int *restrict sval);
sem_post(sem_t *sem);
sem_trywait(sem_t *sem);
sem_wait(sem_t *sem);
```

These functions call the semaphore implementations in the Embox kernel (`src/kernel/thread/sync/semaphore.c`). 

The kernel semaphore module works well at the moment:
- `semaphore_leave()` do a release on a semaphore, and decrement `value` field in a `sem_t` structure.
- `semaphore_enter()` do an occupation, and increment `value` field in a `sem_t` structure. 

In this case, when `value` field of a `sem_t` structure is 0, it means the semaphore is totally free. Meanwhile, the case when `value` field equals `max_value` field, means the semaphore is totally busy.

But its behaviors are not compatible with POSIX semaphore.

# POSIX semaphore

According to Linux manual, some of POSIX semaphore API are defined as below:

sem_init() initializes the unnamed semaphore at the address pointed to by sem. The value argument specifies the initial value for the semaphore.

sem_post() increments (unlocks) the semaphore pointed to by sem.

sem_wait() decrements (locks) the semaphore pointed to by sem.

In conclusion, when value of a semaphore is 0, the semaphore is totally occupied. It seems that there is no max value limit in the specification.

# Embox POSIX semaphore implementation

[POSIX-compliance Implementations](https://github.com/Inokinoki/embox/blob/master/src/compat/posix/semaphore.c).
[In-kernel Implementations](https://github.com/Inokinoki/embox/blob/master/src/kernel/thread/sync/semaphore.c).

```c
void semaphore_init(struct sem *s, int val) {
	waitq_init(&s->wq);
	s->value = 0;
	s->max_value = val;
}
```

`sem_init()` calls `semaphore_init()`, which initializes `max_value` field of sem instance to the given initial value, and `value` to 0. *Incompatible*

```c
int sem_post(sem_t *sem) {
	semaphore_leave(sem);

	return ENOERR;
}
```

`sem_post()` calls `semaphore_leave()`, which decrements `value` field of sem_t instance. *Incompatible*

```c
int sem_wait(sem_t *sem) {
	semaphore_enter(sem);

	return ENOERR;
}
```

`sem_wait()` calls `semaphore_enter()`, which increments `value` field of sem_t instance. *Incompatible*

The other functions which call Embox kernel semaphore implementations have the same problem.

Inokinoki:
POSIX semaphore value begins from its max available value, `sem_post()` decrements it and `sem_wait()` decrements it. The kernel semaphore module does in contrary.

Temporally resolve #1750

### DIFF
--- a/src/compat/posix/semaphore.c
+++ b/src/compat/posix/semaphore.c
@@ -22,7 +22,9 @@ int sem_destroy(sem_t *sem) {
 }
 
 int sem_getvalue(sem_t *restrict sem, int *restrict sval) {
-	return semaphore_getvalue(sem, sval);
+	int ret = semaphore_getvalue(sem, sval);
+	*sval = sem->max_value - *sval;
+	return ret;
 }
 
 int sem_post(sem_t *sem) {


alexkalmuk:
It looks good to me, but probably in the future, we should change the embox's internal semaphores due to POSIX.

anton-bondarev:
@alexkalmuk May be create an issue?

alexkalmuk:
> > but probably in the future, we should change the embox's internal semaphores due to POSIX.
> 
> @alexkalmuk May be create an issue?

I think the current issue https://github.com/embox/embox/issues/1750 if perfect for that.

Issue #1644, opened 2020-01-21
bsd/queue support


Deryugin:
Right now we don't have any modules for bsd compatibility, I suggest creating a separate directory for it, like `src/compat/bsd/`.

We need to implement macros/functions from BSD `queue.h`, it can be tested with #1643

AntonKozlov:
Can we just take them as is
https://github.com/freebsd/freebsd/blob/master/sys/sys/queue.h ? It's BSD
too

src/compat/bsd is nice place, but on my Linux system it's just in
/usr/include
```
$ head /usr/include/sys/queue.h
/*
 * Copyright (c) 1991, 1993
 *      The Regents of the University of California.  All rights reserved.
 *
```

Deryugin:
Yeah, probably we can just take this file.

As for the header directory, we're trying to avoid putting header files into `src/include` by using `@IncludeExport`, otherwise `src/include` directory becomes a bit messy.

subhojit-ghorai:
Hi! Can I work on this issue? I have been doing research on this issue.

Deryugin:
Hi, 
Sure!

PR #1779, opened 2020-03-04
 implemented macros/functions from BSD queue.h. (#1644)

subhojit-ghorai:
queue.h has been added. To make queue.h compatible necessary variables and functions have been added to cdefs.h.


Deryugin:
Please implement a few tests to make sure that this macros can be compiled and are working correctly.

Deryugin:
#1784

PR #1780, opened 2020-03-05
Scsi simplify the logic

alexkalmuk:
The goal of this PR is just to simplify SCSI logic.

* Remove scsi per-command states. While the idea to have scsi state machine is undoubtedly good, introducing a state for each scsi command looks overcomplicated and creates additional troubles to debug and add new states. I would like to refer here to Linux `include/scsi/scsi_device.h`:

```
enum scsi_device_state {
    SDEV_CREATED = 1,   /* device created but not added to sysfs
                 * Only internal commands allowed (for inq) */
    SDEV_RUNNING,       /* device properly configured
                 * All commands allowed */
    SDEV_CANCEL,        /* beginning to delete device
***
};
```

You can see here a state machine, but not for each scsi command. In Embox there are two states only - attached and detached for now (I mean the flag `dev->attached`).

* Make mass storage to handle errors during the probe(), so here we check if scsi failed.

* (possible downgrade, minor) I removed scsi_disk_recover, since it already seems to be broken now in master, and I even have no idea how we can test it. Alternatively, I'd suggest deferring disk recovering to the future, when we would understand how to deal with it.

I verified the changes on `x86/qemu` with both OHCI and EHCI with `scripts/qemu/auto_qemu_with_usb_storage`, and with `arm/iwave_imx6` making `dd` on the flash drive inserted.

### DIFF
--- a/src/drivers/scsi/scsi.c
+++ b/src/drivers/scsi/scsi.c
@@ -12,32 +12,26 @@
 
 #include <drivers/usb/usb.h>
 #include <drivers/usb/class/usb_mass_storage.h>
-
 #include <drivers/scsi.h>
 
-#include <kernel/panic.h>
-
 #include <util/log.h>
+
+#define SCSI_CMD_MAX_LEN 16
 
 static inline struct usb_mass *scsi2mass(struct scsi_dev *dev) {
 	return member_cast_out(dev, struct usb_mass, scsi_dev);
 }
 
-static const struct scsi_dev_state scsi_state_test_unit;
-static const struct scsi_dev_state scsi_state_inquiry;
-static const struct scsi_dev_state scsi_state_capacity10;
-static const struct scsi_dev_state scsi_state_sense;
-
-int scsi_cmd(struct scsi_dev *sdev, void *cmd, size_t cmd_len, void *data, size_t data_len) {
+static int scsi_cmd(struct scsi_dev *sdev, void *cmd, size_t cmd_len,
+		void *data, size_t data_len) {
 	struct usb_mass *mass = scsi2mass(sdev);
 	struct scsi_cmd *scmd = cmd;
-	enum usb_direction usb_dir = USB_DIRECTION_IN;
-	int ret;
+	enum usb_direction usb_dir;
 
 	if (!sdev->attached) {
 		return -ENODEV;
 	}
-	if(scmd->scmd_opcode == SCSI_CMD_OPCODE_WRITE10) {
+	if (scmd->scmd_opcode == SCSI_CMD_OPCODE_WRITE10) {
 		usb_dir = USB_DIRECTION_OUT;
 	} else {
 		usb_dir = USB_DIRECTION_IN;
@@ -46,19 +40,15 @@ int scsi_cmd(struct scsi_dev *sdev, void *cmd, size_t cmd_len, void *data, size_
 	log_debug("opc (0x%x) cmd_len(%d), usb_dir(%d), data_len(%d)",
 		    scmd->scmd_opcode, cmd_len, usb_dir, data_len);
 	
-	ret = usb_ms_transfer(mass->usb_dev, cmd, cmd_len, usb_dir, data, data_len);
-	scsi_request_done(sdev, ret);
-
-	return ret;
+	return usb_ms_transfer(mass->usb_dev, cmd, cmd_len, usb_dir, data, data_len);
 }
 
-#define SCSI_CMD_LEN 16
 int scsi_do_cmd(struct scsi_dev *dev, struct scsi_cmd *cmd) {
-	uint8_t scmd[SCSI_CMD_LEN];
+	uint8_t scmd[SCSI_CMD_MAX_LEN];
 
 	log_debug("opcode 0x%x, len %d", cmd->scmd_opcode, cmd->scmd_len);
 
-	assert(cmd->scmd_len <= SCSI_CMD_LEN);
+	assert(cmd->scmd_len <= SCSI_CMD_MAX_LEN);
 	memset(scmd + 1, 0, cmd->scmd_len - 1);
 	scmd[0] = cmd->scmd_opcode;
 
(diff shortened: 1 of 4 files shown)


src/drivers/scsi/scsi.c
```diff
@@ -69,6 +59,105 @@ int scsi_do_cmd(struct scsi_dev *dev, struct scsi_cmd *cmd) {
...
+	cmd.scmd_olen = 0;
+	return scsi_do_cmd(dev, &cmd);
+}
+
+static int scsi_send_sense(struct scsi_dev *dev) {
+	struct scsi_cmd cmd = scsi_cmd_template_sense;
+	struct scsi_data_sense *data;
+	uint8_t acode;
+
+	log_debug("");
```
anton-bondarev: Rm or print more information?

src/drivers/scsi/scsi.c
```diff
@@ -69,6 +59,105 @@ int scsi_do_cmd(struct scsi_dev *dev, struct scsi_cmd *cmd) {
...
+			!= SCSI_INQIRY_DEVTYPE_BLK) {
+		return -1;
+	}
+	return 0;
+}
+
+static int scsi_send_test_unit(struct scsi_dev *dev) {
+	struct scsi_cmd cmd = scsi_cmd_template_test_unit;
+
+	log_debug("");
```
anton-bondarev: Rm or print more information?

src/drivers/scsi/scsi.c
```diff
@@ -69,6 +59,105 @@ int scsi_do_cmd(struct scsi_dev *dev, struct scsi_cmd *cmd) {
...
+
+	cmd.scmd_obuf = dev->scsi_data_scratchpad;
+	cmd.scmd_olen = sizeof(struct scsi_data_sense);
+	if (scsi_do_cmd(dev, &cmd) < 0) {
+		return -1;
+	}
+
+	data = (struct scsi_data_sense *) dev->scsi_data_scratchpad;
+	acode = data->dsns_additional_code;
+	if (!(acode == 0x28 || acode == 0x29)) {
```
anton-bondarev: May be using MACRO for code definition?

src/drivers/scsi/scsi.c
```diff
@@ -69,6 +59,105 @@ int scsi_do_cmd(struct scsi_dev *dev, struct scsi_cmd *cmd) {
...
+		return -1;
+	}
+	return 0;
+}
+
+static int scsi_send_capacity10(struct scsi_dev *dev) {
+	struct scsi_cmd cmd = scsi_cmd_template_cap10;
+	struct scsi_data_cap10 *data;
+
+	log_debug("");
```
anton-bondarev: Rm or print more information?

src/drivers/scsi/scsi.c
```diff
@@ -69,6 +59,105 @@ int scsi_do_cmd(struct scsi_dev *dev, struct scsi_cmd *cmd) {
...
+int scsi_dev_attached(struct scsi_dev *dev) {
+	int ret;
+	int test_unit_retries = 3;
+
+	if (scsi_send_inquiry(dev)) {
+		log_debug("INQUIRY failed");
+		return -1;
+	}
+
+	/* FIXME This logic derived from the previos variant,
```
anton-bondarev: We send sence command if inquiry failed, flash drive can be not ready yet

alexkalmuk: Are sure about `if inquiry failed`?

src/drivers/scsi/scsi_disk_create.c
```diff
@@ -52,6 +31,8 @@ static void *scsi_create_thread(void *arg) {
 	assert(arg);
 	sdev = arg;
 
+	mutex_init(&sdev->m);
```
anton-bondarev: do we use a thread now?

src/drivers/scsi/scsi.c
alexkalmuk: I am actually not sure, because it is not so in master. We send inquiry if test unit failed

alexkalmuk: I will try to investigate in more detail how it should be...

src/drivers/scsi/scsi_disk_create.c
alexkalmuk: Nope, will try to remove

alexkalmuk: Hmm, it looks like these mutexes are used for the case when multiple threads can use SCSI device. So currently it is not a valid situation, of course, but the code with mutexes lock/unlock  looks like to be valid, so I am not whether to remove it...

src/drivers/scsi/scsi.c
alexkalmuk: Uhh, it's difficult to say what is a correct check for this. I was looking into SCSI specification and found that for example value `0x29` (look at the first column) also requires to check scsi code qualifier (the second column), so I cannot even define such a MACRO, because it may be a combination of two codes, who knows...

```
29 00 Power On, Reset, Or Bus Device Reset Occurred 6
29 01 Power-On Reset Occurred 6
29 02 SCSI Bus Reset Occurred 6
29 03 Bus Device Reset Function Occurred 6
29 04 Internal Reset Occurred 6
29 05 Transceiver Mode Changed To Single-Ended 6
29 06 Transceiver Mode Changed To LVD 6
29 07 Write Log Dump data to disk successful OR IT Nexus Loss 6
29 08 Write Log Dump data to disk fail 6
29 09 Write Log Dump Entry information fail 6
29 0A Reserved disk space is full 6
29 0B SDBP test service contained an error, examine status packet(s) for detail
```

src/drivers/scsi/scsi.c
alexkalmuk: Thanks, done.

src/drivers/scsi/scsi.c
alexkalmuk: Done.

src/drivers/scsi/scsi.c
alexkalmuk: Done

PR #1783, opened 2020-03-05
implemented macros/functions from BSD queue.h. (Fixed #1644)

### LINKED ISSUE #1644: bsd/queue support

subhojit-ghorai:
queue.h has been added. To make queue.h compatible necessary variables and functions have been added to cdefs.h.


Deryugin:
#1784

PR #1797, opened 2020-03-07
driver: clock: Fix bugs and add comments for pl031 timer

layman-n-ish:
driver: clock: Fix bugs and add comments for pl031 timer

Signed-off-by: Nishant Malpani <nish.malpani25@gmail.com>


layman-n-ish:
Is it a good idea to design a macro that concatenates the module's name symbol (in this case, pl031) when we declare the device's attributes like registers, base address, irq number, etc.?

alexkalmuk:
> Is it a good idea to design a macro that concatenates the module's name symbol (in this case, pl031) when we declare the device's attributes like registers, base address, irq number, etc.?

I haven't seen this anywhere. There is a kind `GET/SET` macros in other projects to help working with registers' fields.

Moreover how you are going to use proposed macros in macro definitions? If you are going to do

```
#define DRV(x)     PL031_ ## x
#define DRV(DR)    (DRV(BASE) + 0x00)
```

it wouldn't be compiled.

You have only to do something like

```
#define PL031_BASE  0x12345678
#define PL031_DR    (DRV(BASE) + 0x00)
```

but I do not see any improvement since you still have to use PL031.

alexkalmuk:
You can see a grey unauthorized icon for example here - https://github.com/embox/embox/pull/1797/commits/1a4c7b08b6ecd77c08d3347320ac99822a00d69c. Probably because of https://stackoverflow.com/questions/26004587/git-commits-are-not-getting-linked-with-my-github-account Please fix this next time :)

Issue #1709, opened 2020-02-24
Extend block_dev_test command with new options


alexkalmuk:
`block_dev_test` command is a simple testing tool for block devices like hdd, usb, flash, etc. It checks that read/write routines work correct.

We suggest extending the existing functionality with the following options:

* Add parameter which specifies the block number at which the command should start at.
* Add parameter which specifies blocks count to be tested.
* Add option to read/write multiple blocks in a single iteration (currenly, we read/write only single block in one iteration)

ymdatta:
Hi, i would like to work on this issue. Can i take it up?
Thanks.

alexkalmuk:
@ymdatta Yes, you can. But first of all : I cannot find your proposal in email. Did you send your proposal to anybody of the mentors?

alexkalmuk:
We received your email, thanks,  so now I will assign you to this issue.

alexkalmuk:
Hi @ymdatta, thank you for the https://github.com/embox/embox/pull/1727.

You probably faced these problems with `-i` and `cd0`, so I think it's better to test block_dev_test new options with another stuff:

```
./scripts/qemu/qemu_with_disk.sh

```
In Embox:
```
root@embox:/#block_dev_test hda                                                 
Starting block device test (iters = 1)...
iter 0...
Testing 0 (of 16384)
Testing 1 (of 16384)
Testing 2 (of 16384)
Testing 3 (of 16384)
...
```

ymdatta:
Hi @alexkalmuk, thanks for the suggestion.

When i started trying to solve this issue (#1709), i realized i need something other than `cd0` to test my changes with. So, i followed this [Using-USB-on-qemu](https://github.com/embox/embox/wiki/Using-USB-on-qemu) and right now i am using USB to test my changes.

alexkalmuk:
Nice, are you using `ehci_pci` or `ohci_pci` in your mods.config?

ymdatta:
I tried using `ochi_pci` it didn't work, but with `ehci_pci` it's working. :)

alexkalmuk:
> I tried using ochi_pci it didn't work

Yeah, it's already known bug :) I'm just working on it right now.

alexkalmuk:
@ymdatta I merged your contribution, so before to start with the third bullet please rebase on the top of our master.

ymdatta:
@alexkalmuk I will be keep that in mind before committing. Thanks.

ymdatta:
[...]

PR #1798, opened 2020-03-07
cmd: testing: block_dev_test: add multiple read/writes blocks option

ymdatta:
This patch adds functionality for 'block_dev_test' command to
specify multiple read/write blocks in a single iteration. It also
adds information about this option to be displayed in the help
message. Default value of multiple blocks is set as 4, in case if
the argument is not given by the user but the flag is provided.
This fixes part of #1709.

Signed-off-by: Mohana Datta Yelugoti <ymdatta.work@gmail.com>

### DIFF
--- a/src/cmds/testing/block_dev/block_dev_test.c
+++ b/src/cmds/testing/block_dev/block_dev_test.c
@@ -65,7 +65,7 @@ static struct flash_dev *get_flash_dev(struct block_dev *bdev) {
 	return NULL;
 }
 
-static int flash_dev_test(struct flash_dev *fdev, uint32_t s_block, uint32_t n_blocks) {
+static int flash_dev_test(struct flash_dev *fdev, uint32_t s_block, uint32_t n_blocks, uint32_t m_blocks) {
 #define FLASH_RW_LEN 256
 	int offset;
 	uint8_t rbuf[FLASH_RW_LEN];
@@ -89,24 +89,27 @@ static int flash_dev_test(struct flash_dev *fdev, uint32_t s_block, uint32_t n_b
 	}
 
 	/* Iterate over all flash blocks */
-	for (uint32_t i = s_block; i < blocks; i++) {
-		flash_erase(fdev, i);
+	for (uint32_t i = s_block; i < blocks; i += m_blocks) {
 
-		offset = i * fdev->block_info[0].block_size;
+		for(uint32_t k = i; k < min(blocks, i + m_blocks); k++) {
+			flash_erase(fdev, k);
 
-		/* Write, then read back and check result */
-		for (size_t j = 0; j < fdev->block_info[0].block_size; j += FLASH_RW_LEN) {
-			/* Clean rbuf first */
-			memset(rbuf, 0x0, FLASH_RW_LEN);
-			flash_write(fdev, offset, wbuf, FLASH_RW_LEN);
-			flash_read(fdev, offset, rbuf, FLASH_RW_LEN);
-			if (0 != memcmp(wbuf, rbuf, FLASH_RW_LEN)) {
-				printf("Flash device test failed:\n");
-				printf("  fdev=%s, block=%"PRIu32", offset within block=%zu\n",
-					fdev->bdev->name, i, j);
-				return -1;
+			offset = k * fdev->block_info[0].block_size;
+
+			/* Write, then read back and check result */
+			for (size_t j = 0; j < fdev->block_info[0].block_size; j += FLASH_RW_LEN) {
+				/* Clean rbuf first */
+				memset(rbuf, 0x0, FLASH_RW_LEN);
+				flash_write(fdev, offset, wbuf, FLASH_RW_LEN);
+				flash_read(fdev, offset, rbuf, FLASH_RW_LEN);
+				if (0 != memcmp(wbuf, rbuf, FLASH_RW_LEN)) {
+					printf("Flash device test failed:\n");
+					printf("  fdev=%s, block=%"PRIu32", offset within block=%zu\n",
+					       fdev->bdev->name, k, j);
+					return -1;
+				}
+				offset += FLASH_RW_LEN;
 			}
-			offset += FLASH_RW_LEN;
 		}
 	}
 	return 0;
@@ -152,7 +155,7 @@ static void dump_buf(int8_t *buf, size_t cnt, char *fmt) {
 	printf("============================\n");
 }
 
-static int block_dev_test(struct block_dev *bdev, uint64_t s_block, uint64_t n_blocks, int m_blocks) {
+static int block_dev_test(struct block_dev *bdev, uint64_t s_block, uint64_t n_blocks, uint64_t m_blocks) {
 	size_t blk_sz;
 	int8_t *read_buf, *write_buf;
 	uint64_t blocks, total_blocks;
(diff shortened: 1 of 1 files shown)


ymdatta:
Testing with new changes:
btest_mblocks

src/cmds/testing/block_dev/block_dev_test.c
```diff
@@ -191,30 +197,57 @@ static int block_dev_test(struct block_dev *bdev, uint64_t s_block, uint64_t n_b
...
+
+		size_t data_sz = blk_sz * m_blocks;
+
+		/**
+		 * When writing multiple blocks, if the data we are supposed to
+		 * write, exceeds the device size, then we write till the device
+		 * size is full and ignore the rest.
+		 */
+		if(data_sz + (i * blk_sz) > bdev->size) {
+			uint64_t diff_size = (data_sz + (i * blk_sz)) - bdev->size;
```
alexkalmuk: I am not sure about how the data could exceed the device size, because you already verified above that m_blocks suited the device size.

src/cmds/testing/block_dev/block_dev_test.c
```diff
@@ -499,13 +532,26 @@ int main(int argc, char **argv) {
...
 		}
 	}
 
+	if (m_blocks_flag) {
+		if (m_blocks < 0) {
+			printf("Multiple blocks count argument value should be greater than zero\n");
+			return -EINVAL;
+		} else if (!m_blocks) {
+			m_blocks = DEFAULT_MULTIPLE_BLOCKS_NUMBER;
+		}
```
alexkalmuk: Maybe interpret `-m 0` like an error too? So we do not need DEFAULT_MULTIPLE_BLOCKS_NUMBER then and can just use `m_blocks = 1` if `-m` wasn't specified.

alexkalmuk:
> Testing with new changes:
> btest_mblocks

It looks just fine :) I will verify this a bit later too.

src/cmds/testing/block_dev/block_dev_test.c
alexkalmuk: Or is it possible to provide the `-m` flag without an argument at all?

ymdatta: If we provide -m flag without an argument, and if we try to read the argument we get it as zero. So, i added that case. As Deryugin said 

> have some default value so we don't require user to specify it every time.

The idea is this: In General, the command runs with single block read/write per iteration. If the user calls the command with '-m' flag and without an argument (or) zero argument, it runs with the default block read/writes per iteration. If the user specifies an argument, then the command runs with that many read/writes per iteration.

src/cmds/testing/block_dev/block_dev_test.c
ymdatta: Yeah, you are correct, we don't need that. I will update the commit.

ymdatta: @alexkalmuk: I updated the commit with the changes. :)

alexkalmuk: Thanks

src/cmds/testing/block_dev/block_dev_test.c
alexkalmuk: I want to point here that calling `-m` without an argument shouldn't be allowed at all. When you say that user can run just `-m` you should guarantee it would be handled correctly anyway. Let's see a simple program:

```
#include <stdio.h>
#include <unistd.h>
#include <stdlib.h>

int main(int argc, char *argv[]) {
    int opt;
    int targ, tfnd;

    while ((opt = getopt(argc, argv, "t:")) != -1) {
        switch (opt) {
        case 't':
            targ = atoi(optarg);
            tfnd = 1;
            break;
        default: /* '?' */
            fprintf(stderr, "Usage: %s [-t arg] name\n",
                argv[0]);
            exit(EXIT_FAILURE);
        }
    }
    printf("tfnd=%d; targ=%d; name=%s\n",
        tfnd, targ, argv[optind]);

    exit(EXIT_SUCCESS);
}
```

```
$ gcc - o test test.c

$ ./test -t 7 hda
tfnd=1; targ=7; name=hda

$ ./test -t
./test: option requires an argument -- 't'
Usage: ./test [-t arg] name

$ ./test -t 1231.hda
tfnd=1; targ=1231; name=(null)
```

So you can see that value of `targ` actually depends on what use passed. And if you allow `-t` without arguments, then in the last case `./test -t 1231.hda` the input is pretty correct but you see what happened..

alexkalmuk: One more example when input is correct, but the name becomes null.

```
$ ./test -t hda
tfnd=1; targ=0; name=(null)
```

alexkalmuk: I was wondering why block_dev_test works in your branch when I run it as following:

```
root@embox:/#block_dev_test -m sda
```

But it's actually because of this:

```
bdev = block_dev_find(argv[argc - 1]);
```

so in this case `-m` parse `sda` and then we select sda also like a block device name. It doesn't seem to be an expected behavior.

ymdatta: This is a very good find @alexkalmuk. I learnt something new, thanks. I will update with a new commit.

ymdatta: > I was wondering why block_dev_test works in your branch when I run it as following:
> 
> ```
> root@embox:/#block_dev_test -m sda
> ```
> 
> But it's actually because of this:
> 
> ```
> bdev = block_dev_find(argv[argc - 1]);
> ```
> 
> so in this case `-m` parses `sda` and then we select sda also like a block device name. It doesn't seem to be an expected behavior.

Yes, i will solve this in another PR (i am keeping this PR to solve issue #1709). Is that ok?

ymdatta: > I was wondering why block_dev_test works in your branch when I run it as following:
> 
> ```
> root@embox:/#block_dev_test -m sda
> ```
> 
> But it's actually because of this:
> 
> ```
> bdev = block_dev_find(argv[argc - 1]);
> ```
> 
> so in this case `-m` parses `sda` and then we select sda also like a block device name. It doesn't seem to be an expected behavior.

Yes, you are right. This is not an expected behavior, we can solve this by using second argument of `strtol` to check whether the input was valid. But, i would like to take up on that in another PR, as this PR is specifically for solving #1709. Is that ok?

ymdatta:
Testing with new changes:
arg2_test

src/cmds/testing/block_dev/block_dev_test.c
```diff
@@ -470,18 +489,31 @@ static int part_test(struct block_dev *bdev) {
 	return err;
 }
 
+int is_valid_argument(char *endptr, char *str) {
```
alexkalmuk: `static int` is better because the function is local to this file.

alexkalmuk: Let's fix this small thing an I think your contribution can be merged.

ymdatta: Refactored the function a bit and updated the commit with suggested changes. :)

ymdatta:
Thanks for the help. I think we can close #1709 after this is merged.

PR #1801, opened 2020-03-08
Use GitHub Actions for CI

AntonKozlov:
New Actions infrastructure can be used for CI
* flexible: multiple workflows and triggers, can provide
  * long tests are started nightly
  * shorter tests are started on PR/push
* 20 concurrent workers are provided by default

This PR adds Actions and makes a few changes to improve for CI cycle time. 
There are some thoughts on nightly/on-push a the bottom.

Travis before: 28:46 (https://travis-ci.org/embox/embox/builds/659590134)
Travis after: 25:30 (https://travis-ci.org/AntonKozlov/embox/builds/659745090)
Actions: 14:45 (https://github.com/AntonKozlov/embox/actions/runs/51646799)
(There is some time distribution, numbers are illustrative)

CI run organized in buildgen and actual test jobs. Buildgen only creates `mk/.cache` to avoid each test to do all Mybuilds parsing.
* this should improve test time if concurrent workers number is smaller than number of tests (now it's not: 18 tests vs 20 workers)
* but test jobs are are blocked by buildgen. If you want to compute critical path time, it's buildgen (1:07) + x86/test/qt-vnc (13:10)

Although tests can be splitted between nightly and onpush sets, this PR doesn't do this, all of them are started on each push. One reason is that I'm not sure which ones to move to nightly, the other is that I started to think nightly tests are redundant.
I think, there are no benefits of nightly/on-push compared to all-on-push policy. With the first one, when push tests states I can merge, I do and only after some time I'm notified nightly tests failed or passed. Same effect as with all-on-push, if I stop looking on results when short tests have passed. 
Nightly/on-push and all-on-push policies are equivalent, but with the latter one I can decide how many tests performed by the time of clicking merge button. You may merge in hurry with no tests finished and you can wait a bit (or got distracted) and get many/all of them finished.

### DIFF
--- a/scripts/continuous/net/run.sh
+++ b/scripts/continuous/net/run.sh
@@ -93,10 +93,6 @@ test_end() {
 	fi
 }
 
-determ_dns() {
-	cat /etc/resolv.conf | sed -n 's/nameserver[\ \t]\+//p' | head -n 1 | sed 's/\(127\..\..\..\|localhost\)/$HOST_DNS_IP/'
-}
-
 tap_up() {
 	sudo /sbin/ip tuntap add mode tap tap0
 	sudo /sbin/ifconfig tap0 10.0.2.10 dstaddr 10.0.2.0 netmask 255.255.255.0 down
@@ -120,7 +116,7 @@ tap_down() {
 	sudo /sbin/ip tuntap del mode tap tap0
 }
 
-sed -i "s/CONTINIOUS_RUN_DNS_SERVER/$(determ_dns)/" conf/mods.config
+# FIXME The block can be removed, See discussion to #1703
 cp index.html conf/rootfs/index.html
 make >/dev/null 2>/dev/null
 
--- a/scripts/continuous/net/run.sh
+++ b/scripts/continuous/net/run.sh
@@ -13,7 +13,8 @@ EXPECT_TESTS_BASE=$ROOT_DIR/scripts/expect
 
 EMBOX_IP=10.0.2.16
 HOST_IP=10.0.2.10
-HOST_DNS_IP=10.0.0.11
+
+export PEER_HOST_IP=192.168.128.128 # also hardcoded into x86/test/net start_script
 
 TEST_PING_FORWARING_SCRIPT=$CONT_BASE/net/forwarding/test_ping_forwarding.sh
 
@@ -109,9 +110,22 @@ tap_up() {
 	fi
 
 	sudo service isc-dhcp-server start
+
+	PTTAP=tap78
+	sudo ip tuntap add dev $PTTAP mode tap
+	sudo ip link set   dev $PTTAP address aa:bb:cc:dd:ef:01
+	sudo ip link set   dev $PTTAP up
+	sudo ip addr flush dev $PTTAP
+	sudo ip addr add   dev $PTTAP 192.168.128.1/24
+	sudo ip addr add   dev $PTTAP fe80::192:168:128:1/64
+
+	export EMBOX_USERMODE_TAP_NAME=$PTTAP
+	./ping-target &
 }
 
 tap_down() {
+	pkill ping-target
+	sudo /sbin/ip tuntap del mode tap $PTTAP
 	sudo service isc-dhcp-server stop
 	sudo /sbin/ip tuntap del mode tap tap0
 }
--- a/scripts/continuous/net/run.sh
+++ b/scripts/continuous/net/run.sh
@@ -130,6 +130,9 @@ tap_down() {
 	sudo /sbin/ip tuntap del mode tap tap0
 }
 
+sudo /etc/init.d/ntp restart
+sudo inetd
+
 # FIXME The block can be removed, See discussion to #1703
 cp index.html conf/rootfs/index.html
 make >/dev/null 2>/dev/null
--- a/scripts/continuous/prepare.sh
+++ b/scripts/continuous/prepare.sh
@@ -4,7 +4,7 @@ target=${1//\//__}
 
 x86__test__net() {
 	make confload-x86/test/ping-target
-	make
+	make -j $(nproc)
 	cp build/base/bin/embox ./ping-target
 }
 
--- a/scripts/continuous/prepare.sh
+++ b/scripts/continuous/prepare.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+target=${1//\//__}
+
+x86__test__net() {
+	make confload-x86/test/ping-target
+	make
+	cp build/base/bin/embox ./ping-target
+}
+
+if ! [ $target ] || ! type -t $target > /dev/null; then
+	echo nothing to prepare for \"$1\"
+	exit 0
+fi
+$target "$@"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
     steps:
       - uses: actions/checkout@v1
       - run: make confload-test/buildgen
-      - run: make buildgen
+      - run: make -j $(nproc) buildgen
       - name: compress
         run: |
           mkdir mk-cache
@@ -60,5 +60,5 @@ jobs:
         if: matrix.template == 'x86/test/fs'
       - run: ./scripts/continuous/prepare.sh ${{matrix.template}}
       - run: make confload-${{matrix.template}}
-      - run: make
+      - run: make -j $(nproc)
       - run: ./scripts/continuous/run.sh ${{matrix.template}}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,64 @@
+name: CI
+
+on: [push]
+
+jobs:
+  # does Mybuilds processing and provides mk/.cache to other jobs
+  buildgen:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - run: make confload-test/buildgen
+      - run: make buildgen
+      - name: compress
+        run: |
+          mkdir mk-cache
+          tar zcf mk-cache/archive.tar.gz mk/.cache
+      - uses: actions/upload-artifact@v1
+        with:
+          name: mk-cache
+          path: mk-cache
+
+  ci:
+    needs: buildgen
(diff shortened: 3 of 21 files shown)


scripts/continuous/prepare.sh
```diff
@@ -0,0 +1,15 @@
...
+	make confload-x86/test/ping-target
+	make -j $(nproc)
+	cp build/base/bin/embox ./ping-target
+}
+
+if ! [ $target ] || ! type -t $target > /dev/null; then
+	echo nothing to prepare for \"$1\"
+	exit 0
+fi
+$target "$@"
```
alexkalmuk: What is the point to introduce such a kind of preparation? Could it be executed just in place before ping test in continuous/net/run.sh?

There are already two stages - `build` + `run`. Now you introduce one additional stage and there are `prepare` + `build` + `run`. What is a strict division into `prepare` and others?  At the same time, x86/tests/fs modify start_script and rebuilds embox to test various FS types. So there is no even strict division into `build` and `run` since build `actually` happens during `run`. To keep things simple I'd even propose to combine `build` and `run` stages into a single one - `test`, which can modify modules options, start_script, build whatever required and so on, or just build and run qemu.

Another way is to prepare all the required things before `run`, while `run` builds nothings, justs perform testing. But in that case two stages are enough - `prepare` + `run`.

templates/x86/test/buildgen/build.conf
```diff
@@ -0,0 +1,5 @@
+TARGET = embox
+ARCH = x86
```
alexkalmuk: I like your idea to preserve mk/.cache for CI.

Could it be like `ARCH = foo`, `ARCH = buildgen` or just empty? I'd propose also to move this config to `templates/buildgen` because it does not depend on x86.

scripts/continuous/prepare.sh
AntonKozlov: Tests for `x86/test/net` change template as well. 
Without `prepare`, we would build `x86/test/net`, then `ping-target`, then `x86/test/net'`s (the template with changes). This would give us 3 clean builds, as each goes after `confload`
With `prepare`, we do `ping-target`, `x86/test/net`, `x86/test/net'`. It is 2 clean builds.

`prepare` are steps done before actual build

Build in `build` and build in `run` are different by nature. Build in `build` may fail, it will be a source code issue. Build in `test` fails less likely due to minor changes in config and no changes in code; if the build fails it's tests issue.

prepare + build + run is the semantically the one test procedure. 
Having multiple steps:
1. Allows quick overlook for failed step, you don't need to search for error in a big combined log (tests continue even after failed one) 
2. Each step have separate log, at least Actions interface is more responsive when step is unfolded and a small log appears.

Could you please elaborate simplicity of combined steps. What are its practical advantages?

templates/x86/test/buildgen/build.conf
AntonKozlov: It's there because only `x86` have `test` subfolder :)
I would keep global template namespace as simple as possible. At least it should be `test/buildgen'

alexkalmuk: I am okay with test/buildgen then since `x86/test/buildgen` which then used to build let's say mips/qemu look a bit weird :)

alexkalmuk: Or you do not agree?

alexkalmuk: > I would keep global template namespace as simple as possible

Or you can just keep this `buildgen` to be local for scripts/continuous, and then just copy it to conf/ right before `make buildgen`.

scripts/continuous/net/run.sh
```diff
@@ -113,14 +110,30 @@ tap_up() {
...
 	sudo service isc-dhcp-server stop
 	sudo /sbin/ip tuntap del mode tap tap0
 }
 
-sed -i "s/CONTINIOUS_RUN_DNS_SERVER/$(determ_dns)/" conf/mods.config
+sudo /etc/init.d/ntp restart
+sudo inetd
+
+# FIXME The block can be removed, See discussion to #1703
 cp index.html conf/rootfs/index.html
```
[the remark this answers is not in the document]
anton-bondarev: Answer in #1703

> if there are some problems lets copy from web server (root) directory to test directory (gold file)
> Does It solve the problem?

scripts/continuous/prepare.sh
alexkalmuk: My point of view here is quite simple - it's better do not divide the things into multiple stages while we are not sure which stages are actually required.

A simple example, you add `prepare` stage. Should we move also `nfsd start` and friends there? Because it's also a kind of preparation that doesn't depend on embox build.

So if we claim that `prepare` means a sort of preparation that doesn't involve "target under test" (embox) at all - building some stuff under linux, starting services, etc., it would probably make sense. And only after we will decide so we can introduce a new stage. Probably you already has this in mind, that's why I started this discussion.

alexkalmuk: By the way, if we would claim to start linux services in `prepare` then you should ping-target here too, not in `run` stage.

scripts/continuous/net/run.sh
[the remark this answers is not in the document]
AntonKozlov: Let's integrate this PR first, otherwise I'll get conflicts. I suppose that issue have not been started

scripts/continuous/prepare.sh
AntonKozlov: > doesn't depend on embox build
is not the same
> prepare are steps done before actual build

I should stress that more: steps that are _required_ to run before the build. To build the ping-target before is required (to save one clean build), while run it is not. 

Ping-target cannot be started before the test. Actions are hang when leftover background job is running. I tried :)

> My point of view here is quite simple - it's better do not divide the things into multiple stages while we are not sure which stages are actually required.
> And only after that, we can decide whether to introduce a new stage. Probably you already have this in mind, that's why I started this discussion.

I'm confused by trying to save number of steps, that is, shell script invocations. Why is it important?
I had nothing in mind for optimizing number of steps. I care only about time required for tests to finish. This was a major point of pain, not number of steps
There are strong arguments for `prepare`: it saves one clean build (~1-2 min) and makes using Actions interface more pleasant.
Could you highlight, why one combined step is better that splitted?

alexkalmuk: > This was a major point of pain, not number of steps
There are strong arguments for prepare: it saves one clean build (~1-2 min) and makes using Actions interface more pleasant.

How this prepare.sh saves time? If we would move it at `run` stage testing time will be longer?

AntonKozlov: > Tests for `x86/test/net` change template as well.
> Without `prepare`, we would build `x86/test/net`, then `ping-target`, then `x86/test/net'`s (the template with changes). This would give us 3 clean builds, as each goes after `confload`
> With `prepare`, we do `ping-target`, `x86/test/net`, `x86/test/net'`. It is 2 clean builds.

alexkalmuk: Ugh, now I understand what you are talking about It's a sort of optimization because `ping-target` and `x86/test/net` use different configs, so it is required to go through all these mybuild stages, right?

alexkalmuk: An you optimized this in such a way that`modified x86/test/net' follow by `x86/test/net`?

AntonKozlov: Yes, exactly

alexkalmuk: But now I am wondering why you do not use the prepared `mk/.cache` overall in scripts/continues when some embox build is required instead of introducing new global `prepare` stage and keeping the builds order in mind?

alexkalmuk: Ok, it is probably because of Travis didn't support this cache,

alexkalmuk: Well, you keep all re-builds of the same template to follow each other, ok then.

alexkalmuk:
@AntonKozlov What about to disable Travis then?

alexkalmuk:
@anton-bondarev @Deryugin What do you think about this PR? For me it looks good.

AntonKozlov:
> What about to disable Travis then?

I suggest you to do run on both and search for issues with Actions. If everything will be alright, then it will be a simple task to disable Travis

scripts/continuous/prepare.sh
AntonKozlov: > But now I am wondering why you do not use the prepared `mk/.cache` overall in scripts/continues when some embox build/rebuild is required instead of introducing new global `prepare` stage and keeping the builds order in mind?
> Ok, it is probably because of Travis didn't support this cache,

I didn't get this at all.

> why don't you use the prepared `mk/.cache`  in scripts/continues when some embox build/rebuild is required 

I do. Mybuild does this automatically. mk/.cache contains Mybuild parsing results. mk/.cache is preserved between steps. But not between jobs

> introducing new global `prepare` stage and keeping the builds order in mind?

With prepare we are saving compiler invocations, `.o` files. They are not related to mk-cache

> Well, you keep all re-builds of the same template to follow each other, ok then.

That's correct

alexkalmuk: > With prepare we are saving compiler invocations, .o files. They are not related to mk-cache

Which `.o` files? How they are used then?

alexkalmuk: Ah, you mean files of `x86/test/net`, ok

AntonKozlov:
```
build: 
  make confload-x86/test/net
  make
test:
  sed -i '/smth/htms/' conf/mods.config
  make # incremental build, do reuse most of .o files
```

alexkalmuk: Yes, thanks, I already said this one comment above https://github.com/embox/embox/pull/1801#discussion_r389894948

anton-bondarev:
As for me. It's great. Thanks a lot! :)

alexkalmuk:
@AntonKozlov We are ready to merge :) Notify us when you finished, please (I see two fixup's, so probably you want to squash them).

AntonKozlov:
yeah, I'll rebase and merge soon

alexkalmuk:
> yeah, I'll rebase and merge soon

Ok, thanks

PR #1802, opened 2020-03-08
fixing ls -la /dev/zero failure

Dawody:


src/fs/driver/devfs/devfs_oldfs.c
```diff
@@ -133,7 +133,7 @@ static struct inode *devfs_lookup(char const *name, struct inode const *dir) {
 
 	cdevtab = get_cdev_tab();
 	bdevs = get_bdev_tab();
-	for (i = 0; i < MAX_BDEV_QUANTITY; i++) {
+	for (i = 0; i < MAX_BDEV_QUANTITY && bdevs[i]; i++) {
```
alexkalmuk: But you will break here at the first `bdev[i] == NULL`, and other block devs will not be processed.

alexkalmuk: It's better to skip null devices while processing all array up to the end

Dawody: I thought that reaching Null when iterating on the array means that there is no more block devs.
so I added the condition to make it break when reaching Null pointer.

src/fs/driver/devfs/devfs_oldfs.c
```diff
@@ -134,6 +134,8 @@ static struct inode *devfs_lookup(char const *name, struct inode const *dir) {
 	cdevtab = get_cdev_tab();
 	bdevs = get_bdev_tab();
 	for (i = 0; i < MAX_BDEV_QUANTITY; i++) {
+		if(!bdevs[i])
+			continue;
```
alexkalmuk: Can you please follow our code style - https://github.com/embox/embox/wiki/Code-Style

```
if (...) {
    ...
}
```

src/fs/driver/devfs/devfs_oldfs.c
alexkalmuk: Some of these devices are possibly deleted, so we can skip them, so yes your new checking is pretty fine :)

src/fs/driver/devfs/devfs_oldfs.c
Dawody: I am sorry, I will read your code style to avoid such errors in the future.

Dawody:
@alexkalmuk 
I think that I did something wrong!
could you advise me ?

alexkalmuk:
@Dawody Please squash your commits into a single one and I think it could be merged.

PR #1803, opened 2020-03-09
Updated the fields in thread command

ameetsaahu:
Added two fields in the thread command viz. Task name and Stack size.

### DIFF
--- a/src/cmds/proc/thread.c
+++ b/src/cmds/proc/thread.c
@@ -18,6 +18,7 @@
 #include <kernel/thread.h>
 #include <kernel/task.h>
 #include <kernel/sched.h>
+#include <kernel/thread/thread_stack.h>
 
 static void print_usage(void) {
 	printf("Usage: thread [-h] [-s] [-k <thread_id>]\n");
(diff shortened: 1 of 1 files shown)


src/cmds/proc/thread.c
```diff
@@ -29,22 +30,24 @@ static void print_stat(void) {
...
+	printf(" %4s %16s %8s %7s %10s %12s\n", "id", "tid/Task name", "priority", "state", "time", "Stack size");
 
 	running = sleeping = suspended = 0;
 
 	sched_lock();
 	{
 		task_foreach(task) {
 			task_foreach_thread(t, task) {
-				printf(" %4d %4d %8d %c %c %c %c %9lds\n",
+				printf(" %4d %4d/%-11s %8d %c %c %c %c %9lds %9zu\n",
```
anton-bondarev: what is difference between `%-11s` and `%11s`

ameetsaahu: Its just making the printf look good.
%11s  does it as : Task_Name__
%-11s does it as : __Task_Name

where '_' denotes white space.

ameetsaahu:
Thanks...
And there goes my first ever PR :)

PR #1825, opened 2020-03-12
Implemented pthread_barrier_functions

### LINKED ISSUE #1645: Extend pthread compatibility
We need to implement following functions:
```
int pthread_barrier_destroy(pthread_barrier_t );
int pthread_barrier_init(pthread_barrier_t restrict, const pthread_barrierattr_t restrict, unsigned);
int pthread_barrier_wait(pthread_barrier_t );
int pthread_barrierattr_destroy(pthread_barrierattr_t *);
```
This changes are required for #1643

sobhaniiest:
Fixes #1645


sobhaniiest:
There are some mistakes in code style. I'll do another commit along with any further changes suggested by you after your reviews. Thanks :)

Deryugin:
Thanks for your contribution!

I'm trying to build this test, but it gives me compilation error

```
src/tests/posix/pthread/pthread_barrier_test.c:26:24: error: initialization of ‘pthread_barrier_t’ {aka ‘struct _pthread_barrier_t *’} from incompatible pointer type ‘struct _pthread_barrier_t **’ [-Werror=incompatible-pointer-types]
   26 |  pthread_barrier_t b = (pthread_barrier_t *) arg;
      |        
```

Fix it, please

sobhaniiest:
Hi @Deryugin. I have fixed the error and have updated a test and the code style. Please review.
Thanks :)

sobhaniiest:
The CI failed for the  TEMPLATE=usermode86/debug and showed
ci_error

I have not changed anything regarding this. Am I doing something wrong?

sobhaniiest:
But in the forked repo, it's showing all checks have passed, please have a check [here](https://github.com/sobhaniiest/embox/runs/506290096).

src/tests/posix/pthread/pthread_barrier_test.c
```diff
@@ -0,0 +1,43 @@
...
+#include <embox/test.h>
+#include <unistd.h>
+#include <stdlib.h>
+
+EMBOX_TEST_SUITE("posix/pthread_barrier_api");
+
+TEST_CASE ("Initialize barrier with pthread_barrier_init()") {
+	pthread_barrier_t barrier = NULL; 
+	pthread_barrierattr_t attr = (pthread_barrierattr_t) malloc(sizeof(*attr));
+	attr->pshared = PTHREAD_PROCESS_PRIVATE;
```
Deryugin: As I can see from the standard (https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_types.h.html), `pthread_barrierattr_t` is not explicitly required to be a pointer, so test should not rely on fact that it is a pointer, i.e. `sizeof(*attr)` may fail to compile if implementation will change (but will still be POSIX-compliant).

Also it means that you can't access `attr->pshared` directly.

sobhaniiest: Okay! [According to this](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_types.h.html), `pthread_barrier_t` also should not be a pointer type.  Right?

Deryugin: It's unspecified, so it's implementation-dependent.

Issue #1827, opened 2020-03-12
Clean `ld` flags in build.conf


Deryugin:
`-g` flag is ignored for gcc, but is being widely used in lots of templates. I suggest to remove it from all templates where gcc is used for compilation

```
> man ld
...
-g  Ignored.  Provided for compatibility with other tools.
```

`-N` flags is supposed to do the following thing:
```
       -N
       --omagic
           Set the text and data sections to be readable and writable.  Also, do not page-align the data segment, and disable linking against shared libraries.  If the output format supports Unix style magic numbers, mark the output as "OMAGIC". Note: Although a writable text section is allowed for PE-COFF targets, it does not conform to the format specification published by Microsoft.
```

So it
1) Sets `.text` and `.data` to be readable/writable, which we don't care about (or do we?)
2) Disables page-align for `.data` which is useless (we align it anyway in `mk/image/lds.S`)
3) Disables linking against shared libraries which can be done with `-static` flag with no side-effects

But also this -N flag breaks compilation of simple programs (as `src/cmds/load_app/hello.c`)
```
> cat main.c
int main() { return 0; }
> gcc main.c -N -nostdlib
/usr/bin/ld: warning: cannot find entry symbol _start; defaulting to 0000000000400241
/usr/bin/ld: a.out: error: PHDR segment not covered by LOAD segment
collect2: error: ld returned 1 exit status
```
Tested it with GCC 9 and GCC 6, same error message. I don't know if it's a GCC bug or not, but `gcc main.c -static -nostdlib` compiles well.

So I suggest to replace `-N` with `-static` (or even just get rid of this, because `-static` is used in `mk/flags.mk:251:override LDFLAGS  = -static -nostdlib`).

Deryugin:
Well, templates were cleaned in #1830; loadable apps were fixed in #1832. I'm closing this issue as it seems to be no arguments to keep `-N` ld flag.

PR #1828, opened 2020-03-12
Loadable apps fix

Deryugin:
This fixes `x86/test/units` build (some tests do fail, however).
Details are described in #1827, but I'm copying it here for the sake of clarity:

This -N flag breaks compilation of simple programs (as `src/cmds/load_app/hello.c`)
```
> cat main.c
int main() { return 0; }
> gcc main.c -N -nostdlib
/usr/bin/ld: warning: cannot find entry symbol _start; defaulting to 0000000000400241
/usr/bin/ld: a.out: error: PHDR segment not covered by LOAD segment
collect2: error: ld returned 1 exit status
```
Tested it with GCC 9 and GCC 6, same error message. I don't know if it's a GCC bug or not, but `gcc main.c -static -nostdlib` compiles well.

So I suggest to replace `-N` with `-static` (or even just get rid of this, because `-static` is used in `mk/flags.mk:251:override LDFLAGS  = -static -nostdlib`).

### DIFF
--- a/mk/script/loadable-build.mk
+++ b/mk/script/loadable-build.mk
@@ -1,5 +1,5 @@
 
-L_CFLAGS = -nostdlib -emain -fpie -N
+L_CFLAGS = -nostdlib -emain -fpie -static
 
 $(LOADABLE_DIR)/$(DST) : | $(LOADABLE_DIR)
 $(LOADABLE_DIR)/$(DST) : $(SRC)


AntonKozlov:
```
       -N
       --omagic
           Set the text and data sections to be readable and writable.  Also, do not page-align the data segment, and disable linking against shared
           libraries.  If the output format supports Unix style magic numbers, mark the output as "OMAGIC". Note: Although a writable text section is
           allowed for PE-COFF targets, it does not conform to the format specification published by Microsoft.
```

No page align was very important (mb even critical), please check loadable apps are really fine

Deryugin:
@AntonKozlov this flag disables align, so in worst case unaligned data may become page-aligned. May it really break something?

`load_app` is broken due to incorrect ELF parsing (it calculates wrong offset and jumps outside of RAM). But if offset is set properly, `loadable_hello_world` example executes fine. Sadly, with `-N` flag I can't compile it at all, so I can't really tell if it will handle sections in different way.
Do we have something more complicated example to see if it works too?

Deryugin:
@AntonKozlov please take a look at #1832

PR #1829, opened 2020-03-12
Extended PNG, GIF, ELF, XML format for file command

### LINKED ISSUE #1730: Extending file formats for `file` command
There is a `file` command (src/cmds/fs/file.c), currently it just checks if file is RIFF or text file.

This can be extended for ELF, PNG and GIF formats too, and it should be relatively simple.

Optionally, checking for HTML/XML/SVG file formats may be implemented too, but that's a little bit more complicated as header is not in the fixed position.

yog-singh:
Fixes #1730 
I've added PNG, GIF, ELF, XML file formats to the file command

### DIFF
--- a/src/fs/file_format.c
+++ b/src/fs/file_format.c
@@ -14,6 +14,10 @@
 enum fformat raw_get_file_format(uint8_t *file) {
 	/* signature of PNG files */
 	uint8_t png_signature[8] = {0x89, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a};
+	/* signature of ELF files */
+	uint8_t elf_signature[4] = {0x7f, 0x45, 0x4c, 0x46};
+	/* signature of XML files */
+	uint8_t xml_signature[21] = {0x3c, 0x3f, 0x78, 0x6d, 0x6c, 0x20, 0x76, 0x65, 0x72, 0x73, 0x69, 0x6f, 0x6e, 0x3d, 0x22, 0x31, 0x2e, 0x30, 0x22, 0x3f, 0x3e};
 	/* Is it a RIFF file? */
 	if (!memcmp(file, "RIFF", 4)) {
 		/* TODO check file consistency */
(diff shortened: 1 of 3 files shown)


src/fs/file_format.c
```diff
@@ -14,6 +14,10 @@
 enum fformat raw_get_file_format(uint8_t *file) {
 	/* signature of PNG files */
 	uint8_t png_signature[8] = {0x89, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a};
+	/* signature of ELF files */
+	uint8_t elf_signature[4] = {0x7f, 0x45, 0x4c, 0x46};
+	/* signature of XML files */
+	uint8_t xml_signature[21] = {0x3c, 0x3f, 0x78, 0x6d, 0x6c, 0x20, 0x76, 0x65, 0x72, 0x73, 0x69, 0x6f, 0x6e, 0x3d, 0x22, 0x31, 0x2e, 0x30, 0x22, 0x3f, 0x3e};
```
Deryugin: As I can see, you check for XML version 1.0 here. Why limit it to single verison?

It would be nice to print XML version in file command (but that's not necessary)

P.S. This line is way too long, it's recommended to stick to 80 symbols per line.

src/fs/file_format.c
```diff
@@ -14,6 +14,10 @@
 enum fformat raw_get_file_format(uint8_t *file) {
 	/* signature of PNG files */
 	uint8_t png_signature[8] = {0x89, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a};
+	/* signature of ELF files */
+	uint8_t elf_signature[4] = {0x7f, 0x45, 0x4c, 0x46};
```
Deryugin: Please use readable characters when it's meaningful (i.e. `'E'` instead of `0x45`), this way it's easier to see what's going on.

yog-singh: Thanks for the review, I'll make the required changes

src/fs/file_format.c
yog-singh: Should I just check for "<?xml" in the signature?

yog-singh:
One of the check fails. I can't figure out the reason.

src/fs/file_format.c
Deryugin: Yes, it would be better.

Deryugin:
> One of the check fails. I can't figure out the reason.

It's related to network issues with Travis (for some reason it fails to resolve domain name in network test), I'll restart this test.

Issue #1808, opened 2020-03-09
Fix pthread_kill


nilCux:
- Send signal 0 with pthread_kill will return -1 while 0 is expected
- Send signal to terminated threads behave wierd (Even crashed when the signal is SIGALRM) 

Test offered by #1716 , the part who might cause crash was written as comments, un-comment it if necessary.

Issue #1716, opened 2020-02-25
Tests for pthread_kill


alexkalmuk:
There is POSIX function `pthread_kill` - you can read more about this function in `man 3 pthread_kill`.  We have our own implementation of this function in Embox.

Please, add tests for this function.

nilCux:
Hi, I have sent a mail about participating GSoC from chenglin.theo.xu@gmail.com, may I work on this issue please? If I didn't misunderstand, it concerns adding a test for Embox version of `pthread_kill` which should do exactly the same thing as POSIX `pthread_kill`, right?

anton-bondarev:
> it concerns adding a test for Embox version of pthread_kill which should do exactly the same thing as POSIX pthread_kill, right?
Yes pthread_kill the same.
How to add Embox unit test you can see if look at src/test/posix/pthread/

nilCux:
Hi, sorry to bother you. I'm afraid I need some help. I just wrote a unit test, however I don't know how to run it. I put it under "embox/src/tests/posix/pthread" and added a module in the"Mybuild" file in this path. Afterwards, what should I do to run it?

nilCux:
Hi, I have finished a unit test including three cases:

-  A regular case of sending signal to a running thread
- Send signal 0 to a running thread
- Send signal to a terminated thread

However, there are three things I need to verify:

1. Using pthread_kill to send signal 0 should not be considered as an error, right?
2. Using pthread_kill to send a signal to a terminated thread will cause a crash, what should I do about that?
3. In at least Ubuntu, when I use pthread_kill to send a signal to a terminated thread, the return value of pthread_kill is 3. Is this value POSIX? Should I use this value in my test?

alexkalmuk:
All the things you asked are pretty good questions!

> Using pthread_kill to send signal 0 should not be considered as an error, right? Currently, it returns -1 in this case, I think it's wierd. At least in Ubuntu, the return value is 0.

https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_kill.html says
`As in kill(), if sig is zero, error checking shall be performed but no signal shall actually be sent.`

When we take a look at kill() now - https://pubs.opengroup.org/onlinepubs/9699919799/functions/kill.html we can see `If sig is 0 (the null signal), error checking is performed but no signal is actually sent. The null signal can be used to check the validity of pid.`

So yes, it looks like zero sig can be used to check if the thread is correct and do nothing.

[...]

PR #1838, opened 2020-03-13
Fix pthread_kill

nilCux:
#1808 
pthread_kill is  fixed. @anton-bondarev 
This commit is the modification in signal.c and pthread.h for the fix.
I think unit tests should be based on this.

### DIFF
--- a/src/compat/posix/include/pthread.h
+++ b/src/compat/posix/include/pthread.h
@@ -153,6 +153,8 @@ extern int   pthread_join(pthread_t, void **);
 extern int   pthread_key_create(pthread_key_t *, void (*)(void *));
 extern int   pthread_key_delete(pthread_key_t);
 
+extern int   pthread_kill(pthread_t thread, int sig);
+
 extern int   pthread_mutex_destroy(pthread_mutex_t *);
 //extern int   pthread_mutex_getprioceiling(const pthread_mutex_t *, int *);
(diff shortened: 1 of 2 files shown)


anton-bondarev:
@nilCux 
Your idea is good, but I think it's better to analyze kernel object (thread) in sigstate_send() routine 
I mean
```
	if (t->state & TS_EXITED)
		return ESRCH;
```
also there is following code
```
if (!check_range(sig, 1, _SIG_TOTAL))
		return -EINVAL; 
```
It checked that the number more or equal 1 and less _SIG_TOTAL
Why it's '1' not '0' I don't understand enough. It's required more investigation.

nilCux:
> @nilCux
> Your idea is good, but I think it's better to analyze kernel object (thread) in sigstate_send() routine
> I mean
> 
> ```
> 	if (t->state & TS_EXITED)
> 		return ESRCH;
> ```
> 
> also there is following code
> 
> ```
> if (!check_range(sig, 1, _SIG_TOTAL))
> 		return -EINVAL; 
> ```
> 
> It checked that the number more or equal 1 and less _SIG_TOTAL
> Why it's '1' not '0' I don't understand enough. It's required more investigation.

Hi, about signal 0, maybe you can check my conversation with alexkalmuk in #1716 .
Brief, for signal 0, no signal will be sent, pthread_kill executes error checks (if the target thread is available, etc) and returns 0.

PR #1839, opened 2020-03-13
scsi: Make scsi commands filling linear

alexkalmuk:
Remove `fixup` callbacks used to fill SCSI commands, it only makes the code more complicated.

### DIFF
--- a/src/drivers/scsi/scsi.c
+++ b/src/drivers/scsi/scsi.c
@@ -22,55 +22,50 @@ static inline struct usb_mass *scsi2mass(struct scsi_dev *dev) {
 	return member_cast_out(dev, struct usb_mass, scsi_dev);
 }
 
-static int scsi_cmd(struct scsi_dev *sdev, void *cmd, size_t cmd_len,
-		void *data, size_t data_len) {
+int scsi_do_cmd(struct scsi_dev *sdev, struct scsi_cmd *scmd) {
 	struct usb_mass *mass = scsi2mass(sdev);
-	struct scsi_cmd *scmd = cmd;
 	enum usb_direction usb_dir;
+	uint8_t scmd_opcode;
 
 	if (!sdev->attached) {
 		return -ENODEV;
 	}
-	if (scmd->scmd_opcode == SCSI_CMD_OPCODE_WRITE10) {
+	assert(scmd->scmd_len <= SCSI_CMD_MAX_LEN);
+	assert(scmd->scmd_buf);
+
+	scmd_opcode = scmd->scmd_buf[0];
+	if (scmd_opcode == SCSI_CMD_OPCODE_WRITE10) {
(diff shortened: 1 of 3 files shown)


src/drivers/scsi/scsi.c
```diff
@@ -22,81 +22,84 @@ static inline struct usb_mass *scsi2mass(struct scsi_dev *dev) {
...
-
-	assert(cmd->scmd_len <= SCSI_CMD_MAX_LEN);
-	memset(scmd + 1, 0, cmd->scmd_len - 1);
-	scmd[0] = cmd->scmd_opcode;
-
-	if (cmd->scmd_fixup) {
-		cmd->scmd_fixup(scmd, dev, cmd);
-	}
+static int scsi_send_inquiry(struct scsi_dev *dev) {
+	struct scsi_cmd cmd;
```
Deryugin: Looks like `cmd.scmd_opcode` is not set in this function, but it's read in `scsi_do_cmd()`. Set it to zero to make sure there's no garbage there?

alexkalmuk: yes, I will just remove this field, it's not used.

alexkalmuk: Fixed.

src/drivers/scsi/scsi.c
```diff
@@ -22,81 +22,87 @@ static inline struct usb_mass *scsi2mass(struct scsi_dev *dev) {
...
+	if ((data.dinq_devtype & SCSI_INQIRY_DEVTYPE_MASK)
 			!= SCSI_INQIRY_DEVTYPE_BLK) {
 		return -1;
 	}
 	return 0;
 }
 
 static int scsi_send_test_unit(struct scsi_dev *dev) {
-	struct scsi_cmd cmd = scsi_cmd_template_test_unit;
+	struct scsi_cmd cmd;
```
Deryugin: `cmd.sdata_buf` is not set here, so it may contain garbage from stack.

As I figured out, this address may end up being passed to dcache_flush here:
```
ehci_hcd.c:680 dcache_flush(req->buf, req->len);
```

It this address will be unaligned, `dcache_flush()` will try to align and flush it even if `size == 0`
```
void dcache_flush(const void *p, size_t size) {
...
	start &= ~(line_size - 1);
	while (start < end) {
		asm volatile ("mcr p15, 0, %0, c7, c14, 1" : : "r" (start));
		start += line_size;
	}
}
```

This may lead to unwanted flushing of random memory fragments.

Deryugin: I have similar concerns for `struct scsi_data_sense data;` & `struct scsi_data_cap10 data;` not being zeroed, but I can't make up example when this may fail for currently implemented drivers.

alexkalmuk: Ok, thank you, I fill these structures with zeroes.

PR #1862, opened 2020-03-14
Added test code for ffs()

### LINKED ISSUE #1845: Add tests for ffs()
Implement some test cases for standard [ffs()](https://pubs.opengroup.org/onlinepubs/009695399/functions/ffs.html) routine

binkoni:
This fixes #1845
ffs() from compat/libc/include/strings.h is declared in the header,
However it is not defined anywhere.
So I made this as an inline function that calls bit_ffs() before adding
test code.

Signed-off-by: Byeonggon Lee <gonny952@gmail.com>


src/compat/libc/string/ffs.c
```diff
@@ -0,0 +1,21 @@
+/**
+ * @file
+ * @brief Implementation of ffs() function.
+ *
+ * @date 20.03.15
+ * @author Byeonggon Lee
+ */
+
+int ffs(int i) {
```
alexkalmuk: Are you sure this implementation is required? You can search for a similar thing in embox.

alexkalmuk: I mean you can try to use existing bit functions.

alexkalmuk: Please do not squash the commits next time :) You can squash them right before merge.

Now I see that `bit_ffs` uses `unsigned long` as an argument. But `ffs` uses int.

alexkalmuk: So probably we cannot use `bit_ffs` here.

alexkalmuk: So we should check something like `ffs(-1)` in the tests too.

alexkalmuk:
@binkoni Can I please ask you to add `ffs` implemention like it was before squash and also addtional tests for negative arguments?

binkoni:
@alexkalmuk I also added negative arguments tests

alexkalmuk:
> @alexkalmuk I also added negative arguments to the test and it passes the test
> 
> > Can I please ask you to add ffs implemention like it was before
> 
> Do you mean my removed ffs implementation?

Yes.

binkoni:
@alexkalmuk 
```C
int ffs(int in) {
	for (int i = 0; i < sizeof(int) * 8; ++i)
		if (in & (1 << i))
			return i + 1;
	return 0;
}
```
My implementation was this. should I add it like before commit?

src/compat/libc/string/ffs.c
binkoni: > Please do not squash the commits next time :) You can squash them right before merge
Okay I got it.

alexkalmuk:
> @alexkalmuk
> 
> ```c
> int ffs(int in) {
> 	for (int i = 0; i < sizeof(int) * 8; ++i)
> 		if (in & (1 << i))
> 			return i + 1;
> 	return 0;
> }
> ```
> 
> My implementation was this. it also passes the test. should I add it again like previous(removed) commit?

Yes, please. By the way, it would be great to have two commits - one with implementation, and one with the test.

binkoni:
I added my ffs() implementation too and it passes the test.
If you are fine, I'll squash the 3 commits into 2 commits.

src/compat/libc/include/strings.h
```diff
@@ -13,6 +13,7 @@
 #include <defines/size_t.h>
 
 #include <sys/cdefs.h>
+#include <util/bit.h>
```
alexkalmuk: Can you please remove this since it is not required anymore?

src/compat/libc/string/ffs.c
```diff
@@ -0,0 +1,16 @@
+/**
+ * @file
+ * @brief Implementation of ffs() function.
+ *
+ * @date 20.03.15
```
alexkalmuk: date?

src/tests/stdlib/ffs_test.c
```diff
@@ -0,0 +1,39 @@
+/**
+ * @file
+ * @brief Test of ffs() function.
+ *
+ * @date 20.03.15
```
alexkalmuk: date?

src/tests/stdlib/ffs_test.c
```diff
@@ -0,0 +1,39 @@
...
+ * @author: Byeonggon Lee
```
alexkalmuk: remove `:` please.

src/compat/libc/string/ffs.c
```diff
@@ -0,0 +1,16 @@
...
+ * @author Byeonggon Lee
+ */
+
+#include <strings.h>
+
+int ffs(int in) {
+	for (int i = 0; i < sizeof(int) * 8; ++i)
+		if (in & (1 << i))
+			return i + 1;
```
alexkalmuk: Can you please use braces? https://github.com/embox/embox/wiki/Code-Style#tabs-and-brackets

src/tests/stdlib/Mybuild
```diff
@@ -82,4 +82,9 @@ module strchr_test {
 	depends embox.framework.LibFramework
 }
 
+module ffs_test {
+	source "ffs_test.c"
 
+	depends embox.compat.libc.str
```
alexkalmuk: Remove these dependencies please, we are trying to avoid test dependencies.

binkoni: Okay

alexkalmuk:
@binkoni Great! Can you squash the changes into 2 commits and we can merge.

binkoni:
Oh sorry, looks like I messed up other commits. I'll close this pull request and create new pull request

PR #1864, opened 2020-03-15
compat/libc: Added test code for ffs()

### LINKED ISSUE #1845: Add tests for ffs()

binkoni:
This fixes #1845
This commit adds test code for ffs()

Signed-off-by: Byeonggon Lee <gonny952@gmail.com>


alexkalmuk:
This PR continues https://github.com/embox/embox/pull/1862

> So I made this as an inline function that calls bit_ffs() before adding
test code.

I think it is not actual anymore.

alexkalmuk:
Can you please just add your test to `x86/test/units`? (you may not run this template, we are still fixing its launching).

binkoni:
@alexkalmuk Thanks I fixed the commit message and updated template.

PR #1867, opened 2020-03-16
execv: fix execv returns 0 and doesn't set errno

### LINKED ISSUE #1335: execv return 0 and does not set errno on errors
Look at #1334 to reproduce this bug

```
	/* difference from posix: execv return 0 and does not set errno on errors */
	if (optind < argc) {
		execv(argv[optind], argv+optind);
		printf("env: execv: No such command '%s'\n", argv[optind]);
		return -ENOENT;
	}
```

### LINKED ISSUE #1702: service_run() does not consider result of starting service

Ahish9009:
Closes #1335:
Behaves like execv in posix: returns -1 on failure and sets errno to the value of error that occurred. Seems to have fixed #1334, but does not seem to have helped #1702.

### DIFF
--- a/src/cmds/sys/env.c
+++ b/src/cmds/sys/env.c
@@ -68,8 +68,7 @@ int main(int argc, char *argv[]) {
 	/* difference from posix: execv return 0 and does not set errno on errors */
 	if (optind < argc) {
 		execv(argv[optind], argv+optind);
-		printf("env: execv: No such command '%s'\n", argv[optind]);
-		return -ENOENT;
+		return -errno;
 	}
 	
 	if (environ)
--- a/src/compat/posix/proc/exec.c
+++ b/src/compat/posix/proc/exec.c
@@ -17,6 +17,8 @@
 #include <kernel/task/resource.h>
 #include <hal/vfork.h>
 
+#include <errno.h>
+
 static const char * exec_cmd_name(const char *path) {
 	size_t path_len;
 
@@ -60,13 +62,11 @@ int exec_call(void) {
 			task_self_module_ptr_set(cmd2mod(cmd));
 			ecode = cmd_exec(cmd, c, v);
 		} else {
-			ecode = ENOENT;
+			ecode = -ENOENT;
+			errno = ENOENT;
 		}
 	}
 
-	if (ecode == 0) {
-		task_exit(0);
-	}
 	return ecode;
 }
 
@@ -116,9 +116,13 @@ int execv(const char *path, char *const argv[]) {
 	vfork_child_done(task, task_exec_callback, NULL);
 
 	/* Not vforked */
-	exec_call();
+	int ecode = exec_call();
 
-	return 0;
+	if (ecode == 0) {
+		task_exit(0);
+	}
+	return -1;
+
 }
 
 int execve(const char *path, char *const argv[], char *const envp[]) {


Ahish9009:
Not sure what's wrong in the last check. Seems to have stopped while pinging 10.0.2.10.
Don't see why my changes should have affected it.

anton-bondarev:
> Not sure what's wrong in the last check. Seems to have stopped while pinging 10.0.2.10.
> Don't see why my changes should have affected it.

Don't worry about this. It's not your fault. We restarted CI and it passed succeed

anton-bondarev:
@Ahish9009 Hi, Thanks
I think You on the right way.
But Why do you want to differ in two situations

1. When cmd_exec return with errors
2. When cmd has not been found

I supposed You can just return ecode from exec_call and then return it in execv

Ahish9009:
I had a look at the ```cmd_exec``` code and realised that the check for whether the command exists is happening already, but is returning EINVAL instead of ENOENT, which I fixed.
Further, not testing whether the cmd exists in ```exec_call()``` caused the issue of passing a NULL pointer to ```cmd2mod()```, which is also fixed now.
Please have a look at the changes.

src/cmds/sys/env.c
```diff
@@ -68,8 +68,7 @@ int main(int argc, char *argv[]) {
 	/* difference from posix: execv return 0 and does not set errno on errors */
 	if (optind < argc) {
 		execv(argv[optind], argv+optind);
-		printf("env: execv: No such command '%s'\n", argv[optind]);
```
anton-bondarev: I think It's better to keep printf but the message change to "env: execv: command '%s' failed with error code %d"

src/include/framework/cmd/types.h
```diff
@@ -37,6 +37,7 @@ struct cmd_mod {
 };
 
 static inline const struct mod *cmd2mod(const struct cmd *cmd) {
+	if (!cmd) return NULL;
```
anton-bondarev: I think You should check this before calling cmd2mod

src/compat/posix/proc/exec.c
```diff
@@ -55,18 +57,10 @@ int exec_call(void) {
 		const struct cmd *cmd;
 
 		cmd = cmd_lookup(cmd_name);
-
```
anton-bondarev: Why you remove this?
Revert please

src/cmds/sys/env.c
Ahish9009: Okay, will make this change, but that'll mean I'll be printing errno(since ecode won't be returned from execv, as POSIX requires execv to return -1 on errors). 
So if I'm printing errno, isn't it better to use perror()?

src/compat/posix/proc/exec.c
Ahish9009: I removed it because cmd_exec checks if the cmd exists anyways, so found that this check is redundant. Should I revert back though?

src/include/framework/cmd/types.h
Ahish9009: I did this so that I don't have to perform this check in the exec_call() function.

src/cmds/sys/env.c
anton-bondarev: Ok, just remove.
Thanks

src/compat/posix/proc/exec.c
anton-bondarev: Yes, I see. But I don't understand why did you move checking from here? Why did you place it in cmd2mod(). why did you move it at all?

src/framework/cmd/core.c
```diff
@@ -23,8 +23,10 @@ extern void getopt_init(void);
 int cmd_exec(const struct cmd *cmd, int argc, char **argv) {
 	int err;
 
-	if (!cmd)
-		return -EINVAL;
+	if (!cmd) {
```
anton-bondarev: the same question.
Why did you change src/framework/cmd/core.c?
why you could not do it in src/compat/posix/proc/exec.c?

src/compat/posix/proc/exec.c
Ahish9009: My aim was to remove setting errno to ENOENT in the exec_call() function, so I removed the check in exec_call() and simply used the check already present in cmd_exec(). 
On running however, as cmd was NULL, cmd2mod() caused errors as it was trying to perform its operation on a NULL pointer. So I included the check there too.
So should I put the check back here and remove the check in cmd2mod()?

src/framework/cmd/core.c
Ahish9009: It can be done in either place, but I thought it'd be better to do it in the cmd_exec() function as the error is during execution of the cmd, but I can revert back if needed, not a problem :)

src/compat/posix/proc/exec.c
anton-bondarev: > My aim was to remove setting ecode and errno to ENOENT
Why? :)

I thought your aim is "execv returns 0 and doesn't set errno" :)

Ahish9009: > But Why do you want to differ in two situations
>
> When cmd_exec return with errors
> When cmd has not been found
> I supposed You can just return ecode from exec_call and then return it in execv

I thought when you said this in the earlier comment, you wanted me to set all errors in the cmd_exec() function itself, my bad. Should I revert it back to the way it was?

anton-bondarev: > Should I revert it back to the way it was?
Yes please.
I'm sorry that uncleanly put it

Ahish9009: Done :)

src/framework/cmd/core.c
Ahish9009: Reverted

src/include/framework/cmd/types.h
Ahish9009: Reverted

anton-bondarev:
@Ahish9009 Thank you. I've approved PR.
Please rebase and squash commits

Ahish9009:
> @Ahish9009 Thank you. I've approved PR.
> Please rebase and squash commits

Done

Issue #1837, opened 2020-03-13
Arch-independent `cpuinfo` command


Deryugin:
Right now `cpuinfo` commands just prints content of this structure, which is x86-specific.

```
struct cpu_info {
	char vendor_id[13];
	unsigned int base_freq;
	unsigned int max_freq;
	unsigned int FPU;
	unsigned int SSE;
	unsigned int MMX;
};
```

It would be nice to extend it somehow to work with other architectures as well.

I guess it's ok to have some generic fields like `vendor_name[]` and `cpu_frequency`, but for extensions we need to rewrite this utility to use arch-agnostic interface.

May be something like this would work

```
struct cpu_feature {
    char name[];
    bool is_supported;
}

int get_cpu_features(struct cpu_feature **features_array) {
    /* Returns number of known features for this arch
        and fills array with corresponding info */
}
```

checker5965:
Hi @Deryugin - I can take a look at this issue. Could you please assign it to me?

Just to be clear, we want to make the `cpuinfo` command arch independent to solve this issue. Do we also need to implement the functionality for other architectures apart from x86?

anton-bondarev:
> Hi @Deryugin - I can take a look at this issue. Could you please assign it to me?

@checker5965 I've assigned it

> Just to be clear, we want to make the `cpuinfo` command arch independent to solve this issue. Do we also need to implement the functionality for other architectures apart from x86?

Yes, we want to make `cpuinfo` arch-independent. It's useful to enable it on some arm template

Deryugin:
@checker5965 
> Do we also need to implement the functionality for other architectures apart from x86?
Sure, it would be nice to detect if FPU/Neon extensions are implemented for given ARM platform, for example. But that's another issue, I guess.

checker5965:
@Deryugin Could you please close this issue :)

PR #1872, opened 2020-03-16
cmds: x86/lib: Update `cpuinfo` command to be arch-independent

checker5965:
The `cpuinfo` command previously was x86 specific.
Now, it expects `struct cpu_info` to return number
of features known for an arch, and prints the results
accordingly.

This closes issue #1837.


src/arch/x86/lib/cpu_performance/cpu_info.h
```diff
@@ -1,13 +1,15 @@
...
+};
+
 struct cpu_info {
 	char vendor_id[13];
-	unsigned int base_freq;
-	unsigned int max_freq;
-	unsigned int FPU;
-	unsigned int SSE;
-	unsigned int MMX;
+	struct cpu_feature feature[5];
```
Deryugin: Same here, use macro to define array size.

src/arch/x86/lib/cpu_performance/cpu_info.c
```diff
@@ -21,10 +21,12 @@ static void set_cpu_features(struct cpu_info *info) {
 					: "=g"(reg)
 					:
 					: "%eax", "%ebx", "%ecx", "%edx");
-
-	info->FPU = reg & 0x1;
-	info->MMX = (reg >> 23) & 0x1; 
-	info->SSE = (reg >> 25) & 0x1;
+	strcpy(info->feature[2].name, "FPU");
```
Deryugin: Bad thing here is that programmer should track feature numbers manually.
What if we move some piece of code to another function or another file?
Some indices may end up being unused, other may be less than `current_cpu.num_features` if programmer forgets to increment it.
It's easy to track if you have 5 features, but, for example, `lscpu` prints 112 features for my CPU on my host machine, tracking this stuff may become painful :)

So a better solution would be tracking it within some simple handle, like
```
int set_cpu_feature(struct cpu_info *info, const char *name, unsigned val) {
   /* Set info->feature[info->num_features],
    * then increment info->num_features;
    * check for overflow etc */
}
```

src/arch/x86/lib/cpu_performance/cpu_info.h
```diff
@@ -1,13 +1,15 @@
 #ifndef X86_CPU_INFO_H_
 #define X86_CPU_INFO_H_
 
+struct cpu_feature {
```
Deryugin: As this structure is arch-agnostic, it's better to put it in `src/include/hal/cpu_info.h`

checker5965: Done in the latest fixup commit!

src/arch/x86/lib/cpu_performance/cpu_info.c
checker5965: Done in the latest fixup commit!

src/arch/x86/lib/cpu_performance/cpu_info.h
checker5965: Done in the latest fixup commit!

src/arch/x86/lib/cpu_performance/cpu_info.h
```diff
@@ -1,13 +1,15 @@
...
+	char name[13];
```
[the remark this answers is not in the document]
checker5965: Done in the latest fixup commit!

src/arch/x86/lib/cpu_performance/cpu_info.h
```diff
@@ -1,16 +1,9 @@
 #ifndef X86_CPU_INFO_H_
 #define X86_CPU_INFO_H_
 
-struct cpu_feature {
-	char name[13];
-	unsigned int val; 
-};
+#define FEATURE_NAME_LEN 32
```
checker5965: `FEATURE_NAME_LEN` can either be defined here or in `include/hal/cpu_info.h`.

If we let it be here, then the name length can be different for every Arch. What do you think will be better?

Deryugin: I think it's better to define it big enough for all architectures.

checker5965: Alright, if there is anything else that needs to change with this PR, let me know. I will create another fixup commit accordingly. Thanks!

src/arch/x86/lib/cpu_performance/cpu_info.c
```diff
@@ -21,10 +23,18 @@ static void set_cpu_features(struct cpu_info *info) {
...
+	assert(current_feature < CURRENT_ARCH_NUM_FEATURES);
+	strcpy(info->feature[current_feature].name, "FPU");
+	info->feature[current_feature++].val = reg & 0x1;
+	
+	assert(current_feature < CURRENT_ARCH_NUM_FEATURES);
+	strcpy(info->feature[current_feature].name, "MMU");
+	info->feature[current_feature++].val = (reg >> 23) & 0x1;
+	
+	assert(current_feature < CURRENT_ARCH_NUM_FEATURES);
+	strcpy(info->feature[current_feature].name, "SSE");
```
Deryugin: Why not create function to avoid copying of `assert/strcpy/current_feature++`? I mean it may look like this

```
int set_cpu_feature(struct cpu_info *info, const char *name, unsigned val) {
  int current_feature = info->feature_count;
  /* todo: sanity checks for argument */
  strcpy(info->feature[current_feature], name);
  info->feature[current_feature] = val;
  info->feature_count++;
}

set_cpu_feature(info, "FPU", reg & 0x1);
set_cpu_feature(info, "MMU", (reg >> 23) & 0x1);
set_cpu_feature(info, "SSE", (reg >> 25) & 0x1);
```

src/arch/x86/lib/cpu_performance/cpu_info.h
checker5965: Fixed.

src/arch/x86/lib/cpu_performance/cpu_info.c
checker5965: Have tried to implement something of the sort in the latest fixup. Please review, thanks :)

src/arch/x86/lib/cpu_performance/cpu_info.c
```diff
@@ -7,11 +7,20 @@
...
 
 struct cpu_info current_cpu;
 
+void set_feature_val(struct cpu_info *info, const char *name, unsigned int val) {
+	int current_feature = info->feature_count;
+	assert((current_feature < CURRENT_ARCH_NUM_FEATURES) && "Exceeded Number of Features!");
+	strcpy(info->feature[current_feature].name, name);
+	info->feature[current_feature].val = val;
+	info->feature_count += 1;
+}
```
Deryugin: This function is arch-independent, so it can be moved out of this (x86/lib/cpu_info) module.

checker5965: I tried moving it in `hal/cpu_info.h` but I was getting build errors.

src/arch/x86/lib/cpu_performance/cpu_info.c
```diff
@@ -39,10 +47,10 @@ static void set_cpu_freq(struct cpu_info *info) {
 					: "=g"(reg[0]), "=g"(reg[1])
 					:
 					: "%eax", "%ebx", "%ecx", "%edx");
-
-	info->base_freq = reg[0];
-	info->max_freq = reg[1];
-
+	
+
```
Deryugin: Please, remove second blank line; we're avoiding placing more than one blank line in a row https://github.com/embox/embox/wiki/Code-Style

src/cmds/cpuinfo.c
```diff
@@ -12,12 +12,12 @@
...
 	struct cpu_info *info = get_cpu_info();
-	printf("CPU Vendor ID:              %s\n", info->vendor_id);
-	printf("CPU Base Frequency (MHz):   %u\n", info->base_freq);
-	printf("CPU Max Frequency (MHz):    %u\n", info->max_freq);
-	printf("CPU FPU Support:            %s\n", info->FPU == 1 ? "true" : "false");
-	printf("CPU MMX Support:            %s\n", info->MMX == 1 ? "true" : "false");
-	printf("CPU SSE Support:            %s\n", info->SSE == 1 ? "true" : "false");
+	printf("%-20s %s\n", "CPU Vendor ID ", info->vendor_id);
+	
+	for(int i = 0; i < CURRENT_ARCH_NUM_FEATURES; i++) {
```
Deryugin: No need to iterate to `CURRENT_ARCH_NUM_FEATURES`; we can simply look at `info->feature_count`

checker5965: Wouldn't `CURRENT_ARCH_NUM_FEATURES` and `info->feature_count` be the same number?

src/arch/x86/lib/cpu_performance/cpu_info.c
checker5965: Sorry for this! Will fix.

src/cmds/cpuinfo.c
Deryugin: I think that better solution would be to define `CURRENT_ARCH_NUM_FEATURES` big enough for all architectures (so we don't need to change it every time we add check for a new feature)

checker5965: Okay, fair point. Will fix this, thanks!

src/arch/x86/lib/cpu_performance/cpu_info.c
checker5965: Why does `x86/lib/cpu_performance/cpu_counter.c` have `#include <hal/cpu_info.h>` ? This was causing double compilation and hence a build failure. Should I remove it?

checker5965: And we don't need the x86 `cpu_info.h` file anymore.

Deryugin: Hm.. It should not lead to compilation error; header guard should protect header file from double-include
```
#ifndef HAL_CPU_INFO_H_
#define HAL_CPU_INFO_H_
...
#endif
```

checker5965: Yeah I thought so, but as soon as I removed the include in `cpu_counter.c` the build was successful. And in the other case was indeed recompilation of the function.

Deryugin: Ah, you tried to move the whole function.

C preprocessor works straightforward: when you do `#include <file.h>`, it simply substitutes content of the file.

So imagine you have files `f1.c, f2.c, f3.h`,
```
f3:
int some_func() { return 0; }
```
```
f2:
#include <f3.h>

int other_func() { return some_func(); }
```

```
f1:
#include <f3.h>

int some_other_func() { return some_func(); }
```

After preprocessor does its stuff you get:
```
f2:
int some_func() { return 0; }

int other_func() { return some_func(); }
```

```
f1:
int some_func() { return 0; }

int some_other_func() { return some_func(); }
```

After that if you try to link `f1.o` and `f2.o`, linker doesn't know which `some_func()` you actually want to call (non-static functions are visible to all other c-files).

Deryugin: As you can see, this problem raises when you define (not just declare) functions in header files.

So there are two solutions:
1) Declare function as `static`, so it will be unique for every C file.
2) Define it in some other C file and replace definition with declaration in the header file.

checker5965: Alright, in which directory should I place the new C file?

Deryugin: Well, I guess it's a good idea to create `src/lib/cpu_info` and create a module for this library.

That's pretty narrow scope for a library; may be it will be merged with something else in the future; but for now let's do it this way :)

checker5965: Hey @Deryugin, just committed these changes in a new fixup commit :)

src/lib/cpu_info/libcpu_info.h
```diff
@@ -1,7 +1,5 @@
-#ifndef HAL_CPU_INFO_H_
```
Deryugin: Please add header guard to avoid double `#include`

src/lib/cpu_info/cpu_info_funcs.c
```diff
@@ -0,0 +1,11 @@
+#include <lib/libcpu_info.h>
+#include <string.h>
+#include <assert.h>
+
+extern void set_feature_val(struct cpu_info *info, const char *name, unsigned int val) {
+    int current_feature = info->feature_count;
+	assert((current_feature < MAX_NUM_FEATURES) && "Exceeded Number of Features!");
```
Deryugin: I see mixing of tabs and spaces here, use only tabs, please.

src/lib/cpu_info/cpu_info_funcs.c
```diff
@@ -0,0 +1,11 @@
+#include <lib/libcpu_info.h>
+#include <string.h>
+#include <assert.h>
+
+extern void set_feature_val(struct cpu_info *info, const char *name, unsigned int val) {
```
Deryugin: No need to write `extern` if you define function here.

src/lib/cpu_info/cpu_info_funcs.c
checker5965: Really Sorry! I copy-pasted and this happened probably due to my text editor's settings. Will fix.

src/lib/cpu_info/libcpu_info.h
checker5965: This is outdated, `hal/cpu_info.h` is not needed anymore.

src/lib/cpu_info/cpu_info_funcs.c
checker5965: Right, Thanks!

src/lib/cpu_info/libcpu_info.h
checker5965: Added in `lib/cpu_info/libcpu_info.h`.

checker5965:
Done, Thanks!

PR #1874, opened 2020-03-16
Add : pthread_barrier functions

### LINKED ISSUE #1645: Extend pthread compatibility

sobhaniiest:
Fix #1645 
Hi, @Deryugin 
I was facing a problem with the previous branch. So, I made a new branch and a new PR for the pthread_barrier functions #1645.
Please review this.
Thanks :)

### DIFF
--- a/src/compat/posix/include/pthread.h
+++ b/src/compat/posix/include/pthread.h
@@ -18,7 +18,7 @@
  * the <time.h> header.
  */
 #include <sched.h>
-
+#include <semaphore.h>
 
 #include <kernel/thread/thread_flags.h>
 #include <kernel/thread/sync/cond.h>
@@ -50,7 +50,7 @@ typedef struct thread *pthread_t;
 #define PTHREAD_INHERIT_SCHED       THREAD_FLAG_PRIORITY_INHERIT
 #define PTHREAD_CREATE_DETACHED     THREAD_FLAG_DETACHED
 
-
+#define PTHREAD_BARRIER_SERIAL_THREAD -1
 
 typedef struct pthread_attr {
 	uint32_t flags; /* scope, inherit, detachstate */
@@ -59,7 +59,6 @@ typedef struct pthread_attr {
 	int policy;
 	struct sched_param sched_param;
 } pthread_attr_t;
-
 
 typedef cond_t pthread_cond_t;
 
@@ -88,6 +87,24 @@ typedef size_t pthread_key_t;
 
 
 typedef pthread_mutex_t pthread_once_t;
+
+struct _pthread_barrier_t {
+    unsigned int in;
+    unsigned int round;
+    unsigned int count;
+    int istep;
+    int pshared;
+    int initialize_flag;
+    int in_use;
+    sem_t lock[2];
+    pthread_mutex_t mutex_lock;
+};
+
+typedef  struct _pthread_barrierattr_t {
+    int pshared;
+} pthread_barrierattr_t;
+
+typedef struct _pthread_barrier_t *pthread_barrier_t;
 
 #define PTHREAD_ONCE_INIT PTHREAD_MUTEX_INITIALIZER
 
@@ -201,6 +218,13 @@ extern int   pthread_setspecific(pthread_key_t, const void *);
 extern void  pthread_testcancel(void);
 
 extern int   pthread_setschedprio(pthread_t, int);
+
+extern int   pthread_barrier_init(pthread_barrier_t *, 
+    const pthread_barrierattr_t *, unsigned int);
+extern int   pthread_barrier_wait(pthread_barrier_t *);
+extern int   pthread_barrier_destroy(pthread_barrier_t *);
+extern int   pthread_barrierattr_init(pthread_barrierattr_t *);
+extern int   pthread_barrierattr_destroy(pthread_barrierattr_t *);
 
 __END_DECLS
 
--- a/src/compat/posix/pthread/Mybuild
+++ b/src/compat/posix/pthread/Mybuild
@@ -6,7 +6,10 @@ static module pthreads {
 	source "pthread_cond.c"
 	source "pthread_mutex.c"
 	source "pthread_cancel.c"
+	source "pthread_barrier.c"
 
+	depends embox.compat.libc.all
+	depends embox.compat.posix.semaphore
 	depends embox.kernel.thread.core
 	depends embox.kernel.thread.sync
 	depends embox.kernel.task.syslib.thread_key_table
--- a/src/compat/posix/pthread/pthread_barrier.c
+++ b/src/compat/posix/pthread/pthread_barrier.c
@@ -1,0 +1,209 @@
+/**
+ * @file   
+ * @brief  pthread_barrier functions
+ * @date   11.03.2020
+ * @author Sobhan Mondal <inro20fdm@gmail.com>
+ */
+
+#include <stdlib.h>
+#include <errno.h>
+#include <pthread.h>
+#include <semaphore.h>
+#include <util/err.h>
+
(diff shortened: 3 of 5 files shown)


src/tests/posix/pthread/pthread_barrier_test.c
```diff
@@ -0,0 +1,44 @@
...
+#include <unistd.h>
+#include <stdlib.h>
+
+EMBOX_TEST_SUITE("posix/pthread_barrier_api");
+
+TEST_CASE ("Initialize barrier with pthread_barrier_init()") {
+	pthread_barrier_t barrier = NULL; 
+	pthread_barrierattr_t *attr;
+	attr = (pthread_barrierattr_t *) malloc(sizeof(pthread_barrierattr_t));
+	attr->pshared = PTHREAD_PROCESS_PRIVATE;
```
Deryugin: But why do you still try to access fields of `pthread_barrierattr_t`? It's implementation-defined, it means that this test is POSIX-incompatible; implementation may be changed later and still comply to standard, but this test will not work.

Let me give you example: `.pshared` field may be renamed to `.random_name12345`, it still will be POSIX (because it's details of implementation).
Another example: for glibc/x86 this structure defined this way:
```
typedef union
{
  char __size[__SIZEOF_PTHREAD_BARRIERATTR_T];
  int __align;
} pthread_barrierattr_t;
```
As you can see, no `pshared` field at all.

Tests should not exploit any implementation details besides API (i.e. test behavior) for two reasons:
1. If you think about implementation details when writing tests, you may end up tune test to just make it work.
2. Implementation may be changed, but behavior should be the same. If you'll need to rewrite tests every time you change implementation, you'll spend much more time for that.

sobhaniiest: > It's implementation-defined, it means that this test is POSIX-incompatible; implementation may be changed later and still comply to standard, but this test will not work.
Thanks for the detail explanation, I misunderstood this part.
So, the `pthread_barrierattr_t` will be initialized by the `pthread_barrier` API, in that case, we can use `pthread_barrierattr_init` or can initilize in `pthread_barrier_init`, right?

sobhaniiest: I have added `pthread_barrierattr_init` which will initialize `barrier attr`.

src/compat/posix/pthread/pthread_barrier.c
```diff
@@ -0,0 +1,207 @@
...
+
+	if (NULL == barrier || 0 == count) {
+		return EINVAL;
+	}
+
+	if (NULL != (*barrier)) {
+        if (1 == (*barrier)->initialize_flag) {
+            return EBUSY;
+        }
+    }
```
Deryugin: I see mix of spaces and tab here. Please use tabs :)

src/compat/posix/include/pthread.h
```diff
@@ -60,6 +60,22 @@ typedef struct pthread_attr {
...
 
+struct _pthread_barrier_t {
+    unsigned int in;
+    unsigned int round;
+    unsigned int count;
+    int istep;
+    int pshared;
+    int initialize_flag;
+    int in_use;
+    sem_t lock[2];
```
Deryugin: I don't quite understand why do we need two locks. Isn't 1 semaphore enough?

src/compat/posix/pthread/pthread_barrier.c
sobhaniiest: Done

src/compat/posix/include/pthread.h
sobhaniiest: Two semaphores are used to avoid race condition in between semaphore wait and post in case of multiple rounds.
Supposing, in a program, we initialize [count](https://github.com/sobhaniiest/embox/blob/7b445551739aedf49ee5812286ddc45553c92f7a/src/compat/posix/pthread/pthread_barrier.c#L38) by 3 and we are calling `pthread_barrier_wait` 6 times, so there will be 2 rounds. In the first round, when the 3rd `pthread_barrier_wait` will be called, it will go for [sem_post_multiple](https://github.com/sobhaniiest/embox/blob/7b445551739aedf49ee5812286ddc45553c92f7a/src/compat/posix/pthread/pthread_barrier.c#L128) and at that time if 4th `pthread_barrier_wait` is called then we don't want to use the same semaphore as it is in use (sem_post for the threads in 1st round). [Here](https://github.com/sobhaniiest/embox/blob/7b445551739aedf49ee5812286ddc45553c92f7a/src/compat/posix/pthread/pthread_barrier.c#L110) the value of `ib->istep` is alternated to use 2nd semaphore in the second round.

src/compat/posix/pthread/pthread_barrier.c
sobhaniiest: Done.

src/compat/posix/pthread/pthread_barrier.c
```diff
@@ -0,0 +1,207 @@
...
+	} else {
+		ib->istep = 1;
+	}
+	step = ib->istep;
+
+	pthread_mutex_init(&mutex_lock, NULL); 
+	pthread_mutex_lock(&mutex_lock); 
+	(ib->in)++;
+	pthread_mutex_unlock(&mutex_lock); 
+	pthread_mutex_destroy(&mutex_lock);
```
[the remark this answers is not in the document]
sobhaniiest: > Other fields are not protected like this (e.g. ib->istep few lines above).

I have added mutex_lock for the others also.

> You create mutex on stack, it means that another function trying to access the same barrier will have it's own mutex (as different threads have different stacks), so this mutex will be never accessed from another thread. It means it doesn't actually protect against race condition. It's more reliable to create this mutex in struct _pthread_barrier_t, so different threads will "see" the same mutex.

Thank you for clearing this concept. I've done the mistake. Now, I have declared `mutex_lock` in `pthread_barrier_t`.

src/compat/posix/include/pthread.h
sobhaniiest: I've also added a test for multiple rounds. [Here](https://github.com/sobhaniiest/embox/blob/7b445551739aedf49ee5812286ddc45553c92f7a/src/tests/posix/pthread/pthread_barrier_test.c#L65)

anton-bondarev:
@sobhaniiest Hi, My apologies for the late response.
Could you split the commit into 3 parts (tests, implementation and pthread.h modification)?

src/compat/posix/pthread/Mybuild
```diff
@@ -6,7 +6,10 @@ static module pthreads {
 	source "pthread_cond.c"
 	source "pthread_mutex.c"
 	source "pthread_cancel.c"
+	source "pthread_barrier.c"
 
+	depends embox.compat.libc.all
```
anton-bondarev, line 11: I'm not sure that it's a good idea to add such common depends

src/compat/posix/include/pthread.h
```diff
@@ -18,7 +18,7 @@
  * the <time.h> header.
  */
 #include <sched.h>
-
+#include <semaphore.h>
```
anton-bondarev, line 21: Does it need here? As I see it use only in pthread barriers module

sobhaniiest:
Hi All,
Sorry for the late response. I'll look into this and get back to you soon.
Thanks.

anton-bondarev:
The changes were merged in #2456

PR #1879, opened 2020-03-17
driver: clock: Add driver support for SP804

### LINKED ISSUE #1769: Implementing sp804 driver
sp804 is a timer module which is used in various ARM boards.
Technical reference manual can be found [here](http://infocenter.arm.com/help/topic/com.arm.doc.ddi0271d/DDI0271.pdf).

It's emulated for qemu vexpress machine, which is already supported (`arm/vexpress-a9` template), so it's required to implement a driver for this timer.

```
(qemu) info mtree
    ...
    0000000010011000-0000000010011fff (prio 0, i/o): sp804
    0000000010012000-0000000010012fff (prio 0, i/o): sp804
```

layman-n-ish:
This adds driver support for ARM Dual-Timer Module (SP804)
which is emulated in qemu-vexpress-a9.

This closes #1769.

Signed-off-by: Nishant Malpani <nish.malpani25@gmail.com>

### DIFF
--- a/src/drivers/clock/sp804_timer/sp804_timer.c
+++ b/src/drivers/clock/sp804_timer/sp804_timer.c
@@ -29,14 +29,14 @@
 #define SP804_T1_MIS    (SP804_BASE + 0x14) /* Timer1 Masked Interrupt Status Register */
 #define SP804_T1_BGL    (SP804_BASE + 0x18) /* Timer1 Background Load Register */
 
-#define SP804_CR_MASK           (0b11101111) /* Mask for bits 0-7 */
-#define SP804_CR_ONESHOT        (0 << 0) /* OneShot bit; 0 -> wrapping mode, 1 -> one-shot mode */
-#define SP804_CR_TIMERSIZE      (0 << 1) /* TimerSize bit; 0 -> 16-bit counter, 1 -> 32-bit counter */
-#define SP804_CR_TIMERPRE       (0 << 2) /* Prescale bits; 0 -> div by 1, 1 -> div by 16, 2 -> div by 256 */
-#define SP804_CR_INTEN          (1 << 5) /* InterruptEnable bit */
-#define SP804_CR_TIMERMODE      (0 << 6) /* TimerMode bit; 0 -> free-running mode, 1 -> periodic mode */
-#define SP804_CR_TIMEREN        (1 << 7) /* TimerEnable bit */
-#define SP804_ICR_CLEAR         (1 << 0) /* InterruptClear bit */
+#define SP804_CR_MASK       (0b11101111) /* Mask for bits 0-7 */
+#define SP804_CR_WRAPPING   (0 << 0) /* OneShot bit */
+#define SP804_CR_32BIT      (1 << 1) /* TimerSize bit */
+#define SP804_CR_TIMERPRE   (0 << 2) /* Prescale bits */
+#define SP804_CR_INTEN      (1 << 5) /* InterruptEnable bit */
+#define SP804_CR_PERIODIC   (1 << 6) /* TimerMode bit */
+#define SP804_CR_TIMEREN    (1 << 7) /* TimerEnable bit */
+#define SP804_ICR_CLEAR     (1 << 0) /* InterruptClear bit */
 
 #define SP804_PRESCALE          1 /* Prescaler operation - divide the rate of master clock */
 #define SP804_LR_VAL            ((SP804_TIMCLK) / (SP804_PRESCALE * SP804_TARGET_HZ))
@@ -61,15 +61,14 @@ static int sp804_init(void) {
 }
 
 static int sp804_config(struct time_dev_conf * conf) {
-    REG32_CLEAR(SP804_T1_CR, SP804_CR_MASK); /* Unset bits 0-7, also disables the timer to proceed with new config settings */
-
-    REG32_ORIN(SP804_T1_CR, (SP804_CR_ONESHOT | SP804_CR_TIMERMODE)); /* Select wrapping, free-running mode */
-    REG32_ORIN(SP804_T1_CR, SP804_CR_TIMERSIZE); /* Select 32-bit counter mode */
-    REG32_ORIN(SP804_T1_CR, SP804_CR_TIMERPRE); /* Set prescaler rate */
-    REG32_ORIN(SP804_T1_CR, SP804_CR_INTEN); /* Enable interrupts */
+    /* Unset bits 0-7, also disables the timer to proceed with new config settings */
+    REG32_CLEAR(SP804_T1_CR, SP804_CR_MASK);
 
     REG32_STORE(SP804_T1_LR, SP804_LR_VAL); /* Don't care in free-running mode */
-    REG32_ORIN(SP804_T1_CR, SP804_CR_TIMEREN); /* Enable timer */
+
+    /* setup Timer 1 as 32-bit interruptable periodic clocksource */
+    REG32_STORE(SP804_T1_CR, SP804_CR_32BIT | SP804_CR_INTEN | 
+                SP804_CR_PERIODIC | SP804_CR_TIMEREN);
 
     return 0;
 }
--- a/src/drivers/clock/sp804_timer/sp804_timer.c
+++ b/src/drivers/clock/sp804_timer/sp804_timer.c
@@ -1,0 +1,104 @@
+/**
+ * @file sp804_timer.c
+ * @brief ARM Dual-Timer (SP804) Module driver
+ * @author Nishant Malpani <nish.malpani25@gmail.com>
+ * @date 2020-03-07
+ */
+
+#include <drivers/common/memory.h>
+#include <hal/clock.h>
+#include <hal/reg.h>
+#include <kernel/irq.h>
+#include <kernel/printk.h>
+#include <kernel/time/clock_source.h>
+#include <kernel/time/time_device.h>
+
+#include <framework/mod/options.h>
+#include <embox/unit.h>
+
+#define SP804_BASE      OPTION_GET(NUMBER, base_addr) /* Base address */
+#define SP804_IRQ       OPTION_GET(NUMBER, irq_nr) /* Interrupt Request number */
+#define SP804_TARGET_HZ OPTION_GET(NUMBER, target_freq) /* Frequency to run the timer at */
+#define SP804_TIMCLK    100000000 /* Frequency of the SP804 clock timer - 100 MHz */
+
+#define SP804_T1_LR     (SP804_BASE + 0x00) /* Timer1 Load Register */
+#define SP804_T1_VR     (SP804_BASE + 0x04) /* Timer1 Current Value Register */
+#define SP804_T1_CR     (SP804_BASE + 0x08) /* Timer1 Control Register */
+#define SP804_T1_ICR    (SP804_BASE + 0x0C) /* Timer1 Interrupt Clear Register */
+#define SP804_T1_RIS    (SP804_BASE + 0x10) /* Timer1 Raw Interrupt Status Register */
+#define SP804_T1_MIS    (SP804_BASE + 0x14) /* Timer1 Masked Interrupt Status Register */
+#define SP804_T1_BGL    (SP804_BASE + 0x18) /* Timer1 Background Load Register */
+
+#define SP804_CR_MASK           (0b11101111) /* Mask for bits 0-7 */
+#define SP804_CR_ONESHOT        (0 << 0) /* OneShot bit; 0 -> wrapping mode, 1 -> one-shot mode */
+#define SP804_CR_TIMERSIZE      (0 << 1) /* TimerSize bit; 0 -> 16-bit counter, 1 -> 32-bit counter */
+#define SP804_CR_TIMERPRE       (0 << 2) /* Prescale bits; 0 -> div by 1, 1 -> div by 16, 2 -> div by 256 */
+#define SP804_CR_INTEN          (1 << 5) /* InterruptEnable bit */
+#define SP804_CR_TIMERMODE      (0 << 6) /* TimerMode bit; 0 -> free-running mode, 1 -> periodic mode */
+#define SP804_CR_TIMEREN        (1 << 7) /* TimerEnable bit */
+#define SP804_ICR_CLEAR         (1 << 0) /* InterruptClear bit */
+
+#define SP804_PRESCALE          1 /* Prescaler operation - divide the rate of master clock */
+#define SP804_LR_VAL            ((SP804_TIMCLK) / (SP804_PRESCALE * SP804_TARGET_HZ))
+
+static struct clock_source sp804_clock_source;
+static irq_return_t clock_handler(unsigned int irq_nr, void *data) {
+    clock_tick_handler(irq_nr, data);
+
+    REG32_STORE(SP804_T1_ICR, SP804_ICR_CLEAR); /* Clear interrupts */
+    REG32_STORE(SP804_T1_LR, SP804_LR_VAL);
+    return IRQ_HANDLED;
+}
+
+static int sp804_init(void) {
+    clock_source_register(&sp804_clock_source);
+
+    return irq_attach(SP804_IRQ,
+                      clock_handler,
+                      0,
+                      &sp804_clock_source,
+                      "SP804");
+}
+
+static int sp804_config(struct time_dev_conf * conf) {
+    REG32_CLEAR(SP804_T1_CR, SP804_CR_MASK); /* Unset bits 0-7, also disables the timer to proceed with new config settings */
+
+    REG32_ORIN(SP804_T1_CR, (SP804_CR_ONESHOT | SP804_CR_TIMERMODE)); /* Select wrapping, free-running mode */
+    REG32_ORIN(SP804_T1_CR, SP804_CR_TIMERSIZE); /* Select 32-bit counter mode */
+    REG32_ORIN(SP804_T1_CR, SP804_CR_TIMERPRE); /* Set prescaler rate */
+    REG32_ORIN(SP804_T1_CR, SP804_CR_INTEN); /* Enable interrupts */
+
+    REG32_STORE(SP804_T1_LR, SP804_LR_VAL); /* Don't care in free-running mode */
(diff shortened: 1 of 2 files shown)


src/drivers/clock/sp804_timer/sp804_timer.c
```diff
@@ -0,0 +1,104 @@
...
+                      "SP804");
+}
+
+static int sp804_config(struct time_dev_conf * conf) {
+    REG32_STORE(SP804_T1_CR, REG32_LOAD(SP804_T1_CR) & (~SP804_CR_MASK)); /* Unset bits 0-7, also disables the timer to proceed with new config settings */
+
+    REG32_STORE(SP804_T1_CR, REG32_LOAD(SP804_T1_CR) | SP804_CR_ONESHOT | SP804_CR_TIMERMODE); /* Select wrapping, free-running mode */
+    REG32_STORE(SP804_T1_CR, REG32_LOAD(SP804_T1_CR) | SP804_CR_TIMERSIZE); /* Select 32-bit counter mode */
+    REG32_STORE(SP804_T1_CR, REG32_LOAD(SP804_T1_CR) | SP804_CR_TIMERPRE); /* Set prescaler rate */
+    REG32_STORE(SP804_T1_CR, REG32_LOAD(SP804_T1_CR) | SP804_CR_INTEN); /* Enable interrupts */
```
Deryugin: You can use `REG32_ORIN()` macro to set these flags, this way it looks more readable.

layman-n-ish:
The disabling of the timer due to delta zero warning is solved; the problem was bad initialisation of the control registers. 

Although the `scripts/qemu/auto_qemu` script still stalls at `embox.driver.cache.pl310`. After using gdb for debugging, I noticed the stall happens at [this](https://github.com/embox/embox/blob/master/src/drivers/cache/pl310/pl310.c#L79) line in pl301.c. I'm investigating further by adding breakpoints in the aforementioned routine; I'll update here!

src/drivers/clock/sp804_timer/sp804_timer.c
```diff
@@ -0,0 +1,104 @@
...
+
+    return irq_attach(SP804_IRQ,
+                      clock_handler,
+                      0,
+                      &sp804_clock_source,
+                      "SP804");
+}
+
+static int sp804_config(struct time_dev_conf * conf) {
+    REG32_STORE(SP804_T1_CR, REG32_LOAD(SP804_T1_CR) & (~SP804_CR_MASK)); /* Unset bits 0-7, also disables the timer to proceed with new config settings */
```
Deryugin: You can use `REG32_CLEAR()` macro to clear mask.

layman-n-ish: You're right! Thanks.

src/drivers/clock/sp804_timer/sp804_timer.c
layman-n-ish: That's handy! Thanks.

Deryugin:
Do you receive any interrupts from this timer? If no, then probably that's the exact problem.

layman-n-ish:
> Do you receive any interrupts from this timer? If no, then probably that's the exact problem.

Exactly, my hypothesis was similar but going down the gdb road led me to believe that the stall happens [here](https://github.com/embox/embox/blob/master/src/kernel/sched/sched.c#L402), in sched.c! It's taking me a while to understand how `ksleep()` works in multi-threaded scenarios with preemptive scheduling (as it does here). Furthermore, I'm trying to understand what `process()` function does while scheduling. The corresponding comments for `process()` say:

> Process function is called in the function schedule() after extracting the next schedee from the runq. This function performs all necessary actions with a specific schedee implementation. It has to enable ipl as soon as possible. Returns schedee to which context switched.

What does it mean to '_enable ipl_' here?

**Auxiliary information which might be useful for your investigation:** 
The debug output from [here](https://github.com/embox/embox/blob/master/src/kernel/sched/sched.c#L399) is:
`debug: __schedule: prev: 0x621bc000, next: 0x621e0800`

where,
`debug: sched_start_init: boot_thread = 0x621bc00` and
`debug: __sched_wakeup: schedee 0x621e0800`.

layman-n-ish:
> Do you receive any interrupts from this timer? If no, then probably that's the exact problem.

Should I rather try timer and IRQ tests available [here](https://github.com/embox/embox/tree/master/src/tests/kernel) first?

alexkalmuk:
> > Do you receive any interrupts from this timer? If no, then probably that's the exact problem.
> 
> Should I rather try timer and IRQ tests available [here](https://github.com/embox/embox/tree/master/src/tests/kernel) first?

You can add printk() in your clock_handler() to check if it called, for example.

layman-n-ish:
> Do you receive any interrupts from this timer? If no, then probably that's the exact problem.

You were partially right; interrupts do get generated (by reading `SP804_T1_MIS` - the Masked Interrupt Status Register) when the timer zeroes down but the control is not passed down to my `clock_handler()` which is supposed to be my ISR. I verified this by checking if my `clock_handler()` is called or not, as suggested by @alexkalmuk. This behaviour is despite receiving `ENOERR` from the `irq_attach()` routine which registers the ISR for that particular IRQ!

Deryugin:
I have a following suggestion: IRQ number is incorrect. What `irq_attach()` does is just registering handler for interrupt, so when it's happening, this handler will be called.

There is no way to check that given IRQ number is actually related to this peripheral device. So if you attach clock handler to keyboard IRQ, then this handler will be called on key press, which is not related to clock at all. OS can't check it.

layman-n-ish:
> I have a following suggestion: IRQ number is incorrect.

You were absolutely right! I'm a bit embarrassed I missed this previously, but well, here we are with the driver now implemented :)

> There is no way to check that given IRQ number is actually related to this peripheral device. So if you attach clock handler to keyboard IRQ, then this handler will be called on key press, which is not related to clock at all. OS can't check it.

Thanks @Deryugin for this reasoning. It makes complete sense and certainly taught me a bit more.

layman-n-ish:
The driver for SP804 timer has been tested with the available timer tests on `arm-vexpress-a9`. I've squashed all commits into one.

anton-bondarev:
@layman-n-ish 
I tested it on arm/vexpress-a9 template. And it does not work.
For testing you should comment or delete
`@Runlevel(0) include embox.driver.clock.cortexa9(periph_base_addr=0x1e000000,irq_num=29)`
and add `@Runlevel(0) include embox.driver.clock.sp804`

layman-n-ish:
> @layman-n-ish
> I tested it on arm/vexpress-a9 template. And it does not work.
> For testing you should comment or delete
> `@Runlevel(0) include embox.driver.clock.cortexa9(periph_base_addr=0x1e000000,irq_num=29)`
> and add `@Runlevel(0) include embox.driver.clock.sp804`

Yep, I did exactly the same previously and all the timer tests ran flawlessly. Nevertheless, I shall test again and update here. Thanks for reviewing.

layman-n-ish:
Turns out the timer in 32-bit counter mode causes the problem. I had tested on the 16-bit mode and it works fine (see attached image below).
Screenshot from 2020-03-31 18-14-11

I have modified the code accordingly. You can test it now. Thanks!

I'll update here if I understand why 32-bit counter mode causes an issue.

src/drivers/clock/sp804_timer/sp804_timer.c
```diff
@@ -0,0 +1,104 @@
...
+#include <kernel/printk.h>
+#include <kernel/time/clock_source.h>
+#include <kernel/time/time_device.h>
+
+#include <framework/mod/options.h>
+#include <embox/unit.h>
+
+#define SP804_BASE      OPTION_GET(NUMBER, base_addr) /* Base address */
+#define SP804_IRQ       OPTION_GET(NUMBER, irq_nr) /* Interrupt Request number */
+#define SP804_TARGET_HZ OPTION_GET(NUMBER, target_freq) /* Frequency to run the timer at */
```
anton-bondarev, line 21: I think you should use SYS_CLOCK instead. It's defined in hal./system.h
and setup in config as `include embox.arch.system(core_freq=48054841)`

src/drivers/clock/sp804_timer/sp804_timer.c
```diff
@@ -0,0 +1,104 @@
...
+#define SP804_T1_LR     (SP804_BASE + 0x00) /* Timer1 Load Register */
+#define SP804_T1_VR     (SP804_BASE + 0x04) /* Timer1 Current Value Register */
+#define SP804_T1_CR     (SP804_BASE + 0x08) /* Timer1 Control Register */
+#define SP804_T1_ICR    (SP804_BASE + 0x0C) /* Timer1 Interrupt Clear Register */
+#define SP804_T1_RIS    (SP804_BASE + 0x10) /* Timer1 Raw Interrupt Status Register */
+#define SP804_T1_MIS    (SP804_BASE + 0x14) /* Timer1 Masked Interrupt Status Register */
+#define SP804_T1_BGL    (SP804_BASE + 0x18) /* Timer1 Background Load Register */
+
+#define SP804_CR_MASK           (0b11101111) /* Mask for bits 0-7 */
+#define SP804_CR_ONESHOT        (0 << 0) /* OneShot bit; 0 -> wrapping mode, 1 -> one-shot mode */
```
anton-bondarev, line 33: What doe it mean?
Usually only (1 << bit) number using

src/drivers/clock/sp804_timer/sp804_timer.c
```diff
@@ -0,0 +1,104 @@
...
+
+#define SP804_T1_LR     (SP804_BASE + 0x00) /* Timer1 Load Register */
+#define SP804_T1_VR     (SP804_BASE + 0x04) /* Timer1 Current Value Register */
+#define SP804_T1_CR     (SP804_BASE + 0x08) /* Timer1 Control Register */
+#define SP804_T1_ICR    (SP804_BASE + 0x0C) /* Timer1 Interrupt Clear Register */
+#define SP804_T1_RIS    (SP804_BASE + 0x10) /* Timer1 Raw Interrupt Status Register */
+#define SP804_T1_MIS    (SP804_BASE + 0x14) /* Timer1 Masked Interrupt Status Register */
+#define SP804_T1_BGL    (SP804_BASE + 0x18) /* Timer1 Background Load Register */
+
+#define SP804_CR_MASK           (0b11101111) /* Mask for bits 0-7 */
```
anton-bondarev, line 32: What does it mean?

src/drivers/clock/sp804_timer/sp804_timer.c
```diff
@@ -0,0 +1,104 @@
...
+#define SP804_CR_TIMERMODE      (0 << 6) /* TimerMode bit; 0 -> free-running mode, 1 -> periodic mode */
+#define SP804_CR_TIMEREN        (1 << 7) /* TimerEnable bit */
+#define SP804_ICR_CLEAR         (1 << 0) /* InterruptClear bit */
+
+#define SP804_PRESCALE          1 /* Prescaler operation - divide the rate of master clock */
+#define SP804_LR_VAL            ((SP804_TIMCLK) / (SP804_PRESCALE * SP804_TARGET_HZ))
+
+static struct clock_source sp804_clock_source;
+static irq_return_t clock_handler(unsigned int irq_nr, void *data) {
+    clock_tick_handler(irq_nr, data);
```
anton-bondarev, line 46: maybe call clock_tick_handler() after clear interrupt and load a new value?

src/drivers/clock/sp804_timer/sp804_timer.c
```diff
@@ -0,0 +1,104 @@
...
+                      clock_handler,
+                      0,
+                      &sp804_clock_source,
+                      "SP804");
+}
+
+static int sp804_config(struct time_dev_conf * conf) {
+    REG32_CLEAR(SP804_T1_CR, SP804_CR_MASK); /* Unset bits 0-7, also disables the timer to proceed with new config settings */
+
+    REG32_ORIN(SP804_T1_CR, (SP804_CR_ONESHOT | SP804_CR_TIMERMODE)); /* Select wrapping, free-running mode */
```
anton-bondarev, line 66: I think its more clear if use direct register access
I mean 
REG32_STORE(SP804_T1_CR, 0)
REG32_STORE(SP804_T1_CR, (SP804_CR_ONESHOT | SP804_CR_TIMERMODE | SP804_CR_TIMERSIZE | ...)
https://github.com/torvalds/linux/blob/master/drivers/clocksource/timer-sp804.c

src/drivers/clock/sp804_timer/sp804_timer.c
layman-n-ish: You mean set the target frequency of the timer same as the core's frequency? I thought all these timers are external and operate at a lower frequency than the master clock. I've used other drivers implemented as a reference [1] [2].

[1] https://github.com/embox/embox/blob/master/src/drivers/clock/cortexa9_timer/cortexa9_timer.c#L40
[2] https://github.com/embox/embox/blob/master/src/drivers/clock/epit/epit.c#L22

src/drivers/clock/sp804_timer/sp804_timer.c
layman-n-ish: Here I mean to set the 0th bit with the value '0' - to choose the wrapping mode rather than the one-shot mode. I could have done it the way you suggested and done a 'NOT' when loading the appropriate register, although I thought this provides more readability to the driver's consumer (with the corresponding comments) while setting the timer parameters. 

I shall change if you think your way is more suitable.

src/drivers/clock/sp804_timer/sp804_timer.c
layman-n-ish: I use this mask to clear the Control Register during configuration. The bit at index '4' is reserved and should not be modified, as communicated in the manual's 3.2.3 section [1], hence the '0' at that position.

[1] http://infocenter.arm.com/help/topic/com.arm.doc.ddi0271d/DDI0271.pdf

src/drivers/clock/sp804_timer/sp804_timer.c
layman-n-ish: You mean to set the value of the Control Register in one go rather than loading it multiple times and storing it multiple times? You still want me to use the REG32_ORIN macro, right? Because I'm clearing the Control Register in L64 anyways.

src/drivers/clock/sp804_timer/sp804_timer.c
layman-n-ish: I cannot think of any issues the way you've suggested may create for now, but here I've gone along with what other drivers have implemented [1] [2]. You want me to go ahead and still change it and call `clock_tick_handler()` in the end? 

[1] https://github.com/embox/embox/blob/master/src/drivers/clock/cortexa9_timer/cortexa9_timer.c#L48
[2] https://github.com/embox/embox/blob/master/src/drivers/clock/epit/epit.c#L44

layman-n-ish:
Just leaving a comment here to remind you to resolve some questions asked; no hurry though :)
Thanks!

anton-bondarev:
@layman-n-ish I already left a few comments earlier. In fact, I've checked your implementation and found that it does not work correctly. You also can check it if you run 'ticker' command (embox.cmd.testing.ticker) you will see that timer frequency is slower than need, and it is impossible to achieve the correct frequency of the timer. After that, I tried to research the problem. 
Let's try to figure it out together, but for it, You should make the code more understandable.
Start please with Control Register. I think it has not to be ONE_SHORT timer and has 32 bits access

layman-n-ish:
> @layman-n-ish I already left a few comments earlier. 

Yes, I've replied to those inline a fortnight back. Could you please have a look at those? Thanks.

> Let's try to figure it out together, but for it, You should make the code more understandable.
Start please with Control Register. I think it has not to be ONE_SHORT timer and has 32 bits access

Alright, for the Control Register I've to use direct register access w/ `REG32_STORE` as you mentioned in the review, right?

src/drivers/clock/sp804_timer/sp804_timer.c
anton-bondarev: I think code in Linux driver is cleaner
```
	/* setup timer 0 as free-running clocksource */
	writel(0, base + TIMER_CTRL);
	writel(0xffffffff, base + TIMER_LOAD);
	writel(0xffffffff, base + TIMER_VALUE);
	writel(TIMER_CTRL_32BIT | TIMER_CTRL_ENABLE | TIMER_CTRL_PERIODIC,
		base + TIMER_CTRL);
```
I suggest somthing like this

```
REG32_STORE(SP804_T1_CR, TIMER_CTRL_32BIT | TIMER_CTRL_ENABLE | TIMER_CTRL_PERIODIC);
```

layman-n-ish:
> Let's try to figure it out together, but for it, You should make the code more understandable.
> Start please with Control Register. I think it has not to be ONE_SHORT timer and has 32 bits access

Sorry for the late reply. I've tried to make the code look cleaner while configuring the control register. Let me know if you require any other changes.

The timer tests do pass although as you previously pointed out the timer runs awfully slow as seen in while executing the 'ticker' command. I wasn't able to resolve this issue. Any help would be appreciated :)

layman-n-ish:
> > Let's try to figure it out together, but for it, You should make the code more understandable.
> > Start please with Control Register. I think it has not to be ONE_SHORT timer and has 32 bits access
> 
> Sorry for the late reply. I've tried to make the code look cleaner while configuring the control register. Let me know if you require any other changes.
> 
> The timer tests do pass although as you previously pointed out the timer runs awfully slow as seen in while executing the 'ticker' command. I wasn't able to resolve this issue. Any help would be appreciated :)

anton-bondarev:
@layman-n-ish Thanks, looks much cleanly!

Now I was able to set up the right frequency!

Timer frequency depends on the value which is loaded in Timer1Load (from the manual).
You load 
```
#define SP804_PRESCALE          1 /* Prescaler operation - divide the rate of master clock */
#define SP804_LR_VAL            ((SP804_TIMCLK) / (SP804_PRESCALE * SP804_TARGET_HZ))
```
Try to define SP804_LR_VAL in different value (2500) for example and look at results

anton-bondarev:
These changes were merged in #2360

Issue #1884, opened 2020-03-19
usermode86/debug often fails on CIs


anton-bondarev:
Checking usermode86/debug was disable on github actions and travis because it fails too often

alexkalmuk:
Example - https://github.com/embox/embox/runs/518891252#step:10:81

PR #1885, opened 2020-03-19
Disable usermode86/debug on ci

anton-bondarev:
Issue: https://github.com/embox/embox/issues/1884


alexkalmuk:
Issue: https://github.com/embox/embox/issues/1884

PR #1890, opened 2020-03-20
Run on Rpi1 model b rev2.0

alexkalmuk:
* Improve PL011 serial driver to program baud rate, FIFO, etc.
* Remove raspi_uart serial driver because it's not raspi uart actually, but usual PL011.
* Update dcache_flush/inval to work on ARMv6 (it doesn't have cache line register like ARMv7). It's more like workaround now.

Wiki: https://github.com/embox/embox/wiki/Running-Embox-on-Raspberry-Pi#running-on-raspberry-pi-model-b-rev-20-rpi1


src/arch/arm/mmu/cache.c
```diff
@@ -8,14 +8,40 @@
...
+	/*
+	 * I am not sure about if it is arch defined cache line size,
+	 * but Linux uses 32 bit cache line size for both I and D caches
+	 * for any ARMV6: arch/arm/mm/cache-v6.S
+	 *
+	 * For ARMv5 I am not sure at all, just add this to fix arm/qemu.
+	 *
+	 * P.S. It is added because of ARM1176JZF-S (Ra Pi model B rev 2.0)
+	 */
+	return 32;
```
Deryugin: Is it 32 bits or 32 bytes? `line_size` seems to be inerpreted as bytes count in `dcache_inval()`, but in comments it's `32 bit cache line size`

alexkalmuk: 32 bytes, it's a typo. Thank you.

alexkalmuk: Fixed.

alexkalmuk:
I created an issue about buggy serial input - https://github.com/embox/embox/issues/1891

PR #1894, opened 2020-03-21
1840 memccpy implementation

### LINKED ISSUE #1840: Implement memccpy()
`memccpy()` is a part of POSIX standard.

For more details refer to POSIX specification: https://pubs.opengroup.org/onlinepubs/9699919799/

 Also implement some tests as well to make sure it works correctly.

tech-gian:
closes #1840 (Good First issue)
After correcting my mistake in the first PR, I made this try and now I hope everything is ok.
I have fixed memccpy_test.c, renamed memccpy.c, so everyone can use it, and I think I have connected these with Mybuild correctly.
I ran embox emulator and it didn't get me any error. Please let me know if something is wrong or need to be fixed.

Thank you for your time

### DIFF
--- a/src/compat/libc/include/string.h
+++ b/src/compat/libc/include/string.h
@@ -261,6 +261,30 @@ extern char *strcasestr(const char *haystack, const char *needle);
 extern char *strerror(int err);
 
 /**
+ * Copies no more than @a n bytes from memory area @a src
+ * to memory area @a dest,
+ * stopping when the character @a c is found.
+ * 
+ * @param dest
+ *  Destination buffer
+ * @param src
+ *  The object in memory being copied
+ * @param n
+ *  The number of bytes to copy
+ * @param c
+ *  Character to stop copying when found.
+ * @return
+ *  Pointer to the destination buffer.
+ * 
+ * @note
+ *  If the memory areas overlap, the results are undefined.
+ * @see memcpy()
+ *  It does the same thing, but it doesn't stop at c.
+ */
+extern void* memccpy(void* dest, const void* src, int c, size_t n);
+
+
+/**
  * Copies @a n bytes from @a src to @a dst which must be two non-overlapping
  * buffers.
  *
--- a/src/compat/libc/string/memccpy.c
+++ b/src/compat/libc/string/memccpy.c
@@ -1,0 +1,48 @@
+/**
+ *  @file
+ *  @brief Implementation of #memccpy() function.
+ *
+ *  @date 18/3/20
+ *  @author Giannis Zapantis (github name: tech-gian)
+*/
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+
+void* memccpy(void* dest, const void* src, int c, size_t n) {
+
+/* If overlap happens the results
+ * will be undefined.
+ */
+
+/* Getting dest and srt into byte
+ * format.
+ */
+    char* temp_dest;
+    char* temp_src;
+    
+    temp_dest = (char*) dest;
+    temp_src = (char*) src;
+
+    size_t i;
+
+
+/* copying the value from src
+ * to dest until c found.
+ */
+    for (i = 0 ; i < n ; i++) {
+        if (temp_src[i] == c) {
+            return &temp_dest[i];
+        }
+        temp_dest[i] = temp_src[i];
+    }
+
+    return NULL;
+
+/* return NULL if c wasn't found in src.
+ *
+ * else the next pointer of dest.
+ */
+}
--- a/src/tests/posix/memccpy_test.c
+++ b/src/tests/posix/memccpy_test.c
@@ -1,0 +1,176 @@
+/**
+ * @file
+ * @brief Tests for memccpy function
+ * 
+ * @author Giannis Zapantis  (github name: tech-gian)
+ * @date 18/3/2020
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include <embox/test.h>
+
+
+EMBOX_TEST_SUITE("memccpy");
+
+
+/* TEST CASES for testing memccpy
+ * on strings
+ */
+
+
+TEST_CASE("empty static allocated string with character off the string") {
+    
+    char original[] = "To be copied";
+    char tested[128];
+
+    void* test;
+
+    test = memccpy(tested, original, 'w', strlen(original)+1);
+    test_assert_zero(strcmp(tested, original));
+    test_assert_null(test);
+}
+
+
+TEST_CASE("non-empty static allocated string with character off the string") {
+
+    char original[] = "To be copied";
+    char tested[] = "Before copy";
+
+    void* test;
+
+    test = memccpy(tested, original, 'w', strlen(original)+1);
+    test_assert_zero(strcmp(tested, original)); 
+    test_assert_null(test);
+}
+
+
+TEST_CASE("non-empty static allocated string with character on the string") {
+
+    char original[] = "To be copied";
+    char tested[] = "Before copy";
+
+    void* test;
+
+    test = memccpy(tested, original, 'c', strlen(original)+1);
+    test_assert_zero(strcmp(tested, "To be  copy"));
+    test_assert_not_null(test);
+}
+
+
+TEST_CASE("empty static allocated string with character on the string") {
+
+    char original[] = "To be copied";
+    char tested[128];
+
+    void* test;
+
+    test = memccpy(tested, original, 'c', strlen(original)+1);
+    test_assert_zero(strcmp(tested, "To be "));
+    test_assert_not_null(test);
+}
+
+
+TEST_CASE("dynamic allocated string with character off the string") {
+
+    char original[] = "To be copied";
+    char* tested;
+
+    tested = malloc(128*sizeof(char));
+
+    void* test;
+
+    test = memccpy(tested, original, 'w', strlen(original)+1);
+    test_assert_zero(strcmp(tested, original));
+    test_assert_null(test);
+
+    free(tested);
+}
+
+
+TEST_CASE("dynamic allocated string with character on the string") {
+
+    char original[] = "To be copied";
+    char* tested;
+    
+    tested = malloc(128*sizeof(char));
+
+    void* test;
+
+    test = memccpy(tested, original, 'c', strlen(original)+1);
+    test_assert_zero(strcmp(tested, "To be "));
+    test_assert_not_null(test);
+
+    free(tested);
+}
+
+
+/* TEST CASES for testing memccpy
+ * on ints
+ */
+
+
+TEST_CASE("empty static allocated array with number on array") {
+
+    int original[] = {1, 2, 3, 4};
+    int tested[4];
+    int stoping_c = 3;
+
+    void* test;
+    int i;
+
+    test = memccpy(tested, original, stoping_c, sizeof(original));
+    for (i = 0 ; i < sizeof(original) / sizeof(int) ; i++) {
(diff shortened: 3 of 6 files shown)


anton-bondarev:
@tech-gian Thanks. It looks better. But

-  there is weird formatting in memccpy.c
-  embox_memccpy() use in some test, but I did not found any implemetation of the routine
-  some odd variables name t6 for example
-  ..
-

tech-gian:
- I corrected embox_memccpy in some tests (forgotten by mistake).
- I think that I have just "over commented" the memccpy.c, but it's just to be fully defined.
- I have named them t6 etc, because they are just for testing (there comes the 't').
- I have forgotten to add the function in the .h (include) file, but I did. Except you mean something different that I missed.

In the way that now is this branch, it runs fine in the embox emulator. I don't know if I have written the Mybuild files correctly. If you can guide me, so I could built it right, I would be greatful! Thank you!

src/compat/libc/string/memccpy.c
```diff
@@ -0,0 +1,95 @@
...
+    size_t pos = n;
+
+    char* ar1[n];
+
+/* getting each byte for c. */
+    char temp_c[4];
+    temp_c[0] = (c >> 24) & 0xFF;
+    temp_c[1] = (c >> 16) & 0xFF;
+    temp_c[2] = (c >> 8) & 0xFF;
+    temp_c[3] = c & 0xFF;
```
alexkalmuk: Are you sure there are 4 chars?

```
The  memccpy()  function  copies  no more than n bytes from memory area
src to memory area dest, stopping when the character c is found.
```
So `c` is just a character, not 4 characters.

alexkalmuk: If we correct this the implementation could be much simpler.

src/tests/posix/memccpy_test.c
```diff
@@ -0,0 +1,174 @@
...
+    void* test = memccpy(t5, t1, 'c', strlen(t1)+1);
+    test_assert_zero(strcmp(t5, "To be  copy"));
+    test_assert_not_null(test);
+}
+
+
+TEST_CASE("empty static allocated string with character on the string") {
+
+    char t1[] = "To be copied";
+    char t6[128];
```
alexkalmuk: I agree with @anton-bondarev, the variables names are weird. In case if you want to use `t` then why `t1, t6` but not `t1, t2`?

src/compat/libc/string/memccpy.c
tech-gian: Below that I am checking only one byte (temp_c[3]), because as you said, it's only one char. But I think that because c is int, we have to check all 4 bytes, considering user's wrong input. The problem came up when I tested with ints instead of char* and then if I sent c=0, it gave me wrong results.

src/tests/posix/memccpy_test.c
tech-gian: You are right. I thought that it would be better to have an order for the dest variables and keep the same t1 for the src variable. What names do you suggest to put instead?

src/compat/libc/string/memccpy.c
```diff
@@ -0,0 +1,95 @@
...
+
+
+/* if c!=0 then we check the last byte.
+ * if there is in src.
+ */
+if (c!=0) {
```
anton-bondarev: What is the weird formatting
Use tabs at the start of the lines please

tech-gian: https://github.com/embox/embox/pull/1894/commits/1fb0ae2368e6c6fa61a094bdeec09d23563959a4

This is what you mean with tabs at the start of the lines?

anton-bondarev: No
I mean [this part of Embox code-style](https://github.com/embox/embox/wiki/Code-Style#tabs-and-brackets)

tech-gian: Oh, I'm really sorry. I didn't even noticed that. Thank you :) I corrected it. Is there anything else, that need to be changed? I want to do a very good work, even for this "small" function.

src/compat/libc/string/memccpy.c
```diff
@@ -0,0 +1,99 @@
...
+            }
+        }
+    }
+    else {
+/* if c==0 we check every byte to be
+ * the same (4 bytes of int).
+ */
+        for (size_t i = 0; i < n; i++) {
+            if (i < n - 4) {
+                if (temp_src[i] == temp_c[0] && temp_src[i+1] == temp_c[1] && temp_src[i+2] == temp_c[2] && temp_src[i+3] == temp_c[3]) {
```
anton-bondarev: what do you check here?

tech-gian: If c==0 then we have to check all 4 bytes of c (because it's int), to see if it really equals with an element of src. Else we only have to check byte to byte to see if it is the same. The problem came up when I tested this function with ints and send 0 as c (stopping character).

tech-gian:
So, do I have to correct something? Am I waiting for you to review the PR? (I'm asking because I saw that the previous PR closed).

anton-bondarev:
> So, do I have to correct something? Am I waiting for you to review the PR? (I'm asking because I saw that the previous PR closed).

Yes, I've closed the wrong PR.
This PR is rather dirty try clean it yourself.

tech-gian:
> Yes, I've closed the wrong PR.
> This PR is rather dirty try clean it yourself.

I cleaned up the code. Deleted some comments, that referenced to .h file and renamed properly the variables in the test file. I think everything's ready.

src/compat/libc/include/string.h
```diff
@@ -260,6 +260,30 @@ extern char *strcasestr(const char *haystack, const char *needle);
...
+ *  Character to stop copying when found.
+ * @return
+ *  Pointer to the destination buffer.
+ * 
+ * @note
+ *  If the memory areas overlap, the results are undefined.
+ * @see memcpy()
+ *  It does the same thing, but it doesn't stop at c.
+ */
+void* memccpy(void* dest, const void* src, int c, size_t n);
```
anton-bondarev: `extern void *memccpy(void *dest, const void *src, int c, size_t n);`

src/tests/posix/memccpy_test.c
```diff
@@ -0,0 +1,174 @@
...
+    void* test = memccpy(tested, original, 'c', strlen(original)+1);
+    test_assert_zero(strcmp(tested, "To be "));
+    test_assert_not_null(test);
+}
+
+
+TEST_CASE("dynamic allocated string with character off the string") {
+
+    char original[] = "To be copied";
+    char* tested = malloc(128*sizeof(char));
```
anton-bondarev: Where do you free memory

src/tests/posix/memccpy_test.c
```diff
@@ -0,0 +1,174 @@
...
+    char original[] = "To be copied";
+    char tested[128];
+
+    void* test = memccpy(tested, original, 'c', strlen(original)+1);
+    test_assert_zero(strcmp(tested, "To be "));
+    test_assert_not_null(test);
+}
+
+
+TEST_CASE("dynamic allocated string with character off the string") {
```
anton-bondarev: you can don't check dynamic memory allocation

anton-bondarev:
@tech-gian The memccpy is a little easy :)
[https://pubs.opengroup.org/onlinepubs/9699919799/functions/memccpy.html](https://pubs.opengroup.org/onlinepubs/9699919799/functions/memccpy.html)

`void *memccpy(void *s1, const void *s2, int c, size_t n);`

> 
> DESCRIPTION
> 
>     The memccpy() function shall copy bytes from memory area s2 into s1, stopping after the first occurrence of byte c (converted to an unsigned char) is copied

It's not required to convert `int c` to bytes array. It's just unsigned char

If it's necessary you can use memcpy() instead of you ` if (temp_src[i] == temp_c[0] && temp_src[i+1] == temp_c[1] && temp_src[i+2] == temp_c[2] && temp_src[i+3] == temp_c[3])`

src/compat/libc/include/string.h
tech-gian: Added it. Forgotten by mistake

src/tests/posix/memccpy_test.c
tech-gian: I add free for all the pointers, that I called with malloc.

src/tests/posix/memccpy_test.c
tech-gian: I just thought it would be good, if I check with pointers from malloc too. In the previous PR, someone commented that was a good add.

tech-gian:
@anton-bondarev I corrected memccpy implementation. I had understand it wrong about how the function behaves when c=0. I fixed it and made it much simpler. I also removed tests about ints with c=0, because they will stop from the first byte of the array.

anton-bondarev:
@tech-gian Thanks. it's better but I think it can be written easier yet. :)

And care about code style
```
    void* test = memccpy(tested, original, 3, sizeof(original));
    for (int i=0 ; i<sizeof(original)/sizeof(int) ; i++) {
        if  (original[i]==3) break;
```

must be like
```

    void* test;
    int i;

    test = memccpy(tested, original, 3, sizeof(original));
    for (i=0 ; i < sizeof(original) / sizeof(int) ;  i++) {
        if  (original[i] == 3) {
              break;
       }
```

I don't now what is 3 and other things, but it is too different to read bad formatting code

tech-gian:
Thank you for the feedback:) I fixed the code style's issues and I named 3 as stoping_c, for c that stops the memccpy() function.

src/compat/libc/string/memccpy.c
```diff
@@ -0,0 +1,67 @@
...
+ * to dest until c found.
+ */
+    for (i = 0 ; i < n ; i++) {
+        if (i < pos) {
+            temp_dest[i] = *p_src[i];
+        }
+        else {
+            return dest+i;
+        }
+    }
```
alexkalmuk: Why do you need these 3 loops? You can iterate char by char and copy `n` chars from `src` to `dest` only if the current char is not `c`. And if the current char is `c` you just exit from the loop.

tech-gian: I don't need 3 loops, you are right. I replaced it with one loop doing exactly what you are telling me.

src/compat/libc/string/memccpy.c
```diff
@@ -0,0 +1,50 @@
...
+    temp_dest = (char*) dest;
+    temp_src = (char*) src;
+
+    size_t i;
+
+
+/* copying the value from src
+ * to dest until c found.
+ */
+    for (i = 0 ; i < n ; i++) {
```
anton-bondarev:
```

        if (temp_src[i] != c) {
            temp_dest[i] = temp_src[i];
        }
        else {
            return dest+i;
        }
```
Maybe follow is more clear
```
        if (temp_src[i] == c) {
            return &temp_dest[i];
        }
```

src/compat/libc/string/memccpy.c
alexkalmuk: Ok, now it looks better.

src/compat/libc/string/memccpy.c
tech-gian: I will agree with you. I fixed it.

anton-bondarev:
@tech-gian Thanks
I've approved. Please rebase under master and squash commits. There should be three commits (implementation, tests, template modification)

tech-gian:
@anton-bondarev Thank you! I have squash the commits into the 3 that you mentioned above. I have also rebase under master (in my fork, tech-gian/embox). Is something else I should do? It's my first PR, as I said before, so I am not sure if I did everything correct.

PR #1901, opened 2020-03-25
Implement logmod command

Deryugin:
`logmod` command changes log level for given module in runtime.

### DIFF
--- a/src/cmds/module/logmod.c
+++ b/src/cmds/module/logmod.c
@@ -1,0 +1,74 @@
+/**
+ * @file logmod.c
+ * @brief
+ * @author Denis Deryugin <deryugin.denis@gmail.com>
+ * @version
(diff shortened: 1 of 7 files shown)


src/cmds/module/logmod.c
```diff
@@ -0,0 +1,74 @@
...
+			break;
+		default:
+			return -EINVAL;
+		}
+	}
+
+	mod_foreach(mod) {
+		if (!strstr(mod_name(mod), argv[1])) {
+			continue;
+		}
```
alexkalmuk: Why do not strcmp?

Also, Embox suffers from not very good namings like `core`, `all`, `arch`. At the same time, we do not forbid to have several modules with the same names in different packages, so formally it is ok. Maybe `package_name.mode_name` would be more deterministic?

Issue #1821, opened 2020-03-10
Add tests for pthread_cancel


anton-bondarev:
Add test cases for 
```pthread_cancel()
pthread_cleanup_push()
pthread_cleanup_pop()
...
```

sobhaniiest:
Hi, I am working on this issue and am trying with a simple test case to test `pthread_cancel`, the code snippet is the following: 

> static pthread_t tid;
static void *thread_fun(void *arg) {
    int r = pthread_cancel(tid);
    test_assert_equal(r, 0);
    return NULL;
}
TEST_CASE("Cancel self thread") {
    void *res;
    test_assert_zero(pthread_create(&tid, NULL, thread_fun, NULL));
    test_assert_zero(pthread_join(tid, &res));
}

The test falls in an infinite loop with the repetition of this:

> run 0x001599bf <boot_stub+0x0>                src/kernel/sched/boot_thread.c:20
EXCEPTION [0xe]: error = 00000000
EAX=a5a5a5a5    EBX=00258620 ECX=00000000 EDX=a5a5a5a5
 GS=00000010     FS=00100010  ES=001a0010  DS=00100010
EDI=00000000    ESI=0025a428 EBP=01b8e600 EIP=0010279d
 CS=00000008 EFLAGS=00010046 ESP=01b8e630  SS=00258620
 --   00001000 * A R    thread 0  task 1 --------------------------------------
 62 0x0015568c <exception_handler+0x125> ...h/x86/kernel/exception_handler.c:20
 61 0x00100079 <excep_stub+0x1c>       src/arch/x86/kernel/exception_entry.S:49
 60 0x0010279d <stack_iter_next+0x66>    src/arch/x86/lib/debug/stack_iter.c:40
 59 0x00101bc5 <do_traceback_dump_thread+0x5b>     src/lib/debug/whereami.c:135
 58 0x00101d6c <whereami+0x96>                     src/lib/debug/whereami.c:170
 57 0x0015568c <exception_handler+0x125> ...h/x86/kernel/exception_handler.c:20
 56 0x00100079 <excep_stub+0x1c>       src/arch/x86/kernel/exception_entry.S:49
 55 0x0010279d <stack_iter_next+0x66>    src/arch/x86/lib/debug/stack_iter.c:40
 54 0x00101bc5 <do_traceback_dump_thread+0x5b>     src/lib/debug/whereami.c:135
 53 0x00101d6c <whereami+0x96>                     src/lib/debug/whereami.c:170
[...]

PR #1906, opened 2020-03-27
Fixed pthread_cancel & added test

### LINKED ISSUE #1889: Fix pthread_cancel
- According to [POSIX standards](https://pubs.opengroup.org/onlinepubs/009695399/functions/pthread_setcancelstate.html): The cancelability state and type of any newly created threads, including the thread in which main() was first invoked, shall be `PTHREAD_CANCEL_ENABLE` and `PTHREAD_CANCEL_DEFERRED` respectively. So if we call `pthread_cancel` for a newly created thread it should be terminated and should return 0.
But here it invokes [this](https://github.com/embox/embox/blob/37bbb1e10d4d12e017015b2a8c7a70c34fc69c40/src/kernel/thread/thread_cancel_disable.h#L18) not [this](https://github.com/embox/embox/blob/37bbb1e10d4d12e017015b2a8c7a70c34fc69c40/src/kernel/thread/thread_cancel_enable.c#L15) and returns an error no and crashes the system.

- By adding `embox.kernel.thread.thread_cancel_enable` in `mods.config`, `pthread_cancel` invokes [this](https://github.com/embox/embox/blob/37bbb1e10d4d12e017015b2a8c7a70c34fc69c40/src/kernel/thread/thread_cancel_enable.c#L15) but still crashes the system when it is tested. Check issue #1821

sobhaniiest:
This will fix #1889 and #1821 

1. The error was occurring because of `thread_cancel_enable` was not default [here](https://github.com/embox/embox/blob/61921411738b0cfeb27d4b096337ca5e285bfc86/src/kernel/thread/Mybuild#L94) and in the `pthread_create`, the `*thread` was being initialized after the execution of thread [here](https://github.com/embox/embox/blob/61921411738b0cfeb27d4b096337ca5e285bfc86/src/compat/posix/pthread/pthread.c#L178).
`pthread_cancel` was failing as it was using `thread` as an argument which was not initialized.
2. I have modified `thread_cancel` for handling `PTHREAD_CANCEL_DISABLE` and calling `pthread_cancel` after the execution of the thread.
3. I have added tests for `pthread_cancel`, `pthread_cleanup_push` and `pthread_cleanup_pop`.

Please review this. Thanks :)
Signed-off-by: Sobhan Mondal <inro20fdm@gmail.com>


src/kernel/thread/thread_cancel_enable.c
```diff
@@ -13,6 +13,19 @@
...
+	}
+	if (PTHREAD_CANCEL_DISABLE == t->cleanups.state) {
+		while (1) {
+			if (PTHREAD_CANCEL_ENABLE == t->cleanups.state) {
+				break;
+			}
+			if (t->state == TS_EXITED) {
+				return 0;
+			}
+		}
```
alexkalmuk: What are you doing here? Are you waiting for the thread state to be exited? Did you investigate where this state will be changed?

alexkalmuk: You are waiting for PTHREAD_CANCEL_ENABLE, but where it will be set?

sobhaniiest: > What are you doing here? Are you waiting for the thread state to be exited? Did you investigate where this state will be changed?

Yes. It is waiting for either the state change or the completion of the thread. If the state changes from `PTHREAD_CANCEL_DISABLE` to  `PTHREAD_CANCEL_ENABLE` then thread cancelation can be done and if the thread executes without changing the state (`PTHREAD_CANCEL_DISABLE`) then we can not apply `pthread_cancel` on it.

> You are waiting for PTHREAD_CANCEL_ENABLE, but where it will be set?

I have added a test for this situation where a thread disables its cancelation then executes some task then again enables its cancelation. [Here](https://github.com/sobhaniiest/embox/blob/da704f750c7d0649c6779f53efe2f7fb5a8e4e3d/src/tests/posix/pthread/pthread_cancel_test.c#L83), please have a look.

I've got this example from [here](http://man7.org/linux/man-pages/man3/pthread_cancel.3.html).

sobhaniiest:
Hi @alexkalmuk 
For this patch `x86/qt-app` fails on CIs and shows :

> ./build/base/obj/embox-2.o: In function `QEmboxVC::~QEmboxVC()':
/home/inro/Desktop/embox_new/embox/build/extbld/third_party/qt/plugin/platform/emboxvc/../../../core/qt-everywhere-opensource-src-4.8.5/src/plugins/platforms/emboxvc/emboxvcwindowsurface.cpp:342: undefined reference to `thread_cancel(thread*)'

Though I have added `thread_cancel.h` in `thread.h` and have also tried by adding `@Runlevel(1) include embox.kernel.thread.thread_cancel` in `mods.config`. Could you please help me to understand where am I doing the mistake. Thanks :)

alexkalmuk:
> Hi @alexkalmuk
> For this patch `x86/qt-app` fails on CI and shows :
> 
> > ./build/base/obj/embox-2.o: In function `QEmboxVC::~QEmboxVC()': /home/inro/Desktop/embox_new/embox/build/extbld/third_party/qt/plugin/platform/emboxvc/../../../core/qt-everywhere-opensource-src-4.8.5/src/plugins/platforms/emboxvc/emboxvcwindowsurface.cpp:342: undefined reference to `thread_cancel(thread*)'
> 
> Though I have added `thread_cancel.h` in `thread.h` and have also tried by adding `@Runlevel(1) include embox.kernel.thread.thread_cancel` in `mods.config`. Could you please help me to understand where am I doing the mistake. Thanks :)

Thanks for the reminder! Is this problem still actual?

alexkalmuk:
Please, update your branch up to embox/master.

PR #1921, opened 2020-03-31
Toggle led through Lua UDP server on stm32f7-disco

alexkalmuk:
* Add a very simple Lua example of how to toggle LEDs over the network.
* Split config for stm32 gpios because different boards have different gpio count.

On Linux:
```
$ lua5.3 echoclnt.lua 192.168.0.128 1027
Using remote host '192.168.0.128' and port 1027...
pin gpioi 1 toggle
pin gpioi 1 toggle

```

On Embox:
```
embox>lua udp_server.lua                                                        
Lua UDP server started on port 1027...
Execute 'pin gpioi 1 toggle' from 192.168.0.102:34300
Execute 'pin gpioi 1 toggle' from 192.168.0.102:34300
```

### DIFF
--- a/src/cmds/hardware/pin/pin.c
+++ b/src/cmds/hardware/pin/pin.c
@@ -25,28 +25,27 @@ static void print_usage(void) {
 		"    pin GPIOA 12 blink  -- Toggle pin number 12 at GPIOA\n");
 }
 
-static int gpio_by_name(char *name) {
-	if (!strncmp("GPIO", name, 4) || !strncmp("gpio", name, 4)) {
-		name += 4;
-	}
+static int gpio_by_name(const char *name) {
+	int id = -1;
 
-	if (strlen(name) == 0) {
+	if (strncmp("GPIO", name, 4) && strncmp("gpio", name, 4)) {
 		return -1;
 	}
-
-	switch (name[0]) {
-	case '0': case 'a': case 'A': return GPIO_PORT_A;
-	case '1': case 'b': case 'B': return GPIO_PORT_B;
-	case '2': case 'c': case 'C': return GPIO_PORT_C;
-	case '3': case 'd': case 'D': return GPIO_PORT_D;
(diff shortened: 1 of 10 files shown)


src/cmds/hardware/pin/pin.c
```diff
@@ -41,7 +41,18 @@ static int gpio_by_name(char *name) {
 	case '3': case 'd': case 'D': return GPIO_PORT_D;
 	case '4': case 'e': case 'E': return GPIO_PORT_E;
 	case '5': case 'f': case 'F': return GPIO_PORT_F;
+	case '6': case 'g': case 'G': return GPIO_PORT_G;
+	case '7': case 'h': case 'H': return GPIO_PORT_H;
+	case '8': case 'i': case 'I': return GPIO_PORT_I;
+	case '9': case 'j': case 'J': return GPIO_PORT_J;
+	case 'k': case 'K': return GPIO_PORT_K;
 	default:
+		if (strlen(name) > 1 && name[0] == '1') {
```
Deryugin: `name = "101"` will return `GPIO_PORT_K`

May be rewrite this function to something like

```
if (toupper(name[0]) >= 'A' && toupper(name[0]) <= 'Z')) {
   idx = toupper(name[0]) - 'A';
} else {
   if (1 != sscanf(name, "%d", &id)) {
        fprintf(stderr, "Failed...");
        return 0;
   }

   return id;
}


```

alexkalmuk: yeah, good suggestion :)

alexkalmuk: Please, take a look.

alexkalmuk: Hm, with a proposed solution

```
pin gpioabcdef 1 toggle
```

will also work.

We should check that the sting is one in size before checking for `a - z`.

alexkalmuk: Fixed.

PR #1922, opened 2020-04-01
Memmove test suite

kouamdo:
`memmove` function shall return the pointer of the destination data


kouamdo:
src should be null because after memmove , the data source should be free

src/tests/stdlib/memmove.c
```diff
@@ -0,0 +1,59 @@
...
+#include <string.h>
+#include <stdlib.h>
+#include <stdio.h>
+
+EMBOX_TEST_SUITE("Test suite for memmove() function");
+
+
+TEST_CASE("move full data")
+{
+    const    char    src[]="string";
```
anton-bondarev, line 11: why can you not use char  src[]="string"; (without constant)?

src/tests/stdlib/memmove.c
```diff
@@ -0,0 +1,59 @@
...
+    char    dest[sizeof(src)+1];
+
+    test_assert_not_null(dest) ;
+    memmove((void*)dest,(void*)src,(size_t)strlen(src));
```
anton-bondarev, line 15: Too many casts

strlen returns size_t without casting
I think 
`memmove(dest, src, strlen(src));`
will enough

src/tests/stdlib/memmove.c
```diff
@@ -0,0 +1,59 @@
...
+
+EMBOX_TEST_SUITE("Test suite for memmove() function");
+
+
+TEST_CASE("move full data")
+{
+    const    char    src[]="string";
+    char    dest[sizeof(src)+1];
+
+    test_assert_not_null(dest) ;
```
anton-bondarev, line 14: excess cheking

src/tests/stdlib/memmove.c
kouamdo: ohhh sorry , i have done mistake ; thank you

src/tests/stdlib/memmove.c
kouamdo: yes , to be very careful .... :) :) :+1:

src/tests/stdlib/memmove.c
```diff
@@ -0,0 +1,59 @@
...
+    test_assert_equal(*dest , *src) ;
+}
+
+
+TEST_CASE("check for NULL argument")
+{
+    const   char  src[]="\0";
+    char    dest[strlen(src)+1];
+
+    test_assert_not_null(dest) ;
```
anton-bondarev, line 25: excess checking
Remove all strings like
```
test_assert_not_null(dest) ;
```

PR #1925, opened 2020-04-03
Rework skb xmit

anton-bondarev:


src/net/l0/net_entry.c
```diff
@@ -30,75 +32,114 @@ static DLIST_DEFINE(netif_rx_list);
...
-	ipl_restore(sp);
+	ipl_restore(ipl);
+
+	return 0;
 }
 
-static void netif_rx_dequeued(struct net_device *dev) {
-	ipl_t sp;
+/* we can be in irq mode */
+int netif_rx(void *data) {
```
Deryugin, line 60: May be change this function signature to ` int netif_rx(struct sk_buff *skb) {` to make sure correct type is passed here?

anton-bondarev: Thanks, 
It's the only for pnet support, but if it's changed pnet will breake

PR #1926, opened 2020-04-03
net: Rework neighbour

anton-bondarev:


src/net/neighbour.c
```diff
@@ -362,68 +349,39 @@ int neighbour_foreach(neighbour_foreach_ft func, void *args) {
 	return 0;
 }
 
-int neighbour_send_after_resolve(unsigned short ptype,
+int neighbour_resolve(unsigned short ptype,
 		const void *paddr, unsigned char plen,
-		struct net_device *dev, struct sk_buff *skb) {
-	int allocated, resolved;
+		struct net_device *dev, struct sk_buff *skb,
+		unsigned char hlen_max, void *out_haddr) {
```
alexkalmuk, line 355: As far as I can see `hlen_max` is unused in this function.

alexkalmuk: Btw, it looks like we can remove `hlen` as a parameter, also in neighbour_add for example.

anton-bondarev: Thanks, it's out_haddr length.
I think It must be checked that the buffer size is enough for dev->addr (dev->addr_len)

alexkalmuk: Ok, please check also neighbour_add. There we also fill haddr but do not rely on hlen (only checking for `(hlen > sizeof(nbr->haddr`)

anton-bondarev: thanks, done

PR #1932, opened 2020-04-06
drivers/net/lan91c: Rework

Deryugin:
* Make sure to avoid parallel execution of xmit and rx operations as
  they mess up inner state of net card
* Check for successfull packet allocation on xmit
* Check for non-empty FIFO on rx
* Other minor fixes

### DIFF
--- a/src/drivers/net/lan91c111/lan91c111.c
+++ b/src/drivers/net/lan91c111/lan91c111.c
@@ -52,6 +52,7 @@ EMBOX_UNIT_INIT(lan91c111_init);
 /* Bank 2 */
 #define BANK_MMU_CMD    (BANK_BASE_ADDR + 0x0)
 #define BANK_PNR        (BANK_BASE_ADDR + 0x2)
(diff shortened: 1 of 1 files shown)


src/drivers/net/lan91c111/lan91c111.c
```diff
@@ -252,55 +262,68 @@ static irq_return_t lan91c111_int_handler(unsigned int irq_num, void *dev_id) {
...
+		packet = (REG16_LOAD(BANK_FIFO_PORTS) >> 8) & 0xFF;
+		if (packet == FIFO_EMPTY) {
+			goto out;
+		}
 
-	skb = skb_alloc(len + 2);
-	assert(skb);
-	skb->len = len;
-	skb->dev = dev_id;
+		REG16_STORE(BANK_POINTER, (0x8000 | AUTO_INCR | 2));
```
anton-bondarev, line 288: 0x8000 | | 2 may be set macro for them

Deryugin: Done

PR #1941, opened 2020-04-10
Clean lan91c111 netcard

anton-bondarev:
* Clean lan91c111 netcard
* Enable deferred sending if ALLOC_TX was not succed

### DIFF
--- a/src/drivers/net/lan91c111/lan91c111.c
+++ b/src/drivers/net/lan91c111/lan91c111.c
@@ -9,10 +9,10 @@
 
 #include <util/log.h>
 #include <drivers/common/memory.h>
-#include <hal/reg.h>
-#include <kernel/printk.h>
+
 #include <kernel/irq.h>
 #include <mem/misc/pool.h>
+#include <hal/reg.h>
 
 #include <net/inetdevice.h>
 #include <net/l0/net_entry.h>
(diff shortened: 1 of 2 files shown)


src/drivers/net/lan91c111/lan91c111.c
```diff
@@ -252,89 +204,127 @@ static const struct net_driver lan91c111_drv_ops = {
...
+	struct smc_local *smc;
+	int ret;
+
+	smc = netdev_priv(dev);
+
+	skb = skb_queue_pop(&smc->pending_queue);
+	if (!skb) {
+		return 0;
+	}
+	lan91c111_send_hw(skb);
```
Deryugin, line 218: Looks like theese lines can be removed as `while` look does the same thing.

src/drivers/net/lan91c111/lan91c111.c
```diff
@@ -252,89 +204,127 @@ static const struct net_driver lan91c111_drv_ops = {
...
-		skb_data = skb_data_cast_in(skb->data);
-		assert(skb_data);
+	lan91c111_set_cmd(CMD_RX_POP_AND_RELEASE);
 
-		for (i = 0; i < len >> 2; i++) {
-			buf = REG32_LOAD(BANK_DATA);
-			*((uint32_t *)(skb_data + i * 4)) = buf;
-		}
+	if (buf & (ODD_CONTROL << 8)) {
+		skb->len++;
```
Deryugin, line 274: Are you sure it's neccessary? This also may cause buffer overflow.

src/drivers/net/lan91c111/lan91c111.c
```diff
@@ -164,55 +78,93 @@ static void lan91c111_push_data(uint8_t *data, int len) {
...
+
+static int lan91c111_add_pending(struct smc_local *smc, struct sk_buff *skb) {
+	uint8_t int_mask;
+
+	lan91c111_set_bank(2);
+
+	skb_queue_push(&smc->pending_queue, skb);
+
+	int_mask = REG8_LOAD(BANK_INTERRUPT_MASK);
+	int_mask &= ~ALLOC_INT;
```
Deryugin, line 122: Should it be `int_mask |= ALLOC_INT` instead to enable irq?

src/drivers/net/lan91c111/lan91c111.c
anton-bondarev: Yes, but, when we first-time appear in this point it means that ALLOC_TX command was finished. Of cause we can call ALLOC_TX again, it's not significant I think

src/drivers/net/lan91c111/lan91c111.c
anton-bondarev: this code just move from old file line 315 to own routine lan91c111_recv_hw

src/drivers/net/lan91c111/lan91c111.c
anton-bondarev: ooh, thanks a lot.
Done!

PR #1943, opened 2020-04-10
Mouse and keyboard tests

alexkalmuk:
* Add mouse and keyboard simple tests.
* Fix PS/2 mouse
* Fix PS/2 mouse and PS/2 keyboard race condition and initialization order - now it doesn't matter in which order you included these modules in the config, moreover you can include only mouse or only keyboard.

To verify how these tests work, include the following modules in `x86/qemu`:

```
    @Runlevel(1) include embox.arch.x86.boot.multiboot(video_mode_set=1, video_width=800, video_height=600, video_depth=16)
    @Runlevel(1) include embox.driver.video.bochs
    @Runlevel(1) include embox.driver.video.fb
    @Runlevel(2) include embox.driver.input.keyboard.keyboard
    @Runlevel(2) include embox.driver.input.mouse.ps_mouse
    @Runlevel(3) include embox.init.GraphicMode
    include embox.cmd.testing.mouse_fb_test
    include embox.cmd.testing.keyboard_test
```

Then you can run only mouse test:
```
root@embox:/# mouse_fb_test mouse
```
or keyboard:
```
root@embox:/# keyboard_test keyboard
```
or both:
```
root@embox:/# mouse_fb_test mouse &
root@embox:/# keyboard_test keyboard &
```

Press on QEMU graphical window after running the tests and move your mouse or press keyboard keys.

### DIFF
--- a/src/drivers/input/i8042/i8042.c
+++ b/src/drivers/input/i8042/i8042.c
@@ -1,0 +1,58 @@
+/*
+ * @file
+ *
+ * @date   10.04.2020
+ * @author Alexander Kalmuk
+ */
+
+#include <embox/unit.h>
+#include <drivers/i8042.h>
+
+EMBOX_UNIT_INIT(i8042_init);
+
+int i8042_read_mode(void) {
+	i8042_wait_write();
+	outb(I8042_CMD_READ_MODE, I8042_CMD_PORT);
+	i8042_wait_read();
+	return inb(I8042_DATA_PORT);
+}
+
+void i8042_write_mode(uint8_t val) {
+	i8042_wait_write();
+	outb(I8042_CMD_WRITE_MODE, I8042_CMD_PORT);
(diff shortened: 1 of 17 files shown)


src/drivers/input/i8042/i8042.c
```diff
@@ -0,0 +1,58 @@
...
+
+void i8042_write_aux(unsigned char val) {
+	i8042_wait_write();
+	outb(I8042_CMD_WRITE_MOUSE, I8042_CMD_PORT);
+	i8042_wait_write();
+	outb(val, I8042_DATA_PORT);
+}
+
+static int i8042_init(void) {
+	unsigned char mode;
```
anton-bondarev, line 35: may be uint8_t?

alexkalmuk: You're right, it's better for the driver. Fixed.

PR #1944, opened 2020-04-10
Udp test

anton-bondarev:
* Add udp test server
* Clean net/ethernet


src/net/l0/net_tx.c
```diff
@@ -68,7 +68,12 @@ int net_tx(struct sk_buff *skb, struct net_header_info *hdr_info) {
...
 	assert(dev != NULL);
 
-	if (hdr_info->dst_p == NULL || (dev->flags & IFF_NOARP)) {
+	if (dev->flags & (IFF_LOOPBACK | IFF_NOARP)) {
+		eth_hdr(skb)->h_proto = htons(hdr_info->type);
+		goto send;
+	}
+
+	if (hdr_info->dst_p == NULL) {
 		/* it's local we don't need resolve hardware address */
```
alexkalmuk, line 77: Is `hdr_info->dst_p == NULL` reachable now? You check if it is a loopback device above and go to send if so, it looks like this condition is unreachable anymore.

src/net/l0/net_rx.c
```diff
@@ -33,14 +35,7 @@ int net_rx(struct sk_buff *skb) {
...
 	}
 
-	/* parse L2 header */
-	assert(skb->dev->ops != NULL);
-	assert(skb->dev->ops->parse_hdr != NULL);
-	if (0 != skb->dev->ops->parse_hdr(skb, &hdr_info)) {
-		log_error("%p can't parse header", skb);
-		skb_free(skb);
-		return 0; /* error: can't parse L2 header */
-	}
```
alexkalmuk, line 43: You removed this code, but src/net/l2/ethernet.c:68 still remains:
```
const struct net_device_ops ethernet_ops = {
    .build_hdr = ethernet_build_hdr,
    .parse_hdr = ethernet_parse_hdr,
    .check_addr = is_valid_ether_addr,
    .check_mtu = ethernet_check_mtu
};
```
And now it is unused. We do not need this logic anymore? And now, all our network stack depends on ethernet here because of `eth_hdr(skb)->h_proto`, is it ok?

alexkalmuk: I mean we can completely delete parse_hdr if you are using eth_hdr directly now.

src/net/l0/net_tx.c
anton-bondarev: It's  also for broadcast

src/net/l0/net_rx.c
anton-bondarev: Yes, I agree, Done

PR #1947, opened 2020-04-14
Fix usb mouse

alexkalmuk:
* Fix usb mouse (compilation and functioning).
* (minor) auto_qemu: Multiplex stdio and qemu monitor when running in graphical mode, so you can use qemu monitor in console (like for -nographic mode).

You can verify usb mouse on OHCI in the following way:
* Add to `x86/qemu`:
```
    @Runlevel(1) include embox.arch.x86.boot.multiboot(video_mode_set=1, video_width=800, video_height=600, video_depth=16)
    @Runlevel(1) include embox.driver.video.bochs
    @Runlevel(1) include embox.driver.video.fb
    @Runlevel(1) include embox.driver.usb.class.hid
    @Runlevel(3) include embox.init.GraphicMode
    include embox.cmd.testing.mouse_fb_test
```
* Run `./scripts/qemu/auto_qemu -device usb-mouse`
* `mouse_fb_test usb-mouse`. Then go to qemu graphical window, press there and move your mouse.

On wiki - https://github.com/embox/embox/wiki/Using-USB-on-qemu#usb-mouse

### DIFF
--- a/src/drivers/usb/class/usb_mass_storage.c
+++ b/src/drivers/usb/class/usb_mass_storage.c
@@ -177,7 +177,7 @@ static struct usb_device_id usb_ms_id_table[] = {
 	{ },
 };
 
-struct usb_driver usb_driver_ms = {
+static struct usb_driver usb_driver_ms = {
 	.name = "mass_storage",
 	.probe = usb_ms_probe,
 	.disconnect = usb_ms_disconnect,
--- a/scripts/qemu/auto_qemu
+++ b/scripts/qemu/auto_qemu
@@ -353,9 +353,9 @@ if [ $graphic_n -gt 0 ]; then
 				graphic_lines="-nographic"
 			else
 				if [ $graphic_model = "pl110" ]; then
-					graphic_lines="-serial stdio"
+					graphic_lines="-serial mon:stdio"
 				else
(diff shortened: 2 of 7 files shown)


src/drivers/usb/class/usb_mass_storage.c
```diff
@@ -177,7 +177,7 @@ static struct usb_device_id usb_ms_id_table[] = {
 	{ },
 };
 
-struct usb_driver usb_driver_ms = {
+static struct usb_driver usb_driver_ms = {
```
anton-bondarev, line 180: const?

alexkalmuk: Your suggestion is pretty good, but because of not very good usb driver registration if I add `const`, then I have to change:

```
usb_driver_register(&usb_driver_ms);
```
to 
```
usb_driver_register((struct usb_driver *) &usb_driver_ms);
```

to discard `const` qualifier. That's why I added only `static`.

alexkalmuk: I created an issue with suggestion to make USB drivers statically register - https://github.com/embox/embox/issues/1948

anton-bondarev: Thanks

PR #1949, opened 2020-04-15
Input devices improvements

alexkalmuk:
* Rework input devices event handling logic. There is no more common handler, each input device drivers defines its own irq handler and call `input_dev_report_event` when want to pass an event to input device core. Also, add irq locks to make input device core safer.
* Add `input` command to get list of available input devices.
* Minor fixes and renamings: for instance, mouse -> ps-mouse, keyboard -> ps-keyboard.


src/cmds/testing/input/input_test/Mybuild
```diff
@@ -1,5 +1,4 @@
-
-package embox.cmd
+package embox.cmd.testing.input
```
Deryugin, line 1: Please change package name for templates where `input_test` is included.
```
> grep input_test -rn ./templates/
./templates/arm/ti816x/mods.config:51:	include embox.cmd.input_test
./templates/arm/ti816x-qt/mods.config:44:	include embox.cmd.input_test
```

alexkalmuk: Thank you, fixed in https://github.com/embox/embox/pull/1950

Issue #1886, opened 2020-03-19
mips/qemu don't build with OpenLibm 


anton-bondarev:
If enabele openlibm in mips/qemu
```
	//include embox.compat.libc.math_builtins
	include embox.compat.libc.math_openlibm
```
build fails
```
./build/base/obj/embox.o: In function `print_f':
src/compat/libc/include/ctype.h:(.text+0x5af00): undefined reference to `roundl'
src/compat/libc/include/ctype.h:(.text+0x5afe8): undefined reference to `roundl'
```

puranjaymohan:
I want to work on this issue

puranjaymohan:
@anton-bondarev Please tell me how to proceed with solving this issue. I am kind of stuck here.
Thanks

Deryugin:
@puranjaymohan I think that problem is in `openlibm-0.7.0/src/Make.files`. For some reason `s_roundl.c` (which implements `roundl()`) is compiled only for `i387 amd64 aarch64` architectures. Looks like it can be solved the same way it was solved for arm (take a look at `embox/third-party/lib/openlibm/patch.txt`).

anton-bondarev:
@puranjaymohan Have you reproduced the trouble?

I see in openlibm/patch.txt
```
+ifneq ($(filter $(ARCH),arm),)
+$(CUR_SRCS) += s_roundl.c
+endif
+
```

puranjaymohan:
Hi @Deryugin Thanks for the insight, I have used the same method to solve it for mips. 

@anton-bondarev Yes I have reproduced the issue and also solved it now after your suggestions. A pull request has been created #1956.
Thanks

PR #1956, opened 2020-04-20
third-party: openlibm: add 'mips' in the patch file

puranjaymohan:
Fix the build error related to roundl() function by adding
mips to the list of architectures in patch file of openlibm.
Solves #1886  
Signed-off-by: Puranjay Mohan <puranjay12@gmail.com>

### DIFF
--- a/third-party/lib/openlibm/patch.txt
+++ b/third-party/lib/openlibm/patch.txt
@@ -87,7 +87,7 @@ diff -aur openlibm-0.7.0/src/Make.files ../../../../../../tmp/openlibm/src/Make.
  	s_catanl.c s_csinl.c s_cacosl.c s_cexpl.c s_csinhl.c s_ccoshl.c \
  	s_clogl.c s_ctanhl.c s_ccosl.c s_cbrtl.c
  endif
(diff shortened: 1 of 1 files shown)


third-party/lib/openlibm/patch.txt
```diff
@@ -87,7 +87,7 @@ diff -aur openlibm-0.7.0/src/Make.files ../../../../../../tmp/openlibm/src/Make.
  	s_catanl.c s_csinl.c s_cacosl.c s_cexpl.c s_csinhl.c s_ccoshl.c \
  	s_clogl.c s_ctanhl.c s_ccosl.c s_cbrtl.c
  endif
-+ifneq ($(filter $(ARCH),arm),)
++ifneq ($(filter $(ARCH),arm,mips),)
```
anton-bondarev, line 90: It's work
I think it should be space separated
```
ifneq ($(filter $(ARCH),arm mips),)
```

puranjaymohan: Yes, it was working but now I have changed it to space-separated as you said.

PR #1957, opened 2020-04-21
Fix usb-net

alexkalmuk:
Here usb-net device is fixed. There are some issues like mass storage cannot currently work in parallel with usb-net but you can check usb-net as follows:

* Comment in`x86/qemu`:
```
//@Runlevel(1) include embox.driver.usb.class.mass_storage
//@Runlevel(2) include embox.driver.net.virtio
```
and add usb-net:
```
@Runlevel(1) include embox.driver.net.usbnet
```

Then just start with auto_qemu and use ping, telnet, httpd etc.


src/drivers/usb/usb_hub.c
```diff
@@ -62,27 +68,37 @@ struct usb_dev *usb_new_device(struct usb_dev *parent,
...
+			if (++cfg > 0) {
+				usb_free_configuration(dev);
+			}
+			/* Fill device configuration. */
+			if (usb_get_configuration(dev, cfg) < 0) {
+				log_error("usb_get_configuration failed");
+				goto out_err;
+			}
+		/* Skip Microsoft's RNDIS */
+		} while (is_rndis(dev->iface_desc[0]));
```
anton-bondarev, line 92: If device is RNDIS we call usb_free_configuration() after usb_get_configuration() ?
Or it means that RNDIS use cfg 1 instead 0?

alexkalmuk:
> LGTM
> I've checked it.
> But why did you say that it's required to comment line
> 
> ```
> @Runlevel(1) include embox.driver.usb.class.mass_storage
> ```

It works fine when you are using auto_qemu, but if you try auto_qemu_with_usb_storage, then embox hangs.

src/drivers/usb/usb_hub.c
alexkalmuk: Yes, if device is RNDIS we free current configuration (rndis) and try to get next configuration. We repeat this until we find not rndis configuration and then set up it.

anton-bondarev:
@alexkalmuk Ok. I see. Thanks

PR #1967, opened 2020-05-01
kernel: Enable task resource for atexit

anton-bondarev:


src/kernel/task/Mybuild
```diff
@@ -59,6 +59,8 @@ module task_resource {
 
 	@NoRuntime depends embox.compat.libc.assert
 	@NoRuntime depends embox.util.Array
+	//FIXME
+	@NoRuntime depends embox.kernel.task.resource.atexit
```
alexkalmuk, line 63: Is it because it's required by limits.h ?

anton-bondarev: Yes. I've no idea, how to do this more pretty

PR #1968, opened 2020-05-01
Drivers: virtual: mem: added virtual device /dev/mem

### LINKED ISSUE #1914: /dev/mem compatibility
Some Linux applications access physical addresses via devfs like this

```
    uintptr_t phys_addr; /* Target physical address */
    size_t size; /* Size of memory region */

    m_fd = open("/dev/mem", O_RDWR | O_SYNC);
    virt_addr = mmap(0, size, PROT_READ | PROT_WRITE,
        MAP_SHARED, m_fd, phys_addr);
   ...
```

It would be nice to implement pseudo device `/dev/mem` for devfs to be able to run such programs.

puranjaymohan:
Add a character driver that implements /dev/mem file and
helps in accessing physical memory using devfs.
closes #1914

Signed-off-by: Puranjay Mohan <puranjay12@gmail.com>

### DIFF
--- a/src/drivers/virtual/Mybuild
+++ b/src/drivers/virtual/Mybuild
@@ -14,3 +14,12 @@ module zero {
 	depends embox.fs.driver.devfs
 	depends embox.driver.char_dev
 }
+
+module mem {
+	source "mem.c"
+
+	depends embox.compat.libc.all
+	depends embox.fs.driver.devfs
+	depends embox.driver.char_dev
+	depends embox.fs.syslib.idesc_mmap
+}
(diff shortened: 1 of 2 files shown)


anton-bondarev:
@puranjaymohan Thanks. That's all right. But I'm not sure about mmap_device_memory() using. I need to think a little about this.

puranjaymohan:
@anton-bondarev Do we have a function like `remap_page_range()`, I saw that Linux uses this function to perform the memory mapping of physical address to virtual address. I we have something similar to this then we can use it here.

anton-bondarev:
> @anton-bondarev Do we have a function like `remap_page_range()`, I saw that Linux uses this function to perform the memory mapping of physical address to virtual address. If we have something similar to this then we can use it here.

@puranjaymohan Thanks
No, there is no remap_page_range() function in Embox. You are right to use map_device_memory() instead!

We'll merge it soon.

PR #1976, opened 2020-05-10
Extend da146xx templates with timers and threads

alexkalmuk:


src/drivers/interrupt/cortexm/cortexm0_nvic.c
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * @author  Anton Kozlov
+ * @date    02.07.2012
+ */
+
+#include <hal/reg.h>
+#include <drivers/irqctrl.h>
+#include <kernel/irq.h>
+
+#define NVIC_BASE          0xe000e100
```
anton-bondarev, line 13: Fixed address?

alexkalmuk: Yes, I am still in doubt whether to use it as an option or not, because it is defined in ARM architecture, and it is not a peripheral device...

PR #1982, opened 2020-05-12
kernel : work with disk in lkl_task

FilippChubukov:


src/kernel/task/lkl_task.c
```diff
@@ -38,11 +77,59 @@ static int handler(uint32_t nr, void * data) {
...
+	}
+
+	ret = lkl_sys_write(fd, "test", 4);
+
+	if (ret < 0) {
+		printk("can't write: %s\n", lkl_strerror(ret));
+		return -1;
+	}
+
+	ret = lkl_sys_fsync(fd);
```
AntonKozlov, line 126: The line ensures block device's write function will be called.

src/kernel/task/lkl_task.c
```diff
@@ -38,11 +77,59 @@ static int handler(uint32_t nr, void * data) {
...
+		printk("Can't open the file: %s\n", lkl_strerror(fd));
+		return -1;
+	}
+
+	ret = lkl_sys_write(fd, "test", 4);
+
+	if (ret < 0) {
+		printk("can't write: %s\n", lkl_strerror(ret));
+		return -1;
+	}
```
AntonKozlov, line 124: A test that will be removed after

PR #1996, opened 2020-05-24
Add arch_idle() for arm/qemu

### LINKED ISSUE #1971: arm/qemu uses 100% one core of host cpu
It's enough to use arch_idle() with speeling

sk2sat:
Fix #1971

### DIFF
--- a/src/arch/arm/subarch/armv5te/arm926ejs/Mybuild
+++ b/src/arch/arm/subarch/armv5te/arm926ejs/Mybuild
@@ -1,9 +1,14 @@
 
 package embox.arch.arm.arm9.arm926ejs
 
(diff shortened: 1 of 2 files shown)


anton-bondarev:
@sk2sat Great! It works for me.
But it's better to use a special module instead of source "arch.c"
```
module arch extends embox.arch.arch {
	source "arch.c"
}
```
and add a dependency on the new module

anton-bondarev:
@sk2sat Thanks a lot.
I've approved. Please rebase under Embox master and squash commits

anton-bondarev:
@sk2sat I've merged. Thanks a lot!

PR #2028, opened 2020-06-12
Use actions/cache in CI

AntonKozlov:
In CI, Mybuild cache is prepared at the start and then distributed to testing jobs by using Actions Artifact.

There is actions/cache that is intended to cache data between runs and jobs, allows to eliminate the cache building and save time

A run from master: https://github.com/embox/embox/actions/runs/132366028

A run with actions/cache: https://github.com/embox/embox/actions/runs/133189816 (after the Mybuild cache was preparied)
* fast templates start on ~1.5 min earlier and do not take longer; first results appears after ~2min
* same for slow templates, overall run time is reduced to 14min

Downside is that earlier Mybuilds are parsed to create a cache, so cache always matched Mybilds. With this, cache is searched by hashing Mybuilds, so it's possible to get wrong cache for a set of Mybuilds. Caching algorithm desribed in https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#hashfiles

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,24 +3,7 @@ name: CI
 on: [push]
 
 jobs:
-  # does Mybuilds processing and provides mk/.cache to other jobs
-  buildgen:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v1
-      - run: make confload-test/buildgen
-      - run: make -j $(nproc) buildgen
-      - name: compress
-        run: |
-          mkdir mk-cache
-          tar zcf mk-cache/archive.tar.gz mk/.cache
-      - uses: actions/upload-artifact@v1
-        with:
-          name: mk-cache
-          path: mk-cache
-
   ci:
-    needs: buildgen
     runs-on: ubuntu-16.04 # Embox NFS write test fails with ubuntu 18.04: "RPC: Can't decode result"
     strategy:
       fail-fast: false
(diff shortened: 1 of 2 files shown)


anton-bondarev:
> Downside is that earlier Mybuilds are parsed to create a cache, so cache always matched Mybilds. With this, cache is searched by hashing Mybuilds, so it's possible to get wrong cache for a set of Mybuilds. Caching algorithm desribed in https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#hashfiles

@AntonKozlov Will the problem appear for every change Mybuild files?

AntonKozlov:
Please read description of the problem like below

> Earlier Mybuilds are parsed to create a cache, so cache always matched Mybilds. The new approach downside is that cache is searched by hashing Mybuilds, so it's possible to get wrong cache for a set of Mybuilds. Caching algorithm desribed in https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#hashfiles

I expect the problem will not appear at all in most cases

anton-bondarev:
@AntonKozlov  Yes, I see about hashing Mybuld. But I read caching algorithm description and don't understand where the problem can appear at all. As I see the cache is changed if Mybuild file is changed.

anton-bondarev:
@AntonKozlov Maybe try to change some Mybuilds?

AntonKozlov:
> @AntonKozlov But I read caching algorithm description and don't understand where the problem can appear at all. As I see the cache is changed if Mybuild file is changed.

Hash is updated, that is correct. Hash looses information, its possible (although unlikely) that a set of Mybuilds to have a hash value to be equal to the same hash value of another set of Mybuilds. If that happen, old Mybuild cache would be taken and the build system will not detect this inconsistency in any way.

AntonKozlov:
> @AntonKozlov Maybe try to change some Mybuilds?

Do you suggest to update a Mybuild and see the cache will not be hit? I think that should happen, no problems in everyday life.

PR #2030, opened 2020-06-13
2012 rm slowdown module

### LINKED ISSUE #2012: Rm embox.kernel.time.slowdown module
embox.kernel.time.slowdown module is weird. It's used to make slower system clock, but why system_freq is not been able to use for the aim. Why there some code with this logic exists in clock_tick_handler

```
	if (++cs->jiffies_cnt == (1 << SLOWDOWN_SHIFT)) {
		cs->jiffies_cnt = 0;
		cs->jiffies++;
```
```
templates/x86/test/packetdrill/mods.config:44:	include embox.kernel.time.slowdown(shift=3) /* 1 << 3 */
templates/arm/qt-fb-small/mods.config:39:	include embox.kernel.time.slowdown(shift=2)
templates/arm/qt-app/mods.config:45:	include embox.kernel.time.slowdown(shift=2)
```

anton-bondarev:
Close #2012

### DIFF
--- a/src/kernel/time/clock_tick.c
+++ b/src/kernel/time/clock_tick.c
@@ -28,10 +28,8 @@ void clock_tick_handler(int irq_num, void *dev_id) {
 
 	cs->jiffies++;
 
-	if (cs_jiffies->event_device &&	irq_num == cs_jiffies->event_device->irq_nr) {
-		if (timer_strat_need_sched(cs_jiffies->jiffies)) {
-			lthread_launch(&clock_handler_lt);
-		}
+	if (timer_strat_need_sched(cs_jiffies->jiffies)) {
+		lthread_launch(&clock_handler_lt);
 	}
 }
 
--- a/src/kernel/time/clock_tick.c
+++ b/src/kernel/time/clock_tick.c
@@ -6,7 +6,6 @@
  * @author Alexander Kalmuk
  */
 
-#include <module/embox/kernel/time/slowdown.h>
 #include <kernel/irq_lock.h>
 #include <kernel/time/timer.h>
 #include <kernel/time/clock_source.h>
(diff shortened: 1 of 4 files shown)


src/kernel/time/clock_tick.c
```diff
@@ -27,15 +25,12 @@ void clock_tick_handler(int irq_num, void *dev_id) {
...
-		cs->jiffies_cnt = 0;
-		cs->jiffies++;
-
-		if (cs_jiffies->event_device &&	irq_num == cs_jiffies->event_device->irq_nr) {
-			if (timer_strat_need_sched(cs_jiffies->jiffies)) {
-				lthread_launch(&clock_handler_lt);
-			}
-		}
+
+	cs->jiffies_cnt = 0;
```
alexkalmuk, line 29: `jiffies_cnt` seems to be not required anymore

src/kernel/time/clock_tick.c
```diff
@@ -27,15 +25,12 @@ void clock_tick_handler(int irq_num, void *dev_id) {
...
+	cs->jiffies++;
+
+	if (timer_strat_need_sched(cs->jiffies)) {
```
alexkalmuk, line 32: Also, I proposed here to change `cs` -> `cs_jiffies` to make this code looks more clear.

alexkalmuk: Because we are using cs_jiffies in lthread below.

alexkalmuk: Or alternatevely you can set `assert (cs == cs_jiffies)` instead.

src/kernel/time/clock_tick.c
anton-bondarev: thanks. done

src/kernel/time/clock_tick.c
anton-bondarev: agree. done

PR #2053, opened 2020-06-23
add riscv64

sk2sat:
Add minimal arch/riscv64 & riscv64/qemu template.

It can boot on qemu-system-riscv64(but stuck at initializing clock driver).

PR #2073, opened 2020-06-30
Fix riscv64 interrupt

### LINKED ISSUE #2064: Enable timer for riscv64/qemu

sk2sat:
Fix #2064.

I thought timer initialization was not working but it was wrong.

In my previous PR(#2053), trap handler does not call interrupt handler.
So, a timer interrupt occurs after timer initialization cannot be handled and stuck(and "done" was not printed well).


anton-bondarev:
@sk2sat Thank you very much.
I've merged it!

PR #2087, opened 2020-07-07
Rework riscv interrupts

anton-bondarev:


src/include/hal/clock.h
```diff
@@ -32,7 +32,7 @@ extern void clock_init(void);
  */
 extern void clock_setup(useconds_t useconds);
 
-extern void clock_tick_handler(int irq_num, void *dev_id);
```
alexkalmuk, line 35: It's required for risc-v somehow?

alexkalmuk: I mean removing `irq_num`.

alexkalmuk: Ah, it's just unused. So it's a drawback of the previous timer fixes.

anton-bondarev: Yes, you're right.
It's saved only for compatible with irq_atach()

PR #2095, opened 2020-07-16
Switch build system to absolute paths

### LINKED ISSUE #654: Embox build system should generate absolute paths
See summary. Otherwise it's impossible to build 3rd-party libs which have complex build systems with various levels of changing the current directory.

E.g. stuff like "-I./src/include -I./src/arch/arm/include -I./build/base/src-gen/include" in CFLAGS is no-go.
Cc: @abusalimov

anton-bondarev:
It seems it closes #654

### DIFF
--- a/mk/image2.mk
+++ b/mk/image2.mk
@@ -161,11 +161,6 @@ $(OBJ_DIR)/module/%.o :
 
 # Here goes image creation rules...
 #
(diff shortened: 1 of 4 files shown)


alexkalmuk:
All looks good to me. But qt-vnc is broken, and I cannot recognize if this happened due to these changes or not (it seems due to these changes though).

PR #2131, opened 2020-08-06
add: maixbit flash script

### LINKED ISSUE #2121: Create script for loading firmware to MAiX Bit board
Create script and wiki page

sk2sat:
fix #2121

### DIFF
--- a/scripts/maixbit-flash.sh
+++ b/scripts/maixbit-flash.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+# @file
+# @brief Load firmware to MAiX BiT board
+#		Requires kflash.py(https://github.com/kendryte/kflash.py)
+# @author	sksat
+# @date		05.08.2020
+#
+
+BUILD_DIR=./build/base/bin
+EM_BIN=$BUILD_DIR/embox.bin
+
+K210_PORT=/dev/ttyUSB0
+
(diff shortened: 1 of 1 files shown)


scripts/maixbit-flash.sh
```diff
@@ -0,0 +1,34 @@
...
+# @brief Load firmware to MAiX BiT board
+#		Requires kflash.py(https://github.com/kendryte/kflash.py)
+# @author	sksat
+# @date		05.08.2020
+#
+
+BUILD_DIR=./build/base/bin
+EM_BIN=$BUILD_DIR/embox.bin
+
+KFLASH=/usr/bin/kflash
```
anton-bondarev, line 12: Why do you declare KFLASH, you don't use it as I see
and in my system kflash locate in other place
```
which kflash 
/usr/local/bin/kflash
```

sk2sat: I thought it would be nice to be able to specify kflash location as env, but no problem if kflash installed correctly and the location is in PATH. I will remove it.

anton-bondarev: Yes, you right it is nice to specify the location. it will be able to be added in the future

anton-bondarev:
@sk2sat Thanks/ Please rebase and squash commits

PR #2139, opened 2020-08-11
Nuklear optimize for stm32

alexkalmuk:
Apply double buffering and LTDC features to optimize Nuklear rawfb example on stm32f7 up to 60 FPS.

### DIFF
--- a/third-party/bsp/stmf7cube/arch.c
+++ b/third-party/bsp/stmf7cube/arch.c
@@ -29,6 +29,11 @@
 #include <framework/mod/options.h>
 #include <module/embox/arch/system.h>
 
+#define ENABLE_CPU_CACHE   OPTION_GET(BOOLEAN, enable_cpu_cache)
+
+#define SRAM1_START        0x20010000
+#define SDRAM_START        0x60000000
+
 static void SystemClock_Config(void)
 {
   RCC_ClkInitTypeDef RCC_ClkInitStruct;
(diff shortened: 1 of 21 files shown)


third-party/bsp/stmf7cube/arch.c
```diff
@@ -69,11 +74,58 @@ static void SystemClock_Config(void)
...
+	MPU_InitStruct.Enable = MPU_REGION_ENABLE;
+	MPU_InitStruct.AccessPermission = MPU_REGION_FULL_ACCESS;
+	MPU_InitStruct.IsBufferable = MPU_ACCESS_NOT_BUFFERABLE;
+	MPU_InitStruct.IsCacheable = MPU_ACCESS_CACHEABLE;
+	MPU_InitStruct.IsShareable = MPU_ACCESS_NOT_SHAREABLE;
+	MPU_InitStruct.TypeExtField = MPU_TEX_LEVEL0;
+	MPU_InitStruct.SubRegionDisable = 0x00;
+	MPU_InitStruct.DisableExec = MPU_INSTRUCTION_ACCESS_ENABLE;
+
+	/* Configure the MPU attributes as WT for SRAM
```
embeddedt, line 99: It might be worth noting that certain Cortex-M7 revisions have [errata](https://static.docs.arm.com/epm064408/80/Cortex-M7_Software_Developers_Errata_Notice_v8.pdf#page=11) related to the use of write-through cache.

alexkalmuk: Thanks!

We are having problems with stm32f7629i discovery now, maybe due caches, but I am still not sure. That's why this PR is WIP.

Generally, we would like to avoid using WT cache mode, because it is slower than write-back. So it is temporary solution derived from STM32Cube.

alexkalmuk: After your comment, I am almost sure we should avoid WT cache :)

alexkalmuk:
Now nuklear rawfb widget example works on bot stm32f7 boards, but there is a drawback - different double buffering methods for these boards - https://github.com/embox/embox/issues/2152

Issue #2016, opened 2020-06-02
cortex-m7: Enable Data and Instruction L1 caches


alexkalmuk:
This will possibly increase the performance of our applications (pjsip, opencv, etc.). In stm32 cube you can search for `CPU_CACHE_Enable`. But enabling of these caches may lead to instruction and data barriers in our stm32 drivers.

alexkalmuk:
It was partially enabled (in write-through mode) in https://github.com/embox/embox/pull/2139.

alexkalmuk:
Further work was done in https://github.com/embox/embox/issues/2016. Here we enabled D-cache in write-back mode and added the ability to allocate non-cacheable memory regions - statically in SRAM through linker section `.bss.sram_nocache` or regions with a fixed address and size.

alexkalmuk:
It looks like SD card DMA issue https://github.com/embox/embox/issues/2163 is the only thing to enable cpu cache at least in our basic stm32f7 templates.

PR #2162, opened 2020-08-27
ARMv7-M cpu cache enable

alexkalmuk:
This PR aims to make the previous version of CPU caches for ARMv7-M cleaner. (it's a part of the issue https://github.com/embox/embox/issues/2016).

* Add I-Cache and D-Cache fo ARMv7-M in default write-back mode (basically for SRAM and SDRAM) with the ability to declare non-cacheable memory regions, instead of hardcoded write-through mode before.
* Use non-cacheable memory in stm32 ethernet driver
* Nuklear uses cache clean/ cache inval because it requires two frame buffers which cannot be marked as non-cachable. They were marked as write-through before, but it still requires cache invalidate.

btw, I do not enable d-cache for the stm32f7 basic templates (just add cpu cache commented) because new problems with SD card controller can arise - it uses DMA to work with Embox's block devices and I cannot make this memory non-cacheable right now in a simple elegant way.

### DIFF
--- a/project/nuklear/cmd/grfx_rawfb/rawfb_stm32_ltdc.c
+++ b/project/nuklear/cmd/grfx_rawfb/rawfb_stm32_ltdc.c
@@ -69,6 +69,8 @@ static void dma2d_fill(uint32_t *dst, uint16_t xsize, uint16_t ysize, uint32_t c
 			}
 		}
 	}
+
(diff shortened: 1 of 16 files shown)


anton-bondarev (approved):
LGTM

PR #2168, opened 2020-09-03
templates: Add Raspberry Pi 0 template

Inokinoki:
Add template for Raspberry Pi 0.

Raspberry Pi 0 has the same SoC BCM2835 with Raspberry Pi 1 model B, except that it has a 1GHz CPU and 512M RAM rather than the 700MHz CPU and 256M RAM in Raspberry Pi 1 model B. So the existing modules work well with it.

-----------------

This template also works on Raspberry Pi 0 W (with WiFi and Bluetooth), with `dtoverlay=disable-bt` option in `config.txt`.

Those models which have Bluetooth and WiFi, have more than one UART:

- a PL011 UART
- a mini UART

By default, in Raspberry Pi 0 W, after booting, the PL011 UART will be redirected to the Bluetooth tty. And, the mini UART will work as a UART tty. At the moment, we can disable the Bluetooth to use the default PL011 UART as the UART tty.

For more information:

https://www.raspberrypi.org/documentation/configuration/uart.md

I'll elaborate on a mini UART driver and add a Raspberry Pi 0 W-targeted termplate.


Inokinoki:
Will add a Wiki page as well

anton-bondarev (approved):
LGTM

anton-bondarev:
@Inokinoki Thanks!

> Will add a Wiki page as well

Yes, Could you please add a wiiki page

PR #2093, opened 2020-07-13
Memory aligned allocation

kouamdo:
PR for aligned memory allocation

PR #2170, opened 2020-09-06
Posix mem align 

kouamdo:
POSIX aligned memory allocation solved issue

### DIFF
--- a/src/tests/stdlib/valloc_test.c
+++ b/src/tests/stdlib/valloc_test.c
@@ -1,0 +1,49 @@
+/**
+ * @file
+ * @brief
+ *
+ * @date 04.07.2020
+ * @author KOUAM Ledoux
+ */
+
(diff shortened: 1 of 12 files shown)


src/tests/stdlib/valloc_test.c
```diff
@@ -0,0 +1,49 @@
...
+ * @author KOUAM Ledoux
+ */
+
+#include <embox/test.h>
+#include <errno.h>
+#include <stdlib.h>
+#include <unistd.h>
+
+const int _const_size_ = 0xFFF ;
+
```
alexkalmuk, line 15: Why you need global variable? Use here macro instead

alexkalmuk:
What about https://github.com/embox/embox/pull/2093?

kouamdo:
problem of #2093 was resolve

PR #2172, opened 2020-09-11
add: Raspberry Pi mini UART driver

Inokinoki:
Mini UART is the default console device on Raspberry Pi models with
Bluetooth and WiFi. It's a NS16550-like UART. But some registers are
different. For example, we need enable manually the mini UART by
writting into <base-address> + 0x04 register.

This will help further ports for Raspberry Pi zero W, 3B, 3B+, etc.

### DIFF
--- a/src/drivers/serial/rpi_mini_uart/rpi_mini_uart.c
+++ b/src/drivers/serial/rpi_mini_uart/rpi_mini_uart.c
@@ -1,0 +1,135 @@
+/**
+ * @file
+ * @brief Raspberry Pi 16550-like Mini UART driver
+ * Mini UART is the default console device on Raspberry Pi models with
+ * Bluetooth and WiFi. But some registers are different. For example,
+ * we need enable manually the mini UART by writting into base_address
+ * + 0x04 Auxiliary Enable Register.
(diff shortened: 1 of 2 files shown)


anton-bondarev (approved):
LGTM

anton-bondarev:
@Inokinoki Thanks a lot.
I've approved your PR, but could you add your PR description as a comment to your driver to explain what's different between NS16550 and rpi_mini_uart

src/drivers/serial/rpi_mini_uart/rpi_mini_uart.c
```diff
@@ -0,0 +1,135 @@
+/**
+ * @file
+ * @brief Raspberry Pi 16550-like Mini UART driver
+ * Mini UART is the default console device on Raspberry Pi models with
```
Inokinoki, line 4: Is this OK? Any format issue?
I added more info about the other registers which are different with the ones in NS16550.

In fact, the PR message is the one I added in the commit message.

anton-bondarev: Ok. Thank you very much. It's enough!

PR #2173, opened 2020-09-12
stm32f7 fix nuklear double buffering

alexkalmuk:
* Fix Nuklear graphics widget example to work in the same way on 32F746GDISCOVERY and 32F769IDISCOVERY.
* Add two simple examples for stm32f7 using double buffering - static square and moving square.

We can look towards triple buffering now :)


anton-bondarev (approved):
LGTM

PR #2180, opened 2020-09-15
drivers: Fix clock rate for integrator_pit

alexkalmuk:
This incorrect rate sometimes led to long nanosleep() calibration.


anton-bondarev (approved):
:shipit:

Issue #2054, opened 2020-06-23
GDB reports corrupt stack when debugging x86/qemu


abusalimov:
```
Breakpoint 1, __schedule (preempt=22) at src/kernel/sched/sched.c:374
374		prev = schedee_get_current();
(gdb) bt
#0  __schedule (preempt=22) at src/kernel/sched/sched.c:374
#1  0x00000016 in ?? ()
#2  0x00000016 in ?? ()
#3  0x0015939d in __critical_count_add (count=1) at ./src/include/kernel/critical.h:150
#4  0x0015a271 in sched_preempt () at src/kernel/sched/sched.c:428
#5  0x00000001 in ?? ()
#6  0x0003f000 in ?? ()
#7  0x01b9ae28 in _bss_end ()
#8  0x0015a267 in sched_preempt () at src/kernel/sched/sched.c:427
#9  0x00000016 in ?? ()
#10 0x00000216 in ?? ()
#11 0x01b9ae58 in _bss_end ()
#12 0x0010af18 in critical_dispatch_pending () at src/kernel/critical.c:35
#13 0x01b9ae54 in _bss_end ()
#14 0xfffc0000 in ?? ()
#15 0x00000016 in ?? ()
#16 0x0003ffff in ?? ()
#17 0x0024cd8c in clock_source_list ()
#18 0x00000216 in ?? ()
#19 0x00000000 in ?? ()
```

### Steps to reproduce:
- build x86/qemu
- run `auto_gemu -s -S`
- attach with GDB to `:1234`
- put a breakpoint to some function, I tried in `__schedule`, and into a test case of 'critical'.
- when the breakpoint hits, observe the corrupt stack
- try next on a line with a call, it steps into instead of stepping over the call

Adding `-fno-omit-frame-pointer` has no effect.

WSL2, GCC 9.3.0, GDB 9.1, QEMU 4.2.0

alexkalmuk:
Hi! Today I faced the same problem after installing the recent ubuntu 20.04.

I checked with gcc7 that gdb works fine, but starting with gcc8 (8.4.0 for me) gdb reports corrupt stack. Examing disassembler I found `endbr32` (it's part of recent Control-flow Enforcement Technology (CET)) inserted at almost every function prologue:

```asm
00151b5d <__schedule>:
  151b5d:   f3 0f 1e fb             endbr32
  151b61:   55                      push   %ebp
  151b62:   89 e5                   mov    %esp,%ebp
[...]

PR #2182, opened 2020-09-16
i386: Fix gdb corrupt stack under QEMU

alexkalmuk:
* Fix GDB corrupt stack with GCC >= 8 under QEMU by adding `-fcf-protection=none` GCC flag (more detail here https://github.com/embox/embox/issues/2054).
* Use `-fno-pic` to suppress position-independent executable code with newer GCC versions, which avoids:
```
(gdb) n
0x001a280b in __x86.get_pc_thunk.bx ()
```


anton-bondarev (approved):
LGTM
I've checked it on x86/qemu

PR #2183, opened 2020-09-16
Project qt

anton-bondarev:
Move Qt templates to a project/qt4 folder
How to check it describes on the wiki [https://github.com/embox/embox/wiki/Qt](https://github.com/embox/embox/wiki/Qt)


alexkalmuk:
I'm getting red moveblocks on `project/qt4/arm_qemu`. I thought it is fixed in this PR, or not? If not, then ok.

anton-bondarev:
> I'm getting red moveblocks on `project/qt4/arm_qemu`. I thought it is fixed in this PR, or not? If not, then ok.

Thanks.

Yes, I think that fix framebuffer format in drivers should be done in another PR. I've just enable BRG formats in emboxfb plugin. ANd I've checked that it works if I've changed `var->fmt = RGBA8888;`->`	var->fmt = BGRA8888;` in pl110_lcd_get_var()

alexkalmuk (approved):
LGTM

PR #2185, opened 2020-09-18
pjsip fix on qemu

alexkalmuk:
* x86-qemu:
   * Build fix
   * Fix es1370
* arm-qemu build fix
* Improve audio play to handle chunks. Helps to play WAV's created with `ffmpeg -i file.mp3 file.wav` (before it was incorrect data length there)


anton-bondarev (approved):
LGTM

PR #2191, opened 2020-09-24
riscv: fix cadence gem net driver

vanyaio:


anton-bondarev (approved):
Thanks
I've checked it. ping is work, So although telnet still is not work. it ican be fix in next PRs.
I will merge PR soon.

PR #2197, opened 2020-09-27
Thread heap allocator

alexkalmuk:
The main goal is to provide the ability to use different stack sizes for threads.


anton-bondarev (approved):
LGTM

PR #2206, opened 2020-09-29
Clean kernel thread module

anton-bondarev:
Rework thread_allocator module. Add the opportunity to use user_stack instead of common stacks in threads.
Clean a bit kernel/thread folder


alexkalmuk (approved):
LGTM

PR #2211, opened 2020-10-05
Enable i2c combined transactions

### LINKED ISSUE #2188: i2c combined transaction support
Probably, we can add combined transactions - with mixed read/write's within a single transaction (currently, we support only separate transactions).

https://www.i2c-bus.org/repeated-start-condition/

It seems to be required only for the multi-master case.

anton-bondarev:
Fix #2188


alexkalmuk (approved):
LGTM

PR #2223, opened 2020-10-16
serial: Mv uart init to uart_open() routine

anton-bondarev:


alexkalmuk (approved):
LGTM

I've checked this fix (both uart input and output) on arm/qemu, platform/stellaris/lm3s811evb, stm f4 and f7.

PR #2230, opened 2020-10-27
Fix efm32zg sk3200

alexkalmuk:
Fixed efm32zg stk3200 and checked that it running well on the board.
https://github.com/embox/embox/wiki/EFM32ZG-STK3200


anton-bondarev (approved):
LGTM

Issue #2208, opened 2020-09-30
Investigate whether we can get gcc headers from toolchain


alexkalmuk:
For example, we build GCC for OpenCV to get STL headers, while there is one another way: get the STL library from the toolchain, and get headers from the toolchain. I am not sure about the details (how to get a path to includes) and whether it's possible, but we can investigate this.

p.s.
Another option is to get STL from other places, not from gcc.

alexkalmuk:
We solved this issue in #2208. In short, we have exceptions and STL in this new environment which makes our basic c++ examples (tests, stl example) pass successfully, but there is no std::thread, std::mutex support in STL which makes this solution not possible to apply for Open CV, for instance. 

There is no std::thread support because arm-none-eabi-gcc has `Thread model: single`, which disables threads. So in this solution, we highly rely on the options toolchain was built with.

Another thing is newlib. arm-none-eabi-gcc built with `--with-newlib` which makes libstdc++.a (GCC STL library) have several undefined references to newlib symbols. That's why we added the newlib_compat module.

Finally, libstdc++.a built with wchar support, so we added a bunch on stubs.

As the result, this approach provides C++ support with exceptions and STL, without std::thread support.

alexkalmuk:
Closed via https://github.com/embox/embox/issues/2208

PR #2232, opened 2020-10-29
c++: Add toolchain-only implementation

alexkalmuk:
This solves https://github.com/embox/embox/issues/2208. The main objective is to provide toolchain-only (without gcc building) c++ support with exceptions and STL. We succeed in that at least for ARM (see arm/c++_demo template 4th setup). The main limitation comes from toolchain itself - it depends on newlib (built with --with-newlib), so we added a small newlib compatible layer which makes our tests pass successfully.


anton-bondarev (approved):
LGTM

PR #2234, opened 2020-10-30
Gcc version update

alexkalmuk:
Add two newer gcc versions: `8.3.0`, `9.3.0`. They can be switched in the following way:

```
include third_party.gcc.gcc_build(gcc_version="9.3.0")
```

gcc 6.3.0 is used by default (like it was before this PR).


anton-bondarev (approved):
LGTM

PR #2238, opened 2020-11-04
stm32f7 cpu cache modes testing

alexkalmuk:
Just a sandbox for various cache modes performance comparison - write-back + write allocate / no write allocate, write-through.


anton-bondarev (approved):
LGTM

PR #2241, opened 2020-11-05
stdio: Separate fscanf() from scanf()

anton-bondarev:


src/compat/libc/stdio/Mybuild
```diff
@@ -171,8 +171,17 @@ static module perror {
...
 
+static module scanf_without_file {
+	option number fscanf_support = 0
+
+	source "scanf.c"
+}
+
 static module scanf {
+	option number fscanf_support = 1
+
```
alexkalmuk, line 182: I do not like such a kind of "constant" options, it's difficult to remember that they shouldn't be changed. May we have a single module and use `fscanf_support` inside fscanf.c to make it conditionally compiled? Or use two modules with two different FSCANF_SUPPORT macro definitions.

alexkalmuk (approved):
LGTM

PR #2243, opened 2020-11-05
Improve static assert

alexkalmuk:
Improve `static_assert` and add `static_assertf` to print a message if it fails.


anton-bondarev (approved):
:shipit:

PR #2246, opened 2020-11-07
tcp: Run tcp timer only if there are active connections

alexkalmuk:
Enable TCP timer only if required and stop the timer if there are no active TCP connections.


anton-bondarev (approved):
LGTM

PR #2247, opened 2020-11-09
net: Start neighbour timer only if required

alexkalmuk:
Start the neighbour timer only if there are neighbours in the ARP table and stop the timer otherwise.


anton-bondarev (approved):
LGTM

PR #2259, opened 2020-11-16
Improve timers

alexkalmuk:
* **Remove embox.arch.clock module**. So currently there is no arch clock anymore. Any clock event device can be used as arch clock and should be specified as `include embox.kernel.time.jiffies(cs_name="pit")`, for example.
* All clock sources are statically declared, inter-modules dependencies are simplified.
* Jiffies (main system clock source) inited directly when sched starts, so there is no more (we hope) mess with clocks' runlevels.
* Many clock drivers are fixed and verified.


anton-bondarev (approved):
LGTM

PR #2266, opened 2020-11-25
Sched ticker

alexkalmuk:
Make sched timer to be set only there are ready threads. Also, improve test for two running threads.


anton-bondarev (approved):
LGTM

PR #2272, opened 2020-12-01
Input devices devfs wrapper

alexkalmuk:
* Add devfs wrapper for input devices. It makes them to be used with usual POSIX open/read/...
* Preserve no-fs version base on lthreads for "small" devices.


anton-bondarev (approved):
LGTM

Issue #1698, opened 2020-02-21
Improve 'uname' core util command


anton-bondarev:
Currently, `uname` uses mybuild options for getting platform(-i), processor(-p) and systems (-o). Use some other way to get them.

Nishchay-sopho:
I would like to work on this issue.
I have looked at the uname.c and Uname.my files and basically got an idea about how the workflow of the code is but am not able to understand what exactly do you want from this.
Could you please guide me how to get started?

anton-bondarev:
Hi
The simplest thing in this task is get rid of `option string system = "Embox-OS"` from module description. You can just change it to a macro in C code

Then you can add a module with simple analog sysctl. For example you can look at uname().

Nishchay-sopho:
So, that means you want the information regarding the processor, platform and system to be modular as the things in struct utsname.
But we still would have to hard code the value of the macros for the system, platform and processor in the file where the macro is, wouldn't we?

anton-bondarev:
Yes, I want to remove Embox dependency 
`#include <embox/cmd.h> /* options */`

Of cause,it's better if you do that without hard code at all. I mean in sources of 'uname' command. 
We can look at [https://github.com/wertarbyte/coreutils/blob/master/src/uname.c](https://github.com/wertarbyte/coreutils/blob/master/src/uname.c) and see that processor and platform get by sysctl

Nishchay-sopho:
I have done a small implementation to improve this functionality in 17d310b8a51fc149051baa8e82187853757e3547 . 

Could you please let me know if or not this is what was desired ?

anton-bondarev:
Yes, it looks like what I wish.
But you should prepare separate PR for this task and wait when it'll merge before start another issue.

Nishchay-sopho:
Oh yeah, I didn't realize that while pushing it to the master of my forked repository. 

I will remove this commit for now and wait for #1712 to get merged and then push this one and create a PR for the same.

Is it okay?

anton-bondarev:
Yes, it's ok. Thanks.

alexkalmuk:
@Nishchay-sopho Hi, how goes? If you are not going to continue with this issue we can unassign you from this issue.

kssuraaj28:
Hi, can I try this one?

kssuraaj28:
Hi, I would like to try this :)

anton-bondarev:
> Hi, I would like to try this :)

[...]

PR #2277, opened 2020-12-02
Refactor uname cmd

kssuraaj28:
Attempt on #1698. Please let me know if this is what you require. Also, I would like to know if there is a way to get the 'processor' details (from arch options). I have currently only refactored the existing code, with no new features

### DIFF
--- a/src/cmds/sys/Uname.my
+++ b/src/cmds/sys/Uname.my
@@ -26,10 +26,7 @@ package embox.cmd.sys
 module uname {
 	source "uname.c"
 
-	option string system = "Embox-OS"
-	option string platform = ""
-	option string processor = "unknown"
-
 	depends embox.compat.posix.util.utsname
+	depends embox.lib.sysctl
 	depends embox.framework.LibFramework
 }
--- a/src/cmds/sys/uname.c
+++ b/src/cmds/sys/uname.c
@@ -7,10 +7,10 @@
  * @author Ilia Vaprol
  */
 
-#include <embox/cmd.h> /* options */
 #include <unistd.h>
 #include <stdio.h>
 #include <sys/utsname.h>
+#include <lib/sysctl.h>
 #include <errno.h>
 #include <string.h>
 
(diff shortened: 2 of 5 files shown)


src/compat/posix/include/sys/sysctl.h
```diff
@@ -0,0 +1,27 @@
...
+__BEGIN_DECLS
+
+struct sysct {
+	const char *processor;
+	const char *platform; 
+	const char *system;
+};
+
+
+extern int sysctl(struct sysct *);
```
alexkalmuk, line 23: This looks good, but are you sure it's POSIX? ( I cannot find this here - https://pubs.opengroup.org/onlinepubs/9699919799/). If it's not POSIX may be move this code to `src/lib` for example.

kssuraaj28: Okay, I will do so

alexkalmuk: Thanks. Can I also ask you to push two commits instead of this one? The first one is about adding `sysctl` library, and another one is about using this library for `uname` command. This way makes the commits more atomic, independent, and easy to review.

kssuraaj28: So, do I add the corresponding .h file in src/lib as well? or add it in one of the include directories?

alexkalmuk: Yes, you can add .h file in the same directory along with sysctl.c and then use IncludeExport to export this .h file to be accessed further from uname command.

```
$ grep -nri IncludeExport src/lib/
```

kssuraaj28:
Is this what you wish?

alexkalmuk:
> Is this what you wish?

@kssuraaj28 Yes, thanks, it looks fine now.

alexkalmuk (approved):

anton-bondarev:
@kssuraaj28 Thanks!
I've merged it

PR #2282, opened 2020-12-05
2281 gpio input

### LINKED ISSUE #2281: Make gpio input irq be handled out of irq context
* Run lthread on GPIO trigger instead of a direct call from interrupt context
* Extend `pin` command with the ability to verify GPIO input.

alexkalmuk:
Fixes #2281 

You can verify gpio interrupts with `pin` command (more details here https://github.com/embox/embox/wiki/GPIO#pin-command)


anton-bondarev (approved):
LGTM
I've checked this on stm32f769 discovery

PR #2301, opened 2020-12-22
Qt animatedtiles stm32

alexkalmuk:
Run animatedtiles Qt app on stm32f746g-discovery. Comment moveblocks and uncomment animatedtiles in start_script.inc to verify.


anton-bondarev (approved):
LGTM

PR #2303, opened 2020-12-24
qt: Hide cursor for touch screen

alexkalmuk:
* Hide cursor for touch screen (stm32f7)
* Fix non-cacheable MPU attrs.


anton-bondarev (approved):
LGTM

Issue #2047, opened 2020-06-20
stm32f7-discovery: Network is not working (sometimes) on first reset


alexkalmuk:
When you flash stm32f7-discovery (746g or 769i boards, for example) you can wonder why ping is not working. But if you reboot your board (without powering off) - e.g. `monitor reset halt`, then network works fine.

anton-bondarev:
it seems work If interface set up twice 
I added following into start_scrip.inc
```
"ifconfig eth0 down",
"ifconfig eth0 192.168.2.128 netmask 255.255.255.0 hw ether AA:BB:CC:DD:EE:02 up",
```

PR #2308, opened 2020-12-27
drivers: Fix stm32 ethernet MAC setup

alexkalmuk:
This fixes a strange situation when the ethernet starts working only after board rebooting. So now you don need to reboot your board to make the network working.

It fixes https://github.com/embox/embox/issues/2047


anton-bondarev (approved):
LGTM

PR #2326, opened 2021-01-14
Pjsip nuklear demo

alexkalmuk:
Add PJSIP example with simple GUI on STM32F769I-Discovery.

template: `make confload-project/pjsip/stm32f769i-discovery-sip-nuklear`
test: `embox> sip_nuklear /dev/stm32-ts`


anton-bondarev (approved):
LGTM
I've checked it

PR #2328, opened 2021-01-15
1703 improve ci httpd

anton-bondarev:
Try to fix after discussion in #1703


anton-bondarev:
@AntonKozlov We were discussing that 'cp' failed and HTTP CI test does not check with the copied gold file.
What do you think about this approach?

AntonKozlov:
I woule add https://github.com/AntonKozlov/embox/commit/7cbd4ff85a82ea052fe7f56533dfaa0b00a764dd, it was one of my points. Looks good in the rest

AntonKozlov (approved):
Follow up after discussion with Anton. There is a conflict of index.html coming from a module with `@RootFS` annotation and one coming from template/rootfs/. A reasonable way is to use a different page name for the test. The question is how to get this name. It could be only a convention (which I assumed). Or it could be coded into the test script (the approach with copying file to the config and rebuilding). There are drawbacks in both cases, so considering importance of the problem, we decided to take the initial way.

Issue #2324, opened 2021-01-08
Port LVGL to Embox


anton-bondarev:
Light and Versatile Graphics Library [https://lvgl.io/](https://lvgl.io/)

alexkalmuk:
@embeddedt We would like to mention that we added basic support for LVGL (btw, awesome library!) to Embox. Here is a [video](https://youtu.be/BrR9AWcMBfE) and [wiki](https://github.com/embox/embox/wiki/LVGL).

embeddedt:
Very cool! :clap:

anton-bondarev:
I think It's done so can be closed.

PR #2355, opened 2021-02-16
Port lvgl

alexkalmuk:
It's related to https://github.com/embox/embox/issues/2324. Currently, LVGL is running under x86/qemu:

output


anton-bondarev (approved):
LGTM

PR #2359, opened 2021-02-19
1769 sp804 driver

EugenySvirin:
Completing #1769 based on https://github.com/embox/embox/pull/1879


EugenySvirin:
Oops, wrong one

PR #2361, opened 2021-02-19
Port lvgl to stm32f746g-discovery and stm32f769i-discovery

alexkalmuk:
https://github.com/embox/embox/wiki/LVGL

Add basic support for LVGL on STM32 (related to https://github.com/embox/embox/issues/2324). 

I do not use DMA or DMA2D in this PR, and I hope to extend this support, nevertheless I got 33 FPS on both stm32f746g-discovery and stm32f769i-discovery. Probably, it's a good idea to extract a command part that performs double (and maybe triple) buffering and controls DMA, so in that case we could use this anywhere and do not copy the code related to double buffering from one library to another.


anton-bondarev (approved):
LGTM

PR #2370, opened 2021-02-21
Updated strlen test cases to also test strlen() and not only strnlen()

VedantParanjape:
Earlier the strlen test cases, didn't use `strlen()`, rather only `strnlen()`. Changed it as dicussed with @alexkalmuk 

I had to update the commit message to include issue number, so had to force push :0

Edit: It failed on `x86/smp` platform, don't see the reason behind this.


alexkalmuk (approved):
LGTM

Don't worry about SMP, it's not fully supported and fails from time to time.

PR #2413, opened 2021-04-12
opencv: Add imshowfb

alexkalmuk:
It's an analogue of `imshow`, which draws image directly to the framebuffer instead of a graphical window.


anton-bondarev (approved):
LGTM

PR #2415, opened 2021-04-12
Fix compiling for e2k targets with latest lcc compiler

a1batross:


a1batross:
I'm not sure about replacing PRIu64 by just %lu.

To me all these C99 modifiers aren't really reliable and should be used when we compile same code for different data type models. Current e2k port only supports LP64, so there isn't much sense it it, IMHO.

alexkalmuk:
@a1batross Thank you, can you please tell what errors GCC emits in these two cases? (`PRIu64` and `OUTPUT_FORMAT`).

a1batross:
@alexkalmuk the behavior is the same.

https://godbolt.org/z/bnPq8qdsK (gcc and clang)

https://ce.mentality.rip/z/zcbWGa (lcc and gcc)

alexkalmuk:
Hmm, it means PRIu64 is not correctly defined for E2K (I think it's here https://github.com/embox/embox/blob/master/src/compat/libc/include/inttypes.h#L208). I wonder how it was compiled with an older LCC with PRIu64 evaluated to %llx before.

And what about `OUTPUT_FORMAT`? `e3s` is not correct anymore in the new toolchain?

anton-bondarev:
@a1batross Thanks

`OUTPUT_ARCH(elbrus-v2)`
It seems  it is not a common parameter
I think it's better to use the way similar to we are using in MIPS  arch

I mean follow 
src/arch/mips/embox.lds.S
```

#ifndef OUTPUT_FORMAT_STRING
#define OUTPUT_FORMAT_STRING "elf32-bigmips"
#endif

OUTPUT_FORMAT(OUTPUT_FORMAT_STRING)
OUTPUT_ARCH(mips)
```
therefore you can set OUTPUT_FORMAT_STRING in lds.conf with
`/* define(OUTPUT_FORMAT_STRING, "elf32-tradbigmips") */`

a1batross:
@alexkalmuk I would say that PRIu64 is probably broken for every platform where uint64_t is defined as unsigned long int which is every [LP64](https://en.wikipedia.org/wiki/64-bit_computing#64-bit_data_models) platform. 

OUTPUT_FORMAT seems to be copypasted from old Linux 3.14 kernel. I think can be left that way, unless Embox kernel is compiled for different ELF format and therefore, different data type model. This is the topic of 32-bit Embox port for E2K. 

OUTPUT_ARCH e3s isn't valid anymore for new LCC. The new Linux kernel defines it that way:
```
#ifdef  CONFIG_E2K_MACHINE
 #if defined(CONFIG_E2K_ES2_DSP) || defined(CONFIG_E2K_ES2_RU)
 OUTPUT_ARCH(elbrus-v2)
 #elif defined(CONFIG_E2K_E2S)
 OUTPUT_ARCH(elbrus-v3)
 #elif defined(CONFIG_E2K_E8C)
 OUTPUT_ARCH(elbrus-v4)
 #elif defined(CONFIG_E2K_E1CP)
 OUTPUT_ARCH(elbrus-v4)
 #elif defined(CONFIG_E2K_E8C2)
 OUTPUT_ARCH(elbrus-v5)
 #elif defined(CONFIG_E2K_E2C3)
 OUTPUT_ARCH(elbrus-v6)
 #elif defined(CONFIG_E2K_E12C)
 OUTPUT_ARCH(elbrus-v6)
 #elif defined(CONFIG_E2K_E16C)
 OUTPUT_ARCH(elbrus-v6)
 #else
 error "Invalid e2k machine type"
 #endif /* all machines types */
#else   /* ! CONFIG_E2K_MACHINE */
OUTPUT_ARCH(e2k)
#endif  /* CONFIG_E2K_MACHINE */
```

I took `elbrus-v2` as it already seems to be the [target](https://github.com/embox/embox/blob/master/templates/e2k/monocube/build.conf#L10). We can also use just `e2k` which is seems to be generic option.

a1batross:
Oh, looking at `inttypes.h`, `__e2k__` is treated like 32-bit platform, like ARMv6M. It even has `!defined(__LP64__)` for RISC-V!
No wonder `PRIu64` seems to be broken, as we compile 64-bit kernel.

To be honest, some definitions doesn't make sense to me. These macros should be set depending on data type model, not the target CPU architecture or target OS of the compiler.

anton-bondarev (approved):
LGTM

anton-bondarev:
> @alexkalmuk
> `OUTPUT_FORMAT` seems to be copypasted from old Linux 3.14 kernel. I think can be left that way, unless Embox kernel is compiled for different ELF format and therefore, different data type model. This is the topic of 32-bit Embox port for E2K.
> 
> `OUTPUT_ARCH(e3s)` isn't valid anymore for new LCC. The new Linux kernel defines it that way:
> 
> ```
> #ifdef  CONFIG_E2K_MACHINE
>  #if defined(CONFIG_E2K_ES2_DSP) || defined(CONFIG_E2K_ES2_RU)
>  OUTPUT_ARCH(elbrus-v2)
>  #elif defined(CONFIG_E2K_E2S)
>  OUTPUT_ARCH(elbrus-v3)
>  #elif defined(CONFIG_E2K_E8C)
>  OUTPUT_ARCH(elbrus-v4)
>  #elif defined(CONFIG_E2K_E1CP)
>  OUTPUT_ARCH(elbrus-v4)
>  #elif defined(CONFIG_E2K_E8C2)
>  OUTPUT_ARCH(elbrus-v5)
>  #elif defined(CONFIG_E2K_E2C3)
>  OUTPUT_ARCH(elbrus-v6)
>  #elif defined(CONFIG_E2K_E12C)
>  OUTPUT_ARCH(elbrus-v6)
>  #elif defined(CONFIG_E2K_E16C)
>  OUTPUT_ARCH(elbrus-v6)
>  #else
>  error "Invalid e2k machine type"
>  #endif /* all machines types */
> #else   /* ! CONFIG_E2K_MACHINE */
> OUTPUT_ARCH(e2k)
> #endif  /* CONFIG_E2K_MACHINE */
> ```
> 
> I took `elbrus-v2` as it already seems to be the [target](https://github.com/embox/embox/blob/master/templates/e2k/monocube/build.conf#L10). We can also use just `e2k` which is seems to be generic option.

@a1batross Thanks a lot. I see about OUTPUT_ARCH() now

alexkalmuk (approved):
I cannot check with older LCC and the current changes look good to me. Thanks!

anton-bondarev:
@a1batross Thank you. I've merged it

PR #2416, opened 2021-04-13
Fix duplicating cppflags in extbld

alexkalmuk:
Fix duplicating CPPFLAGS passed to third-party apps. Also, fix the order of CPPFLAGS and CFLAGS.


anton-bondarev (approved):
:shipit:

PR #2422, opened 2021-04-16
Update opencv version to the latest one (4.5.2)

alexkalmuk:
I leave the old version (4.1.0) as default in master. But you can switch to 4.5.2 easily.


anton-bondarev (approved):
LGTM

PR #2428, opened 2021-04-28
Qt qspi load

alexkalmuk:
Simplify QT loading. Now all you need is to build the template, start openocd and run:

```
$ ./scripts/gdb_run_stm32f7_qspi.sh
```

Qt will be automatically started.

Updated wiki: https://github.com/embox/embox/wiki/Qt-on-STM32


anton-bondarev (approved):
LGTM

PR #2435, opened 2021-05-05
Automatically call ctors in C++ main()

alexkalmuk:
Automatically call ctors in C++ main() if `@CxxCmd` annotation specified.


anton-bondarev (approved):
LGTM

PR #2439, opened 2021-05-10
buildsystem: Move apps sections generation to the final stage

alexkalmuk:
Combining module's/app's sections into a single one (that's what lds-apps.mk does) is only correct at the final linking when we get a non-relocatable object file. At the earlier stages attempts to perform such a linkage on a relocatable object file may lead to undefined behaviour.


anton-bondarev (approved):
LGTM

PR #2451, opened 2021-06-03
mk: Fix md5sum output format on MacOS

### LINKED ISSUE #2450: md5 fails on MacOS
```
$ md5
MD5 (stm32cubeh7-v1.8.0.zip) = 285f779b056866ed1bd746e7cfec77c7
```

And in mk/extbld/lib.mk we use

```
$(MD5) $(pkg_archive_name) | $(AWK) '{print $$1}' | grep $(PKG_MD5) 2>&1 >/dev/null
```

The problem here is that `awk { print $1 }` will not work for macOS. We need to use either `md5 -q` or `md5 | awk {print $4}`

anton-bondarev:
Fix #2450

### DIFF
--- a/mk/variables.mk
+++ b/mk/variables.mk
@@ -78,7 +78,11 @@ export MV     := mv -f
 export PRINTF := printf
 export MKDIR  := mkdir -p
 export LN     := ln -s
+ifneq ($(shell uname -s), Darwin) 
 export MD5    := $(call find_tool_impl,MD5,md5 md5sum,md5sum)
+else
+export MD5    := md5 -q
+endif
 export CPIO   := $(call find_tool_impl,CPIO,gcpio gnucpio cpio,cpio)
 export AWK    := $(call find_tool_impl,AWK,gawk awk nawk mawk,awk)
 export TSORT  := tsort


mk/variables.mk
```diff
@@ -78,7 +78,11 @@ export MV     := mv -f
 export PRINTF := printf
 export MKDIR  := mkdir -p
 export LN     := ln -s
+ifneq ($(shell uname -s), Darwin) 
 export MD5    := $(call find_tool_impl,MD5,md5 md5sum,md5sum)
+else
+export MD5    := $(call find_tool_impl,MD5,md5 md5sum,md5sum) -q
```
alexkalmuk, line 84: I would use `md5 -q` by default:

```
export MD5 := $(call find_tool_impl,MD5,md5 md5sum,md5 -q)
```

alexkalmuk (approved):

mk/variables.mk
alexkalmuk: Since I do not see md5sum accepts `-q`

anton-bondarev: Thanks. I set "md5 -q" for MacOS

Issue #2459, opened 2021-06-12
Unsupported items for target raspi0 


kpishere:
Just to be sure I'm understanding correctly, I've gone over the code for embox and it looks like the following items are missing support for the raspi0 target: 

1) At boot, chip is not switched to Exception Level 3 which is typical exception level for the kernel to run in.
2) GPIO is unsupported
3) DMA is unsupported

I'm investigating adding these and would like to port a driver for a few SPI based LCDs, gonna have to get the above stuff working first though.  Also, my SWD ST-LINK, just does not appear to work with BCM2835!  So, will be some weeks before my new j-LINK V9 device arrives, QEMU simply will not do for this level of work.

anton-bondarev:
@kpishere Yes, you are right. Now all three points do not support Embox for RasPi. So it would be great if you added them.

> So, will be some weeks before my new j-LINK V9 device arrives,
No problem.

mfp20:
> my SWD ST-LINK, just does not appear to work with BCM2835! So, will be some weeks before my new j-LINK V9 device arrives, QEMU simply will not do for this level of work.

buy a Raspberry Pi Pico! It's a 5$ SWD programmer (using picoprobe firmware found in rpi fundation repo)! I stomped on this thread searching for someone working on the Pico (rp2040, it's an arm m0) port but I need to run in on a Zero as well.

kpishere:
@mfp20 Hah, I can't say that isn't good advise for those that are new.  Already got a V9 J-Link device, all is wired, preparing a boot firmware option now to set GPIO pins early and as needed and weekend is coming so stay tuned! PR is on it's way for GPIO support and JTAG enabled option on Pi Zero (and similar like 1 and 1+ ...).  Gathering the details has been an archeology exercise so I've been busy with this whilst waiting for hardware delivery to do some testing and validation.  Will probably make a video too to go along with showing test rig etc.

kpishere:
@anton-bondarev So, I'm working on implementing DMA.  The device is working fine, with bus memory addresses.  I'm having issues understanding accessing bus addressed memory, physical memory, and system memory.

Early on, in boot process, DMA memory to memory copying is working fine, but when I try to use same VC based memory allocations, it works until I try un-allocating the memory I'd allocated.  Then, it hangs.  It appears that my allocation then de-allocation is interfering with the other memory functions in embox after those drivers start up.

[...]

PR #2462, opened 2021-06-29
add PiZero GPIO output, add comments for frame buffer, add JTAG command

kpishere:
Adding GPIO support for RaspberryPi Zero (unit tested on) but also likely works for Pi 1, Pi B, ... (BCM283x) models.

Added option to enable configuration for JTAG debugger.

Added a command to call JTAG config for GPIO pins and output pin configuration for the 40-pin header.

Also, reverted an earlier change for default memory address for the frame buffer.  This could be a QEMU bug but text has been added to explain how one value works with QEMU and the other value works with an actual physical device.

This PR relates to ticket #2459

### DIFF
--- a/src/drivers/gpio/bcm283x/bcm283x_gpio.c
+++ b/src/drivers/gpio/bcm283x/bcm283x_gpio.c
@@ -1,0 +1,317 @@
+/**
+ * @file
+ * @brief
+ *
+ * @date    07.06.2021
+ * @author  kpishere
+ */
+
+#include <assert.h>
+#include <hal/reg.h>
+#include <drivers/common/memory.h>
+#include <embox/unit.h>
+#include <util/array.h>
+#include <kernel/printk.h>
+#include <kernel/panic.h>
+
+#include <framework/mod/options.h>
+#include "bcm283x_gpio.h"
+#include <asm/delay.h>
+#include <drivers/gpio/gpio_driver.h>
+
+#define REG_BITS 32
+
+#define PBASE                       OPTION_GET(NUMBER,base_addr)   
+#define BCM283X_PINS_NUMBER         OPTION_GET(NUMBER,gpio_pins)
+#define BCM283X_GPIO_PORTS_COUNT    OPTION_GET(NUMBER,gpio_ports)
+#define GPU_IRQ                     OPTION_GET(NUMBER,gpu_irq_int)
+
+#define BCM283X_PULL_PORTA_DEFLT_FLOAT    OPTION_GET(NUMBER,gpu_pullup_porta_float)
+#define BCM283X_PULL_PORTA_DEFLT_DOWN     OPTION_GET(NUMBER,gpu_pullup_porta_down)
+#define BCM283X_PULL_PORTA_DEFLT_UP       OPTION_GET(NUMBER,gpu_pullup_porta_up)
+
+#define BCM283X_PULL_PORTB_DEFLT_FLOAT    OPTION_GET(NUMBER,gpu_pullup_portb_float)
+#define BCM283X_PULL_PORTB_DEFLT_DOWN     OPTION_GET(NUMBER,gpu_pullup_portb_down)
+#define BCM283X_PULL_PORTB_DEFLT_UP       OPTION_GET(NUMBER,gpu_pullup_portb_up)
+
+#define PORT_PIN_TO_PINNUMBER(pt, pn)   ( ( ((pt) & 0x3 ) * REG_BITS ) + ((pn) % REG_BITS )) 
+#define GPIO_ALTERNATE_INVERT(af) ( ((af) >> 17) & 0x07 )
+#define BCM283X_GPIO_CHIP_ID    0
+
+#define IRQ_NR_PORTA  49
+#define IRQ_NR_PORTB  50
+#define IRQ_NR_PORTC  51
+#define IRQ_NR_ALL    52
+
+typedef enum _GpioPullUpDn {
+    GPUDNone = 0,
+    GPUDDown = 1,
+    GPUDUp  = 2
+} GpioPullUpDn;
+
+#define GPIO_PULL_NONE_UP_DN_STATES 3
+
+struct GpioPinData {
+    /* 0 (reserved) - Pin output set 2 - GPIO pins 64-96 (reserved on Pi Zero)
+     * 1 (data[0]) - Pin output set 0 - GPIO pins 0-31
+     * 2 (data[1])- Pin output set 1 - GPIO pins 32-53 (54-63 not connected/reserved)
+     */
+    volatile uint32_t reserved;
+    volatile uint32_t data[BCM283X_GPIO_PORTS_COUNT];  
+};
+
+struct GpioRegs {
+    volatile uint32_t func_select[3 * BCM283X_GPIO_PORTS_COUNT];
+    struct GpioPinData output_set;              // Set as output
+    struct GpioPinData output_clear;            // Set as input, is default
+    struct GpioPinData level;                   // Read for value of pin
+    struct GpioPinData ev_detect_status;        // Edge event detected
+    struct GpioPinData re_detect_enable;        // Synchoronous rising edge event
+    struct GpioPinData fe_detect_enable;        // Synchoronous falling edge event
+    struct GpioPinData hi_detect_enable;        // Hi state detected
+    struct GpioPinData lo_detect_enable;        // Lo state detected
+    struct GpioPinData re_async_detect_enable;  // Enable async rising detection in ev_detect_status
+                                                //     (no sampling by clock cycle for edge detection)
+    struct GpioPinData fe_async_detect_enable;  // Enable async falling detection in ev_detect_status
+    volatile uint32_t reserved;
+    volatile uint32_t pupd_enable;              // See func bcm283x_gpio_pin_pullupdown() below
+    volatile uint32_t pupd_enable_clocks[BCM283X_GPIO_PORTS_COUNT];
+};
+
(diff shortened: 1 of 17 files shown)


anton-bondarev:
@kpishere Thanks a lot!
Can you split the PR into several commits?
It can be not understandable when there are changes in a single commit touch on: ARM arch, drivers, configs

src/arch/arm/lib/Mybuild
```diff
@@ -6,6 +6,7 @@ static module libarch extends embox.arch.libarch {
 
 	source "stack_ptr.S"
 	source "ptregs_jmp.S"
+	source "delay.S"
```
anton-bondarev, line 9: Maybe do it as a separate module?

src/cmds/jtag.c
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+#include <unistd.h>
+#include <stdio.h>
+#include <drivers/gpio/gpio.h>
+#include <drivers/gpio/bcm283x/bcm283x_gpio.h>
+
+int main(int argc, char **argv) {
+
+	bcm283x_gpio_jtag();
```
anton-bondarev, line 16: It seems as bcm283x specific cmd. So May be place it in platform/raspbery folder?

src/drivers/gpio/bcm283x/bcm283x_gpio.c
```diff
@@ -0,0 +1,332 @@
...
+    volatile uint32_t reserved;
+    volatile uint32_t pupd_enable;              // See func bcm283x_gpio_pin_pullupdown() below
+    volatile uint32_t pupd_enable_clocks[BCM283X_GPIO_PORTS_COUNT];
+};
+
+#define REGS_GPIO ((struct GpioRegs *)(PBASE))
+
+//// Globals
+//
+uint32_t _pinMaskPortA[GPIO_PULL_NONE_UP_DN_STATES];
```
anton-bondarev, line 85: Is it global a symbol? if not you should make it static

src/drivers/gpio/bcm283x/bcm283x_gpio.c
```diff
@@ -0,0 +1,332 @@
...
+uint32_t _pinMaskPortB[GPIO_PULL_NONE_UP_DN_STATES];
```
anton-bondarev, line 86: static ?

kpishere:
@anton-bondarev  OK, will do.  I also see two textural errors and I'd commented out the Schmitt functionality which needs to be re-enabled and verified.  There was some memory copy functions I'd optimized a bit too so, yeah ... will separate those out.

kpishere:
@anton-bondarev How would you like this default value for frame buffer to be?  Works with QEMU initially or works with physical device initially?

src/drivers/gpio/bcm283x/bcm283x_gpio.c
```diff
@@ -0,0 +1,332 @@
...
+            stateSelection[2] = GPUDDown;
+        break;
+    }
+    pullPorts[port][stateSelection[0]] |= (1 << pin);
+    pullPorts[port][stateSelection[1]] &= ~(1 << pin);
+    pullPorts[port][stateSelection[2]] &= ~(1 << pin);
+}
+
+///
+/// The function is set for every pin that has high value set on port
```
anton-bondarev, line 117: Could you please use /* */ as comments. It's a minor, but a code style part

src/drivers/gpio/bcm283x/bcm283x_gpio.c
```diff
@@ -0,0 +1,332 @@
...
+    }
+
+    // Level Hi
+    if(  mode & GPIO_MODE_INT_MODE_LEVEL1 ) {
+        REGS_GPIO->hi_detect_enable.data[port] |= pins;
+        REGS_GPIO->ev_detect_status.data[port] |= pins;        
+    } else {
+        REGS_GPIO->hi_detect_enable.data[port] &= ~pins;
+    }
+/*
```
anton-bondarev, line 206: Could you please use `#if 0` for future developing parts

src/drivers/gpio/bcm283x/bcm283x_gpio.c
```diff
@@ -0,0 +1,332 @@
...
+            REGS_GPIO->ev_detect_status.data[port] = 0xFFFFFFFF;
+
+            gpio_handle_irq(&bcm283x_gpio_chip, port, pins);            
+        }
+    }
+
+    return IRQ_HANDLED;
+}
+
+void bcm283x_gpio_jtag(void) {
```
anton-bondarev, line 293: Is it possible to design a separate bcm283x_jtag module (driver) and in the init routine setup necessary pins through global GPIO API?

src/drivers/gpio/bcm283x/bcm283x_gpio.h
```diff
@@ -0,0 +1,26 @@
...
+    GFOutput = 1,
+    GFAlt0 = 4,
+    GFAlt1 = 5,
+    GFAlt2 = 6,
+    GFAlt3 = 7,
+    GFAlt4 = 3,
+    GFAlt5 = 2
+} GpioFunc;
+
+#define GPU_IRQ     OPTION_GET(NUMBER,gpu_irq_int)
```
anton-bondarev, line 22: You use OPTION_GET() but don't include its header  `#include  <framework/mod/options.h>`

src/drivers/gpio/bcm283x/bcm283x_gpio.h
```diff
@@ -0,0 +1,26 @@
+
+#ifndef SRC_DRIVERS_GPIO_BCM283X_BCM283X_GPIO_H_
+#define SRC_DRIVERS_GPIO_BCM283X_BCM283X_GPIO_H_
+
+#include <hal/reg.h>
```
anton-bondarev, line 5: Why this include is here?

src/drivers/video/Mybuild
```diff
@@ -61,11 +61,11 @@ module raspi_video {
 	 * By adding fb_vc_bus, we tell the GPU to use memory in the following way
 	 * 
 	 * L1 & L2 Cached 	 - 0x00000000 ; Only mode that works on qemu w. raspi0 emulation
-	 * L2 Cache coherent, non allocating - 0x40000000
+	 * L2 Cache coherent, non allocating - 0x40000000 ; This mode works with physical unit
 	 * L2 Cached (only)	 - 0x80000000 
 	 * Direct (no cache) - 0xC0000000
 	 */
-	option number fb_vc_bus = 0x00000000
+	option number fb_vc_bus = 0x40000000
```
anton-bondarev, line 68: Do you know that you can set up any options in mods.config ? It is just default options value in Mybuild files

anton-bondarev:
> @anton-bondarev How would you like this default value for frame buffer to be? Works with QEMU initially or works with physical device initially?

I think you can choose any variant you like and add comments which described the situation into mods.conf .

src/drivers/gpio/bcm283x/bcm283x_gpio.c
kpishere: Yes, if there is another way, I'd like to know.  The issue is that setting of the pull up/down/float state must be done across all pins of a given port at once.  You can't just change one pin at a time.  Also, the setting is persistent over power on/off cycle.  Also, you can't read the setting (that is the big difference from other systems?).  So, I have global defaults set as per data sheet, then with every setup call, modify those defaults and always with every call, define all pull up/down/floats.  The globals retain the configuration between each gpio setup call.

src/drivers/gpio/bcm283x/bcm283x_gpio.c
kpishere: OK

src/arch/arm/lib/Mybuild
kpishere: OK, will put in raspberry arch lib as it is used by gpio

src/cmds/jtag.c
kpishere: Oh, I see,  OK.  saw some examples in other platform targets.  great.

src/drivers/gpio/bcm283x/bcm283x_gpio.c
kpishere: Yes, I suppose it is.  Although, there are only a few viable options available.  In this case, I've made a non-standard selection in effort to keep SPI pins available for other connected devices.  But sure, preferences can be different for other projects.  I'll move it and provide that with 3 configurations where command line can choose another configuration and it will default to one if it is enabled at boot time.

src/drivers/gpio/bcm283x/bcm283x_gpio.c
kpishere: Oh yes!  Of course, forgot.  Static makes scope of variable limited to module.  duh. :)  Done.

anton-bondarev:
@kpishere Thanks. Now the PR Looks Good To Me. Except modify commit "memory clear for faster execution, skip unnecessary memory copies". I'm not quite sure about it. I need to more deeply investigate this part. For example, there is 'stmia' commands but it can be not supported for some subarch (cortex-m0). Could you extract this commit to a new PR to I can review it separately?

kpishere:
@anton-bondarev Ok, i've moved that commit to another branch and reverted it in this one.

anton-bondarev (approved):
LGTM

anton-bondarev:
> @anton-bondarev Ok, i've moved that commit to another branch and reverted it in this one.

Thanks! I've approved the PR, I'll merge it soon. However, for the future, you should just remove wished commit (after moving to another branch) :)

PR #2464, opened 2021-06-30
modify memory clear for faster execution, skip unnecessary memory copies

kpishere:
This is a non-critical change as it only executes on start-up.  With memory virtualization on (I think it is off now?) the memory copy could be faster in page swapping.

But, as you say, if this STMIA isn't supported on some sub-architectures, that becomes a pain to support unless it really has value. (Anybody want to test this?)

Regardless, it is here for your amusement.

### DIFF
--- a/platform/raspberry_pi/rpi0/mods.conf
+++ b/platform/raspberry_pi/rpi0/mods.conf
@@ -2,6 +2,7 @@
 package genconfig
 
 configuration conf {
+	@RunLevel(0) include embox.arch.arm.head(use_fast_boot=true)
 	@Runlevel(0) include embox.arch.arm.armv6.arm11
 	@Runlevel(0) include embox.arch.arm.armlib.interrupt
 	@Runlevel(0) include embox.arch.arm.armlib.context
--- a/src/arch/arm/Mybuild
+++ b/src/arch/arm/Mybuild
@@ -7,5 +7,6 @@ abstract module cmsis { }
 
 module head {
 	option boolean mmu_boot=false
+	option boolean use_fast_boot=false
 	source "head.S"
 }
--- a/src/arch/arm/head.S
+++ b/src/arch/arm/head.S
@@ -7,7 +7,8 @@
 
 #include <framework/mod/options.h>
 
-#define MMU_BOOT OPTION_GET(BOOLEAN,mmu_boot)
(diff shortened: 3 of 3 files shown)


anton-bondarev:
@kpishere Thanks.
I've checked compilation for cortex-m0 and have got the errors
```
/home/anton/test_embox/embox/src/arch/arm/head.S: Assembler messages:
/home/anton/test_embox/embox/src/arch/arm/head.S:90: Error: cannot honor width suffix -- `ldr r9,=_bss_vma'
/home/anton/test_embox/embox/src/arch/arm/head.S:93: Error: cannot honor width suffix -- `mov r5,#0'
/home/anton/test_embox/embox/src/arch/arm/head.S:94: Error: cannot honor width suffix -- `mov r6,#0'
/home/anton/test_embox/embox/src/arch/arm/head.S:95: Error: cannot honor width suffix -- `mov r7,#0'
/home/anton/test_embox/embox/src/arch/arm/head.S:96: Error: cannot honor width suffix -- `mov r8,#0'
/home/anton/test_embox/embox/src/arch/arm/head.S:100: Error: cannot honor width suffix -- `stmia r4!,{r5-r8}'
/home/anton/test_embox/embox/src/arch/arm/head.S:109: Error: cannot honor width suffix -- `sub r2,r1,r0'
```
to reproduce you can build 'efm32zg_stk3200'
```
make confload-platform/efm32/efm32zg_stk3200
make
```
So I suggest two ways. The first is to use `#if !defined(__ARM_ARCH_6M__)`. The second is to add an option like 'use_fast_boot ' and use `#if USE_FAST_BOOT` for split old and your versions of copying memory

kpishere:
@anton-bondarev Ok, made the change, tried to use the architecture directive, no good, not specific enough.  So added the switch defaulting to false.

anton-bondarev (approved):
LGTM

anton-bondarev:
> @anton-bondarev Ok, made the change, tried to use the architecture directive, no good, not specific enough. So added the switch defaulting to false.

I've checked on EFM32 it's ok. I thought that this code can be enabled if MMU is enabled (on big ARM sub-arch). However, someone can want to use this code on cortex-m4 for example (where is not problems with stmia )). So I've merged it as is. Thanks a lot!

PR #2467, opened 2021-07-08
Fix detection of default tunnel on QEMU startup

Kakadu:
The previous version could return two default gateways, so we need to pick the right (I decided it will be the first) one.

P.S. starting qemu fails later because of weird message `qemu-system-i386: Error loading uncompressed kernel without PVH ELF Note`, but it is very likely a separate issue


anton-bondarev (approved):
:shipit:

anton-bondarev:
@Kakadu Thanks for your contributions! I've merged the PR

PR #2470, opened 2021-07-13
Bcm2835 mem dma

kpishere:
Here is basic memory to memory DMA copying support as I see the other DMA support.  There is a test, added as a command.

Note that the GPU isn't fully supported for mailbox messages so mailbox and DMA are only supported on real hardware.

Making this PR as I see that SPI isn't implemented and will need to implement that before doing DMA to peripheral transfers and interrupt enabled transfers.  So, will be back to this at some point to add more features and tests.

I'm hearing DMA does not have posix definition so, thinking of not bothering with making devfs support for it.  thought is that it is only used directly by other device drivers.


anton-bondarev:
@kpishere Thanks a lot.
The changes look ok at the first glance. However, there are some requests before reviewing:
* Today Embox has no standard API for DMA so you add a command which uses RasPi API, could you please move it to 'platform/raspbery_pi' folder
* Split to commits (drivers, tests etc) and rebase the PR under embox/master.  'merged' commits must not be  in the commits list

kpishere:
Thanks for the suggestions.  They are done.  Also, took on an earlier suggestion to make a new target for a physical rpi0.  It is nice to see embox work right away in qemu, don't want to loose that in this project for newcomers and there won't be support in qemu for mailbox or the DMA that depends upon that any time soon so, seemed a good time for the separate target.

kpishere:
@anton-bondarev Do you think this DMA device warents a cross platform API?  Perhaps like SPI driver?  I could add and then retail the STM32 functionality but I wouldn't be able to test as I don't have a HW STM32 device.  For BCM283X devices, the driver also needs way to allocate Video Core or GPU memory.  I'm not familiar with STM32, does it have such requirements for specially allocated memory?  Maybe that could be made generic by calling it dma_malloc() and dma_free() ?  This would be memory for DMA to use for control blocks and for source/target memory.  Then, the driver also needs a way to specify the target device.  A common list of enumerated devices for the API with specific mapping to devices for each target device?

I think this as there is an SPI API but there is no DMA support in that driver right now.  So, DMA API would need to be made to  support SPI in DMA mode in that same driver.  Or, maybe not.  Maybe just extend SPI with options to support DMA in API but have graceful failover when a platform does not support DMA for SPI?  

The latter approach I can act on more independently but the former would need input from others but the former would be simpler and more consistent looking interface.

platform/raspberry_pi/cmds/dma.c
```diff
@@ -0,0 +1,65 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <mem/sysmalloc.h>
+#include <assert.h>
+
+#define MEM_SIZE 1024
+#define CPY_BLK 256
+#define DMA_CHAN 5
+
+void fill_test(uint32_t *memLoc, uint32_t bytes) {
```
anton-bondarev, line 13: please make all internal routines static. It is useful to see use the routine only in this C-file or not

platform/raspberry_pi/cmds/dma.c
```diff
@@ -0,0 +1,65 @@
...
+    uint32_t pattern = 0xDEADBEEF;
+    for(uint32_t i = 0 ; i < bytes / sizeof(uint32_t) ; i+= 2 ) {
+        memLoc[i] = pattern;
+        memLoc[i+1] = i;
+    }
+}
+
+
+// DMA memory-to-memory test
+int dma_test_mem_transfer(Dma_mem_handle *mem)
```
anton-bondarev, line 23: static

src/drivers/dma/bcm283x/bcm283x_dma.c
```diff
@@ -0,0 +1,222 @@
+/**
+ * @file
+ * @brief
+ *
+ * @date    07.06.2021
+ * @author  kpishere
+ */
+
+#include <assert.h>
+#include <kernel/printk.h>
```
anton-bondarev, line 10: Just for the future, I think it is better to use routines from util/log.h (log_debug, log_error etc.) instead printk. In this case you can control output level

src/drivers/dma/bcm283x/bcm283x_dma.c
```diff
@@ -0,0 +1,222 @@
...
+    uint32_t handle = (uint32_t)resp->data.value_32;
+
+    return handle;
+}
+
+// Lock memory to physical address mapping, returns a bus address 
+// accessable only by GPU and DMA (but corresponding physical address
+// can now be accessed by ARM CPU)
+//
+uint32_t vc_mem_lock(uint32_t handle) {
```
anton-bondarev, line 76: static ?

src/drivers/dma/bcm283x/bcm283x_dma.c
```diff
@@ -0,0 +1,222 @@
...
+    bcm2835_property_process();
+    resp = bcm2835_property_get( TAG_LOCK_MEMORY );
+
+    return resp->data.value_32;
+}
+
+// Unlock but don't de-allocate memory
+// (can't be accessed by ARM CPU but still can by GPU and DMA)
+//
+int vc_mem_unlock(uint32_t handle) {
```
anton-bondarev, line 90: static ?

src/drivers/dma/bcm283x/bcm283x_dma.c
```diff
@@ -0,0 +1,222 @@
...
+    bcm2835_property_add_tag( TAG_UNLOCK_MEMORY,  handle );
+    bcm2835_property_process();
+    resp = bcm2835_property_get( TAG_UNLOCK_MEMORY );
+
+    return resp->data.value_32;
+}
+
+// Release the allocated memory
+//
+int vc_mem_free(uint32_t handle) {
```
anton-bondarev, line 103: static ?

src/drivers/dma/bcm283x/bcm283x_dma.c
```diff
@@ -0,0 +1,222 @@
...
+    printk("txfr_len(%d) %08X\n", dma_chan, REGS_DMA(dma_chan)->txfr_len);
+    printk("stride(%d) %08X\n", dma_chan, REGS_DMA(dma_chan)->stride);
+    printk("nextconbk(%d) %08X\n", dma_chan, REGS_DMA(dma_chan)->nextconbk);
+    printk("debug(%d) %08X\n", dma_chan, REGS_DMA(dma_chan)->debug);
+}
+
+static int bcm283x_dma_init(void) {
+	int ret = 0;
+    
+#if 0
```
anton-bondarev, line 212: Interrupts do not use now?

src/include/drivers/bcm283x_dma.h
```diff
@@ -0,0 +1,162 @@
+#pragma once
```
anton-bondarev, line 1: It's more common to use `#if BCM28x_DMA_H` in the project

anton-bondarev (approved):
LGTM

anton-bondarev:
@kpishere Thanks. My apologies for the long time to review. I've approved it, but made some minor comments. Please look at them, and will change them if you agree. Except this, in my mind, it's better to use a separate folder for each command. You placed dma & jtag command in one folder.

anton-bondarev:
> @anton-bondarev Do you think this DMA device warents a cross platform API? Perhaps like SPI driver? I could add and then retain the STM32 functionality but I wouldn't be able to test as I don't have a HW STM32 device. For BCM283X devices, the driver also needs way to allocate Video Core or GPU memory. I'm not familiar with STM32, does it have such requirements for specially allocated memory? Maybe that could be made generic by calling it dma_malloc() and dma_free() ? This would be memory for DMA to use for control blocks and for source/target memory. Then, the driver also needs a way to specify the target device. A common list of enumerated devices for the API with specific mapping to devices for each target device?
> 
> I think this as there is an SPI API but there is no DMA support in that driver right now. So, DMA API would need to be made to support SPI in DMA mode in that same driver. Or, maybe not. Maybe just extend SPI with options to support DMA in API but have graceful failover when a platform does not support DMA for SPI?
> 
> The latter approach I can act on more independently but the former would need input from others but the former would be simpler and more consistent looking interface.

Yes, I think in the future we should design API for DMA, but now, I think it is ok.
I think that it is better to extend existed SPI API today, but do it in another PR

src/drivers/dma/bcm283x/bcm283x_dma.c
kpishere: Oh, Ok, that would be nice on the messages module too I'd done earlier, thanks.

src/drivers/dma/bcm283x/bcm283x_dma.c
kpishere: I think so, only want to expose dma_malloc() and dma_free(), I'm not seeing any other need to make this kind of mailbox request.  Will be static too.

src/drivers/dma/bcm283x/bcm283x_dma.c
kpishere: Yes, still have to figure that out.  and you'd only want to configure an interrupt on specific settup with a DMA channel for the driver that is using it so, but that leads to other quiestion about DMA, is it warrent a general embox driver model.

kpishere: Will remove for now and add back in either SPI driver use and/or in more general DMA device.  Probably specific example with SPI usage first and then generalze from there.

kpishere:
Changes done for this PR. Am I able to start merge? Sorry, not seeing how to do that right now.

anton-bondarev:
> Changes done for this PR. Am I able to start merge? Sorry, not seeing how to do that right now.

Thanks a lot. I've merged it.

PR #2472, opened 2021-07-18
add support for spi (poll and interrupt) for bcm2835

kpishere:
SPI polling and interrupt are supported now in this PR.  Also incorporated all of your suggestions from last review.

SPI command incorporates interrupt tests, default data, and mode switch tests.  There is an interrupt wait for input test but it seems to have a rather short wait time before reporting zero value bytes received.  Don't have a use case for this right now, so will leave it at that.

Maybe somebody else will extend support in stm32 targets for selecting Mode 1/2/3 or interrupts?

### DIFF
--- a/src/drivers/spi/bcm283x/bcm283x_spi0.c
+++ b/src/drivers/spi/bcm283x/bcm283x_spi0.c
@@ -165,7 +165,6 @@ static int bcm283x_spi0_do_transfer(struct spi_device *dev, uint8_t *inbuf
     /* Do not add log_debug() or some another stuff here,
      * because we need to write all tx data before transfer competed. */
     REGS_SPI0->dlen = DLEN_NO_DMA_VALUE;
-    REGS_SPI0->cs |= SPI0_CS_CLEAR( SPI0_tx_fifo | SPI0_rx_fifo ) | SPI0_CS_TA; // clear FIFO and Assert
     while ( ( tx_cnt < tx_count && inbuf != NULL ) 
         ||  (rx_cnt < rx_count && outbuf != NULL ) ) {
         if(tx_cnt < tx_count ) {
@@ -234,7 +233,7 @@ static int bcm283x_spi0_transfer(struct spi_device *dev, uint8_t *inbuf
     if( ( (REGS_SPI0->cs & SPI0_CS_INTD) || (REGS_SPI0->cs & SPI0_CS_INTR) ) ) {
         if(count < 0) {
             // count < 0 is signal to trigger interrupt and return
-            dev->count = -1 * count;    /* set the target amount to output in series of interrupt transfers*/
+            dev->count = -1 * count;    /* set the target amount to output in series of interrupt transfers */
             dev->in = inbuf;
             dev->out = outbuf;
             REGS_SPI0->dlen = DLEN_NO_DMA_VALUE;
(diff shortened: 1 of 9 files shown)


src/drivers/spi/bcm283x/bcm283x_spi0.c
```diff
@@ -0,0 +1,217 @@
...
+
+#include "bcm283x_spi0.h"
+
+#define PBASE		OPTION_GET(NUMBER,base_addr)
+#define AUXBASE		OPTION_GET(NUMBER,aux_addr)
+
+#define SPI_BUS_CLOCK_DIVISOR 	OPTION_GET(NUMBER, spi_bus_clock_divisor)
+
+// Toggle an extra GPIO pin at various points to debug/test timing
+#if 0
```
anton-bondarev, line 34: Maybe it is better to use a option from Mybuild for this purpose?

src/drivers/spi/bcm283x/bcm283x_spi0.c
```diff
@@ -0,0 +1,217 @@
...
+	#define DRIVER_TESTING
+	#define PIN ( 1 << 25 )
+#endif
+
+// Errata to BCM2835 behaviour: documentation states that the SPI0 DLEN register is only used for DMA.
```
anton-bondarev, line 39: Please, try to use /* */ comments

src/drivers/spi/bcm283x/bcm283x_spi0.c
```diff
@@ -0,0 +1,217 @@
...
+	to raise the priority of its AXI requests.*/
+	uint32_t dc; /* SPI DMA DREQ Controls */
+} Bcm283x_spi0;
+
+#define REGS_SPI0 ((Bcm283x_spi0 *)(PBASE))
+#define BCM283X_SPI0_TX_FIFO_LEN (16*sizeof(uint32_t))
+#define BCM283X_SPI0_RX_FIFO_LEN (16*sizeof(uint32_t))
+
+
+/* // Timing, to use elswhere, later
```
anton-bondarev, line 87: Is it for the future?

src/cmds/testing/spi/Mybuild
```diff
@@ -0,0 +1,30 @@
+package embox.cmd.testing
+
+@AutoCmd
+@Cmd(name = "spitest",
```
anton-bondarev, line 4: by the way, do you see the 'spi' command (src/cmds/hardware/spi)
https://github.com/embox/embox/tree/master/src/cmds/hardware/spi

anton-bondarev (commented):
Thank you very much. It is LGTM for me/ However could you please look at my comments, and split PR into several commits. I see at least (update template, add spitest cmd, add an spi driver for raspberry, improve SPI driver API, extend bcm2835_mailbox

src/drivers/spi/bcm283x/bcm283x_spi0.c
kpishere: Considered that but really, the meaning of this pin being set is only understood by careful code inspection, it is there only for performance/behaviour testing.  It could be removed and/or change from version to version depending on what is being actively developed/tested.  No functional or integration with other modules kind if value either so, maybe not?  Unless there are other/more reasons to put there?

src/drivers/spi/bcm283x/bcm283x_spi0.c
kpishere: Will remove, it will go into another driver

src/cmds/testing/spi/Mybuild
kpishere: I'll move the functionality over to that one.

src/drivers/spi/bcm283x/bcm283x_spi0.c
anton-bondarev: Ok, I see and agree with you. It is more convenient for this case to use a macro definition.

anton-bondarev (approved):
LGTM

anton-bondarev:
@kpishere Thanks. I've approved the PR

PR #2473, opened 2021-07-18
Correctly close file descriptor on failure

Commeownist:
I realize it probably doesn't matter at all since nobody uses 16bpp BMPs, 
but for the love of God...


anton-bondarev (approved):
:shipit:

anton-bondarev:
> I realize it probably doesn't matter at all since nobody uses 16bpp BMPs,
> but for the love of God...

@Commeownist Thanks a lot. I've merged it.

PR #2476, opened 2021-08-01
Bcm spi dma

kpishere:
Here is DMA supported transfers with SPI (and likely other devices).

There is still a bug with the DMA completion.  The test is configured to trigger an interrupt but it is not firing.  So, the busy wait loop is also polling until DMA is completed.

I'm thinking it is still worthy of sharing.  A smaller update can follow with fixing that.  (I did having it working at one point but something has changed .. possibly panic levels or mid to high water marks).

Regardless, all bytes transfer as expected.

### DIFF
--- a/platform/raspberry_pi/rpi0-hw/mods.conf
+++ b/platform/raspberry_pi/rpi0-hw/mods.conf
@@ -112,7 +112,7 @@ configuration conf {
 	include raspberry_pi.cmd.jtag
 
 	include embox.cmd.hardware.pin
-//	include embox.cmd.dma.dma // Not implemented yet
+	include embox.cmd.dma.dma
 	include embox.cmd.hardware.spi
 
 	include embox.cmd.fs.mkramdisk
--- a/src/cmds/hardware/spi/Mybuild
+++ b/src/cmds/hardware/spi/Mybuild
@@ -42,4 +42,5 @@ module spi {
     depends embox.compat.libc.stdio.printf
     depends embox.compat.libc.str
     depends embox.compat.posix.util.getopt
+	depends embox.driver.dma
 }
--- a/src/cmds/hardware/spi/spi.c
+++ b/src/cmds/hardware/spi/spi.c
@@ -13,7 +13,10 @@
 #include <util/math.h>
 
 #include <drivers/spi.h>
+#include <drivers/dma/dma.h>
 #include <util/pretty_print.h>
+
+#include <string.h>
 
 static void print_help(char **argv) {
 	printf("Transfer bytes via SPI bus\n");
@@ -81,6 +84,15 @@ static void data_received(struct spi_device *data) {
 	spi_stat = DATA_RECEIVED;
 };
 
+static irq_return_t data_block_complete(unsigned int irq_nr, void *dma_data) {
+	irq_return_t ret = IRQ_HANDLED;
+	Dma *dev = (Dma *)dma_data;
+	spi_bytes = dev->txfr_len;
+	spi_stat = SEND_COMPLETE;
+
+	return ret;
+}
+
 static int test_interrupt(struct spi_device *dev, int spi_line, uint8_t *dataOut, uint8_t *dataIn, int clkdiv, int bytes) {
 	int ret;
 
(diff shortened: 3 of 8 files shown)


kpishere:
It is also super fast.  I have to slow down the clock to 1/8 speed (33Mhz) for the connected logic analyzer but with the targeted SPI display and it's short wires, it should be able to go up to 66 Mhz.  The bytes are fully packed in the stream with no inter-message delays.  A beautiful thing.

anton-bondarev:
@kpishere Thanks a lot.
It's great that it works! However, I see changes in common parts (SPI subsystem and spi command) it seems that they can break other platforms. Could you please extract them in separate commits? and after that think about how to extend existed SPI API with your suggestions. I'll also think about it. It takes me several days to suggest a fit way.

kpishere:
@anton-bondarev Gave it another look over. The 'add support for DMA transfers' updates only the shared spi.h file.  It adds new properties only.  There is an exception, I'd changed the return type on interrupt handlers -- that was only recently added by me so I'd be expecting not too much dependancy yet?

The more problematic one may be the spi test command though.  Should I make a new command in the raspi target for the dma support?  It could be like that until we settle on a generic API for DMA?  Or, maybe you could suggest a way to make only parts compile for a given target?  Think you'd mentioned a compile-time switch based on target in a prior set of comments.  Which would you prefer to see here?

anton-bondarev:
> @anton-bondarev Gave it another look over. The 'add support for DMA transfers' updates only the shared spi.h file. It adds new properties only. There is an exception, I'd changed the return type on interrupt handlers -- that was only recently added by me so I'd be expecting not too much dependancy yet?

Ops, sorry. I see now. You changed handler type 'irq_event_t' -> 'irq_spi_event_t'. Therefore I've seen changes in spi.h. It's normal, but could you extract it into a separate commit. Besides this, we have to make sure that 'irq_event_t' is not used anywhere else.

anton-bondarev:
@kpishere Hi
I have some suggestions about how to add DMA support to 'spi' command and keep it independent on a specific platform. I think we can achieve this if add set_interrupt_handler routine to SPI API. Also, we should extend DMA API to getting DMA status (you use bcm283x_dma_in_progress() now).

The other changes is good for me. Could you please extract "update spi test command for new dma tests" commit to another branch and  if it is not difficult please split "add support for DMA transfers " commit in two (SPI API add DMA support and bcm spi add dma support)

anton-bondarev (approved):
LGTM

kpishere:
@anton-bondarev Hah!  Well I've been working on this.  I've been extending DMA to have an abstract API (No file system device, the bare minimum struct of function pointers only).  Then SPI interface can be extended in abstract way to support DMA.  I'm a little slow here as I'll have to make Qemu environment test for STM32 just to make sure the dma_test command works with a simple DMA memory copy still.

I'll make separate branches for the DMA changes, then the SPI changes and their respective command updates.  I was working with your thought that it would be nice to have one spi command for all methods, polling, interrupt, and dma transfer.

Anybody gonna pickup better support for dma on stm32?  I'm just gonna support the bare minimum that is there now. AKA no stm32 dma to peripheral support from me.

Note: once this stuff is working, I can _finally_ work on incorporating the SPI driver for LCD displays!

anton-bondarev:
@kpishere Great!
There are several STM32 in my hand now, so I can try to check and improve your driver for DMA STM32 as need. But unfortunately, I haven't any SPI devices now. However, I think it is not a big problem, because it can be solved if the problem appears in the future. Now we can do a minimal driver.

anton-bondarev:
@kpishere I even think you may use stubs for STM version at this stage.

kpishere:
@anton-bondarev Ok, so I think it is ready for STM32 testing.  I've made another PR from another branch for just the DMA changes (Bcm dma #2483).  Probably want to test/review that one before this one.  this one includes the changes in the other one as the following SPI changes are dependant on that one.  If you want to break it up and test that one separately, definitely do that one first.  If you think it is all good, then can accept this one and reject the other.

There are no functional changes from the last submission, just re-factoring to separate out device specific info for each implementation of DMA and SPI.

kpishere:
@anton-bondarev This PR for the SPI extension has been updated to the new DMA device name for PL330.  PR Ready to be completed and merged?

kpishere:
@anton-bondarev Consolidated the commits into more functional bits after rebasing off of master.

anton-bondarev:
@kpishere Thanks a lot for your efforts. Great job. I've merged the PR now.

anton-bondarev:
@kpishere I have fixed some dependencies in 2485

PR #2477, opened 2021-08-02
mk/extbld/lib.mk: check md5sum only for source archives and skip the check for git repos

a-lunev:
corrected mk/extbld/lib.mk to check md5sum only for source archives and skip the check for git repos.


anton-bondarev (approved):
LGTM

anton-bondarev:
@a-lunev Thank you for the contribution

PR #2483, opened 2021-08-07
Bcm dma

kpishere:
Here are changes to add target agnostic implementation layer to DMA driver.  The basic memory to memory STM32 copy SHOULD still be retained (needs testing first!).  The PL330 DMA (DMA for BCM283X devices) is fully implemented.  There is a basic test command for DMA too that also returns the free DMA channels for the PL330 DMA.  There is error handling for un-implemented features on STM32 devices.


anton-bondarev:
@kpishere Thanks a lot.

I've approved this PR. However, I think the DMA API doesn't fully correspond to the code style, at least naming convention ([https://github.com/embox/embox/wiki/Code-Style](https://github.com/embox/embox/wiki/Code-Style)). Now I suggest following I just open a new issue and correct DMA API in the future.
Besides this, as you wrote the correct name of the driver is PL330, could you rename it or we can do it in the future?

anton-bondarev (approved):
LGTM

kpishere:
@anton-bondarev Ok, made the name change and scanned for other K&R coding conventions (didn't see anything glaringly out of place).  Did manage to find a malloc bug in the default code! Also made adjustment to config file so that it builds nicely with just these commits.  I think this PR is done then?

anton-bondarev:
> I think this PR is done then?
 Yes, thanks. It is good enough for me.
 
>  Did manage to find a malloc bug in the default code! 
Cool!

About code style. We try to avoid 'typedef' using prefer 'struct struct_name'. Except this, you use 'Dma' naming you should use 'dma' for example. But it is minor wishes and I've already merged the PR. Thank you!

PR #2484, opened 2021-08-09
mk: Fix md5sum output format on MacOS

sikmir:
Fix #2451. GNU md5sum may be used on MacOS as well, e.g. if we build embox using nix.


anton-bondarev (approved):
:shipit:

anton-bondarev:
@sikmir Thanks a lot!

PR #2485, opened 2021-08-09
drivers: Fix dma module dependences

anton-bondarev:
fix after #2476


kpishere:
@anton-bondarev Noted.  Thank-you.

PR #2494, opened 2021-09-02
add missing argument to demo spi command

kpishere:
The demo command needed another option added to the spi_transfer() command.


kpishere:
Oops, don't need want this now/yet.

PR #2496, opened 2021-09-10
third-party: fix curl deps

sikmir:
`gmtime` provided by `embox.compat.posix.time.time` is required for `Curl_gmtime`.


anton-bondarev (approved):
:shipit:

PR #2525, opened 2021-12-14
add platform stm32f429i-discovery

0xYoshee:


anton-bondarev:
@0xYoshee Thanks a lot.
LGTM
I have only a single question. You add some files to gitignor, does it need if QtCreator is used?
Could you split your PR to 2 commits, first is "adding platform" and the second is "add QtCreator artefacts to .gitignore"

anton-bondarev (approved):
LGTM

PR #2569, opened 2022-06-06
initial commit

AlekseyZhmulin:
Add tests for wcsncasecmp()


anton-bondarev (approved):
LGTM
Thanks. I've approved the PR
Could you also add the test case into 'x86/test/unit' template? The test should start automatically. Do this in a separate commit.
And could you change the commit message with more details. "Add tests for wcsncasecmp()"will be enough.

PR #2598, opened 2022-09-01
Fix curses

AlekseyZhmulin:


anton-bondarev (commented):
Could you extract this commit to a new PR? As I see it does not refer with libcurses

anton-bondarev (approved):
LGTM

PR #2638, opened 2022-12-24
Changed SCHAR_MIN and SCHAR_MAX constant

ioannessh:
SCHAR_MIN and SHCAR_MAX have [incorrect](https://learn.microsoft.com/en-us/cpp/cpp/integer-limits?view=msvc-170) value


anton-bondarev (approved):
Thanks a lot
:shipit:

PR #2645, opened 2022-12-28
third-party: efm: changed the path to the sources archive

t0bro:
There is a corrected path to the efm sources in the thirdparty/efm/Makefile now.


anton-bondarev (approved):
LGTM

PR #2668, opened 2023-02-18
Issue 2665 fix stm32 qt

Deryugin:
This is simple script to check if current config has QSPI. If it has no SPI, then it can be loaded to the board with simple GDB command

### DIFF
--- a/src/drivers/input/input_dev_devfs.c
+++ b/src/drivers/input/input_dev_devfs.c
@@ -110,7 +110,7 @@ int input_dev_private_register(struct input_dev *inpdev) {
 
 	memset(dev, 0, sizeof(*dev));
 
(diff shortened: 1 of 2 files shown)


src/drivers/input/input_dev_devfs.c
```diff
@@ -110,7 +110,7 @@ int input_dev_private_register(struct input_dev *inpdev) {
 
 	memset(dev, 0, sizeof(*dev));
 
-	strncat(dev->name, inpdev->name, sizeof dev->name);
+	strncat(dev->name, inpdev->name, sizeof(dev->name) - 1);
```
anton-bondarev, line 113: special thanks to!

anton-bondarev (approved):
:shipit:

PR #2686, opened 2023-03-11
opendir function fix

Yuriy-Paramonov:
opendir function was fixed for case when opendir path argument points to file instead of directory.


anton-bondarev (approved):
Thanks
:shipit:

PR #2688, opened 2023-03-12
Add  sc open max to unistd h

Yuriy-Paramonov:
1. Added #define _SC_OPEN_MAX (The maximum number of open files per process) with minimum value 20 to src\compat\posix\include\unistd.h file.
2. Updated switch statement in src\compat\libc\stdlib\sysconf.c file with added case statement for _SC_OPEN_MAX.


Yuriy-Paramonov:
Remove this pull request as having extra check-in

PR #2689, opened 2023-03-12
add-exp-to-math

Yuriy-Paramonov:
1. Added file src\compat\libc\math\exp.c with implementation for exp function.
2. Added extern definition for exp function to src\compat\libc\math\math_builtins.h.
3. Updated embox.compat.libc.math_simple module in src\compat\libc\math\Mybuild with addition of: source "exp.c" statement.


anton-bondarev (approved):
Thanks
LGTM

PR #2706, opened 2023-03-23
Module embox.compat.libc.stdio.print: Added "stub" option leading to …

bursian:
…__print(...){return 1;}


anton-bondarev (approved):
LGTM

PR #2748, opened 2023-05-04
drivers: Adding qspi flash device for stm32l4-discovery

bursian:
https://github.com/embox/embox/wiki/QSPI-flash


anton-bondarev (approved):
LGTM

PR #2763, opened 2023-05-16
Added stm32 qspi flash driver & qspi_erase utility

bursian:
Most of the time QSPI is in memory mapped mode, switching to indirect modes only for erase & write operation.
QSPI flash driver now works with STM32L475E_IOT01 only, but prepared for using with STM32F769I_DISCOVERY.
The problem with STM32F769I_DISCOVERY is that BSP_QSPI_Init(), used to leave memory mapped mode, for this board takes about 1s. Also, it is necessary to include noncacheable region at 0x90000000 for STM32F769 in mods.conf:
```
	@Runlevel(0) include embox.arch.arm.armmlib.armv7m_cpu_cache(
		log_level=4,
		sram_nocache_section_size=0x4000,
		nocache_region0_addr=0x60000000, nocache_region0_size=0x00200000,
		nocache_region1_addr=0x90000000, nocache_region1_size=0x04000000
	)
```
(`dcache_flush_all()` doesn't work).


anton-bondarev (approved):
LGTM

PR #2765, opened 2023-05-17
Fix potential null pointer dereference

szsam:


src/drivers/usb/hc/ohci/ohci_pci.c
```diff
@@ -88,6 +88,10 @@ static void *ohci_ed_alloc(struct usb_endp *ep) {
 	struct ohci_td *sentinel_td = ohci_td_alloc();
 	struct ohci_ed *ed = pool_alloc(&ohci_eds);
 
+	if (!sentinel_td || !ed) {
+		return NULL;
```
Machiry, line 92: You might have to free `sentinel_td`

szsam: That's right. Fixed.

anton-bondarev: I'm agree with @Machiry you have to free allocated object if one of pointers are not null.
I think it's better to check them separately


```	struct ohci_td *sentinel_td;
	struct ohci_ed *ed; = pool_alloc(&ohci_eds);
	
 	sentinel_td = ohci_td_alloc();
	if (!sentinel_td ) {
	return NULL;
	}
        ed = pool_alloc(&ohci_eds);
        if (!ed) {
       ohci_td_free(sentinel_td);
       }
       ```

anton-bondarev (approved):
Thanks.
LGTM

szsam:
Oh, I haven't done rebasing. There is a fixup commit.

anton-bondarev:
@szsam Oh, yes. It was my fault. I see now. But I think it is not so significant. Next time I will be more careful
Thank you one more time.

anton-bondarev:
Hello @Machiry 
Thanks for the information.
I filled the survey

PR #2766, opened 2023-05-17
cmds: Fix integer overflow involving m_blocks

szsam:
m_blocks is derived from user input (a command-line argument). If m_blocks is large enough, the multiplication to calculate malloc() size could overflow.


anton-bondarev (approved):
Thanks
:shipit:

PR #2769, opened 2023-05-19
Fix comparisons that are always false

szsam:
fwrite() returns a size_t, which means that 0 > fwrite(...) will never hold.
Similarly, the type of uart->idx is unsigned char. Thus, the comparison uart->index < 0 is always false.


anton-bondarev (approved):
Thanks
:shipit:

PR #2770, opened 2023-05-19
mem: Remove 'const' modifier on a function return type

szsam:
A 'const' modifier on a function return type is useless and should be removed for clarity.

### DIFF
--- a/src/include/mem/phymem.h
+++ b/src/include/mem/phymem.h
@@ -18,8 +18,8 @@ extern struct page_allocator *__phymem_allocator;
 extern void *phymem_alloc(size_t page_number);
 extern void phymem_free(void *page, size_t page_number);
 
-extern char *const phymem_allocated_start(void);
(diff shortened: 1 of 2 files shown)


anton-bondarev (commented):
Thanks
Could you please explain why these changes do need
May be it's enough to add 'const' to return types 
return (char *) -> return (char * const)

Or you should modify also 
```
static int phymem_init(void) {
	char *const phymem_alloc_start = phymem_allocated_start();
	char *const phymem_alloc_end = phymem_allocated_end();
...
```

szsam:
`const` qualifier on a function return type is superfluous and should be removed.

Contrarily, `const` qualifier on the type of a variable, e.g. `char *const phymem_alloc_start = phymem_allocated_start();`, does have an effect. We might want to keep it.

anton-bondarev:
Ok, Thanks

PR #2800, opened 2023-06-14
drivers: Get predictable 0 field "flags" in struct spi_device

bursian:


anton-bondarev:
It seems that this is redundant. If a field is not initialized it has to set up to zero

PR #2806, opened 2023-06-17
Ubsan compilation

Deryugin:
It looks like gcc has a bug, and 'stpcpy' function name is a problem 🤷
This function is not used in the OS itself, so just extract it to a separate module, and it works fine

Minimal example to reproduce this bug:

```
> cat stpcpy.c
char *stpcpy(char *, const char *) {
  return 0;
}
> arm-none-eabi-gcc -fsanitize=undefined -c -o stpcpy.o stpcpy.c -save-temps
stpcpy.s: Assembler messages:
stpcpy.s:52: Error: expected comma after name `' in .size directive
```

### DIFF
--- a/src/compat/libc/string/Mybuild
+++ b/src/compat/libc/string/Mybuild
@@ -19,7 +19,6 @@ static module str {
 	source "strcmp.c"
 	source "strcoll.c"
 	source "strxfrm.c"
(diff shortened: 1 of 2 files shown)


anton-bondarev (approved):
:shipit:

PR #2811, opened 2023-06-23
aarch64: fix build and make it boot cleanly

pskrgag:
Hello!

I've come around embox recently and faced some issues with current aarch64 state. 

I had issues with build and boot which I've fixed in scope of this PR. I've tried to separate each change logically to make things easier to revert just in case.

Tested locally on stock 6.2.0 qemu and self-built 8.0.1.


anton-bondarev (approved):
LGTM

anton-bondarev:
@pskrgag Thank you very much!

PR #2825, opened 2023-07-03
fix sqrt to comply with standard on negative, NaN and Inf values

timtim-ww:
sqrt root checks for negative Infinity and NaN arguments and respective returns as per standard


anton-bondarev (approved):
:shipit:

PR #2829, opened 2023-07-04
fmod fix

timtim-ww:
compat: fix for fmod (ubsan complaining about builtin function)
__builtin_clzll(0) behavior is undefined and thus should be avoided
checks for zero argument added


anton-bondarev (approved):
LGTM
But could you please more careful with formatting next time

timtim-ww:
Sorry, I'll do my best =)

Issue #2263, opened 2020-11-22
chibi-scheme breaks console


anton-bondarev:
after executing start_script in `x86/test/lang`, the console starts receiving 'enter' endlessly
If the calling chibi schema is commented out `//"chibi-scheme -xscheme fib.scm",` the console is fine

PR #2833, opened 2023-07-05
Enable tish read from keyboard correctly (through linenoise)

KlyukinSA:
https://github.com/embox/embox/issues/2263

In original linenoise there is [nonblocking API](https://github.com/antirez/linenoise#asyncrhronous-api). In `embox/third-party/lib/linenoise` there is no such, so O_NONBLOCK disabling in linenoiseRawMode scope can be added to protect readline.

Now `x86/test/lang` works correctly

### DIFF
--- a/third-party/lib/linenoise/linenoise.c
+++ b/third-party/lib/linenoise/linenoise.c
@@ -115,6 +115,7 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <sys/ioctl.h>
+#include <fcntl.h>
 #include <unistd.h>
 #include "linenoise.h"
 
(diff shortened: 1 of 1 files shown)


anton-bondarev:
@KlyukinSA Thanks a lot!

It's a possible way to solve the issue.

However I looked at original linenoise.c. there is no nonbloking API using in it. Maybe the problem is that when 'chibi-scheme' returns Embox won't restore termios flags?

KlyukinSA:
@anton-bondarev 
> However I looked at original linenoise.c. there is no nonbloking API using in it. 

It is described here: https://github.com/antirez/linenoise#asyncrhronous-api. In https://github.com/antirez/linenoise/blob/master/linenoise.c there is linenoiseEditFeed() which is a part of API

anton-bondarev:
@KlyukinSA Hi, can I close this PR. It seems it fixed in #2836

KlyukinSA:
@anton-bondarev  hi yes sure

PR #2836, opened 2023-07-08
Enable tish read from keyboard correctly

### LINKED ISSUE #2263: chibi-scheme breaks console

KlyukinSA:
Fix #2263

In linux bash, `sh_unset_nodelay_mode()` is called when bash starts. There `fcntl` is called. Also in interactive mode if cmd sets `O_NONBLOCK` to stdin, bash unsets it before next cmd call.

Now `x86/test/lang` works correctly

### DIFF
--- a/src/cmds/shell/Mybuild
+++ b/src/cmds/shell/Mybuild
@@ -26,6 +26,7 @@ abstract module shell_api {
 	''')
 module tish extends shell_api {
 	option number rich_prompt_support = 0
+	option number unset_nodelay_mode = 0
 
 	option string builtin_commands = "cd export exit logout"
 
--- a/src/cmds/shell/tish.c
+++ b/src/cmds/shell/tish.c
@@ -331,8 +331,10 @@ static int rich_prompt(const char *fmt, char *buf, size_t len) {
 
 static void sh_unset_nodelay_mode(void) {
 	int orig_ifd_flags = fcntl(STDIN_FILENO, F_GETFL, 0);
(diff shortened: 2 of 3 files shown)


anton-bondarev:
@KlyukinSA Thanks!

However the modification of the flag after commands calling directly in shell is rather seems roughly. If I understand correctly the problem is only for 'chibi-scheme'.

I don't sure that I'm right, but I think it's better to save termios flags before a comand calling and than resotre it.
Exept this, it's better if this code can be disable with an options in mybuild

KlyukinSA:
@anton-bondarev 

> If I understand correctly the problem is only for 'chibi-scheme'.

Yes, chibi-scheme changes flag, see [issue](https://github.com/embox/embox/issues/2263#issuecomment-1620155477)

> I think it's better to save termios flags before a comand calling and than resotre it.

```
$ cat change_flags.c
#include <fcntl.h>

	return fcntl(0, F_SETFL, O_NONBLOCK) == -1;
}
```
```
$ cat see_flags.c
#include <stdio.h>
#include <fcntl.h>

void printBits(size_t const size, void const * const ptr)
{
    ...
}

	int a = fcntl(0, F_GETFL, 0);
	printBits(sizeof(a), &a);
}
```
```
$ ./see_flags 
00000000000010000000000000000010

$ ./change_flags 

$ ./see_flags 
00000000000010000000000000000010
```

second `./see_flags` call does not see changes, because bash does not restore flags of `change_flags`

anton-bondarev (approved):
LGTM

PR #2843, opened 2023-07-14
cmds/service:service refactor/minor execv refactor

timtim-ww:
Added executable validation check, so it can return error in case executable not found(especially in vfork) and other minor execv refactoring.
In service.c service_run: handle when execution failed; service_run - service_add already running logic. General service.c refactoring.


anton-bondarev (approved):
LGTM

PR #2844, opened 2023-07-14
shell:hardlink /bin/sh for dropbear. execv add chk

timtim-ww:
One probable scenario: after execv gained ability to return back when executable couldn't be found, an issue in dropbear that otherwise wouldn't have come up, showed itself. On connection dropbear executes '/bin/sh/' which execv answered is unavailable, after which dropbear came to a series of unfortunate events, which need further investigation, and embox laid down flat.
Hardlinking /bin/sh to tish solved that problem(at least temporary).
Additional check in execv, fixing previous commit, added.


anton-bondarev (approved):
LGTM

PR #2846, opened 2023-07-17
third-party: dropbear migrate to v2022.83

timtim-ww:
Latest version of dropbear in Embox
Got rid of some old patches for packet.c and svr-authpubkey.c of doprbear v2018.76 version
Still only in DEBUG_NOFORK with -F option running - which means exit after first logtout from client
The full server functional might be added later


anton-bondarev (approved):
:shipit:

anton-bondarev:
@timtim-ww Thank you very much

PR #2851, opened 2023-07-21
templates: x86/test/net templ fix(ci crashed)

timtim-ww:
In x86/test/net template configuration environment structure for a task(struct task_env) had default env_str_len value equal to 32. Because of that new version of dropbear crashed trying to write long environment variable.


anton-bondarev (approved):
:shipit:

Issue #1720, opened 2020-02-25
sshd (dropbear) rejects second connection


alexkalmuk:
There is an sshd server in Embox - third-party/dropbear/ (what dropbear is - https://matt.ucc.asn.au/dropbear/dropbear.html). So it is a third party application ported to Embox.

_How to start dropbear?_
* `make confload-x86/qemu`. Then add `include third_party.dropbear.dropbear` in conf/mods.config, also add `"service dropbear -F"` in conf/system_start.inc. Build embox.
* Start Embox
* Connect to sshd from your Linux host:
```
$ ssh root@10.0.2.16
root@10.0.2.16's password: 
[5] Jan 01 00:00:58 wtmp_write: problem writing /var/log/wtmp: No such entity
root@embox:/#exit
Connection to 10.0.2.16 closed.
```
_What's the problem?_
Try to connect once more:
```
$ ssh root@10.0.2.16
ssh: connect to host 10.0.2.16 port 22: Connection refused
```

This issue is probably not so easy, but we do not expect fast bug fix from you, but the problem investigation with `gdb`.

zhuanyanchengkong:
I try to do this work.

zhuanyanchengkong:
I added dropbear to embox and checked the running status,
`root@embox://etc/dropbear#service dropbear`
`Service is already running.`
`[5] Jan 0root@embox://etc/d1 01:00:29 Earropbear#ly exit: Failed to daemonize: Address family not supported by protocol family`
Then I used netstat and found the port of 22 was not open, I think maybe this is the reason can't connect the OS via ssh.
I think the first step should be to open the port that dropbear used, but I don't know how to open a port in embox.

zhuanyanchengkong:
I  checked the running status of dropbear and found an error :
> root@embox://etc/dropbear#service dropbear                                      
Starting service: dropbear
Service is already running.
root@embox://etc#[5] Jan 01 00:00:14 Early exit: Failed to daemonize: Address family not supported by protocol family

What does this error mean?

alexkalmuk:
I cannot answer your question right now since we run dropbear only with `-F` flag and only from the root directory, so the problem you described probably is some another bug. I would recommend you to start with the reproducing the behavior described in the issue. So we can follow the same the way and try to figure out what's happen.

zhuanyanchengkong:
[...]

PR #2854, opened 2023-07-24
third-party: dropbear multi-session via vfork

timtim-ww:
Dropbear becomes capable of accepting multiple sessions and reconnecting
Still there's and issue - dropbear stops accepting new connections after reaching certain limit - needs investigation
Big patch makes allocation of server session global structures dynamic. Added functionality - storing pointers to these structures in dropbear task resources.
#1720


src/include/kernel/task/resource/dropbear_res.h
```diff
@@ -0,0 +1,26 @@
...
+ * @brief Declaration of structure of addreses for dropbear session structure
+ * @date 24.07.2023
+ */
+
+#ifndef DROPBEAR_RES_H_
+#define DROPBEAR_RES_H_
+
+#include <kernel/task.h>
+
+typedef unsigned int addr_t;
```
anton-bondarev, line 13: Maybe it's better to use uintptr_t for this purpose? I mean 64 bits platform can have sizeof(void *) == 8 but sizeof(unsiged int) == 4. But I 'm not sure

anton-bondarev (approved):
Great!
LGTM

src/include/kernel/task/resource/dropbear_res.h
timtim-ww: sure thing, you're absolutely right

timtim-ww:
Corrected the pointer type and added freeing of memory allocated to session

anton-bondarev (approved):
LGTM

PR #2847, opened 2023-07-18
Fix shell scripts according to ShellCheck guidelines

wawuwo:
Hi! This PR resolves #2835 

The most part of the changes is trivial. It is quoting variables with double quotes to protect them from word splitting and globbing. This is especially important for variables containing paths (if a path in such a variable would accidentely contain spaces it would be splited into multiple substrings) and for many other variables too. Bash's word splitting can be very surprising even in obvious-looking commands: http://mywiki.wooledge.org/BashPitfalls#echo_.24foo.

Other changes are:
1. Handling of failed `cd`s, either by wrapping operation requiring `cd` in a subshell, or by explicit `|| exit 1`
2. Joining multiple 'echo string >> file' lines into a group command with a single redirect (https://www.shellcheck.net/wiki/SC2129)
3. Replacing indirect exit-code checks with direct ones, i.e. snippets like this
```
egrep '(vmx|svm)' /proc/cpuinfo > /dev/null
if [ $? -eq 0 ]; then
```
was rewritten like this
```
if grep -q -E '(vmx|svm)' /proc/cpuinfo; then
```
A nice benefit of this is that `grep`s `-q` short-circuits computation, terminating grep after first match.
4. Another changes, like style-related, etc. Most commit messages have a list of them.

Though PR name is «fix according to ShellCheck guidelines», changes introduced by it do not follow _all_ ShellCheck suggestions. I skipped SC suggestions if I wasn't sure that following it wouldn't change code's semantic. But if I followed them thinking that a change was safe, I might have been wrong and the change broke something.

PR #2857, opened 2023-07-24
2835 fix auto qemu shell scripts

wawuwo:
Hi @anton-bondarev! This PR is organized like you suggested in comments to [embox/embox/pull/2847](url): pull request includes the commits affecting `qemu/auto_qemu` and any files connected or related to it.

There are actually not many of them:
```
embox/scripts $ grep -rlF qemu/auto_qemu
autotest/run_scripts/run_qemu_default.sh
autotest/run_scripts/run_qemu_ide.sh
autotest/run_scripts/run_qemu_mmc.sh
autotest/run_scripts/run_qemu_usb.sh
continuous/fs/run.sh
continuous/run.sh
qemu/auto_qemu_with_sd_card
qemu/auto_qemu_with_usb_ccid
qemu/auto_qemu_with_usb_hub
qemu/auto_qemu_with_usb_storage
```

As in previous PR the majority of changes is double-quoting variables. Also:
- Using subshell instead of manual cd-in-then-cd-back
- Replacing indirect exit-code checks with direct ones (i.e. `if cmd; then` instead of `cmd; if [ $? -eq 0 ]; then`
- Replacing legacy command-substitution syntax `\`cmd\`` with `$(cmd)`
- Removing variables which appear unused


scripts/qemu/auto_qemu
```diff
@@ -339,18 +330,17 @@ USB2QEMUID=(
 	[uhci_pci]="uhci"
 )
 
-USB_DEV=" " read -a qemu_usb <<< "$AUTOQEMU_USB"
-USB_DEV=" " read -a qemu_usb_config <<< "$AUTOQEMU_USB_CONFIG"
```
anton-bondarev, line 343: Are you sure that this line does not need?

anton-bondarev:
@wawuwo Thanks!

I tried ./scripts/qemu/auto_qemu it seems it works OK, But I see new empty symbals before ` -net ` and `-netdev`

```
sudo -E qemu-system-arm -runas anton -M integratorcp -kernel ./build/base/bin/embox  -m 256   				-net nic,netdev=n0,model=smc91c111,macaddr=AA:BB:CC:DD:EE:02 			-netdev tap,script=./scripts/qemu/start_script,downscript=./scripts/qemu/stop_script,id=n0  -nographic 
```

Also I wrote a pare of commits
You can check this scripts youself if you build and run Embox on qemu. I said that you need a board when I said about ./scripts/gdb_run_stm32f7.. scripts

wawuwo:
@anton-bondarev, the snippets you refer to are not complete :( Complete diffs would look like this:
```diff
-USB_DEV=" " read -a qemu_usb <<< "$AUTOQEMU_USB"
-USB_DEV=" " read -a qemu_usb_config <<< "$AUTOQEMU_USB_CONFIG"
+USB_DEV=" " read -r -a qemu_usb <<< "$AUTOQEMU_USB"
```
```diff
-AUDIO_DEV=" " read -a qemu_sound <<< "$AUTOQEMU_AUDIO"
-AUDIO_DEV=" " read -a qemu_sound_config <<< "$AUTOQEMU_AUDIO_CONFIG"
+AUDIO_DEV=" " read -r -a qemu_sound <<< "$AUTOQEMU_AUDIO"
```

So, these lines are being deleted:
```
USB_DEV=" " read -a qemu_usb_config <<< "$AUTOQEMU_USB_CONFIG"
```
and
```
AUDIO_DEV=" " read -a qemu_sound_config <<< "$AUTOQEMU_AUDIO_CONFIG"
```
while others just being altered a bit.

ShellCheck marks these variables as unused:
```
In auto_qemu line 296:
AUDIO_DEV=" " read -a qemu_sound_config <<< "$AUTOQEMU_AUDIO_CONFIG"
              ^--^ SC2162 (info): read without -r will mangle backslashes.
                      ^---------------^ SC2034 (warning): qemu_sound_config appears unused. Verify use (or export if used externally).
```
```
In auto_qemu line 343:
USB_DEV=" " read -a qemu_usb_config <<< "$AUTOQEMU_USB_CONFIG"
            ^--^ SC2162 (info): read without -r will mangle backslashes.
                    ^-------------^ SC2034 (warning): qemu_usb_config appears unused. Verify use (or export if used externally).
```
And if we grep all scripts for these variables, we can find that they only declared and never used:
```
embox/scripts $ grep -r -F qemu_sound_config .
./mqtt/auto_qemu:AUDIO_DEV=" " read -a qemu_sound_config <<< "$AUTOQEMU_AUDIO_CONFIG"
./qemu/auto_qemu:AUDIO_DEV=" " read -a qemu_sound_config <<< "$AUTOQEMU_AUDIO_CONFIG"

embox/scripts $ grep -r -F qemu_usb_config .
./qemu/auto_qemu:USB_DEV=" " read -a qemu_usb_config <<< "$AUTOQEMU_USB_CONFIG"
```

This makes me believe it's safe to delete these lines :)

wawuwo:
As for strange spaces I;'ve found the reason they appear. Later today I will try to add a comment with explanation and provide a fix.

anton-bondarev:
> This makes me believe it's safe to delete these lines :)

Yes, I see now/ Thanks for the explanation :)

wawuwo:
All those whitespace sequences has lines like this as their cause:
```
nic_lines="$nic_lines \
        -netdev $host_nic_config,id=n$ni"
```
Trailing backslash here escape only newline symbol and for Bash this string actually looks like this:
```
nic_lines="$nic_lines         -netdev $host_nic_config,id=n$ni"
```
Then variable `nic_lines` becomes a part of `ARG_LINE` variable, which is passed in turn as an argument to `info` function:
```
ARG_LINE="$sudo $QEMU $RUNAS $AUTOQEMU_MACHINE $AUTOQEMU_LOAD_ARG $AUTOQEMU_KVM_ARG -m $AUTOQEMU_MEM $usb_lines $nic_lines $sound_lines $graphic_lines $@"

info "$ARG_LINE"
$ARG_LINE
```
This is the point when Bash's quirks come into play. The original version of `info` looked like this:
```
 info() {
       echo -e $@ >&2
 }
```
`$@` expands to all function arguments and word splitting **is** performed, so a call `info "one     two     three" "four five"` becomes `echo -e one two three four five >&2`. All that whitespace loses its meaning because of word splitting.

I've changed `info` to look like this:
```
 info() {
       echo -e "$@" >&2
 }
```
`"$@"` expands to all function arguments, but word splitting **is not** performed. The same call `info "one     two     three" "four five"` in this case becomes `echo -e "one     two     three" "four five"`. All whitespace is kept and that's why we see it when `ARG_LINE` is printed.

Here https://onlinegdb.com/G0gzQKDZg is a quick demo of this behavior.

What's the need to quote `$@` then? If it's not quoted, then a call like `info "2 * 2 = 4"` will print names of all files in current directory, surrounded by 2 — this is exactly what `echo 2 * 2 = 4` does. I believe a side effect like this one isn't a desired property of a logging function.  

I've decided that the easiest solution would be to pass `ARG_LINE` to `info` unquoted. As I understand the goal of calling `info` is to see how `ARG_LINE` looks like when it is used as an actual argline. Without quoting `ARG_LINE` would be passed to `info` as a set of arguments, just like when it's used to execute a command and this is exactly what we want to see. (I've added a fixup commit for this)

anton-bondarev (approved):
LGTM

anton-bondarev:
> I've decided that the easiest solution would be to pass `ARG_LINE` to `info` unquoted. As I understand the goal of calling `info` is to see how `ARG_LINE` looks like when it is used as an actual argline. Without quoting `ARG_LINE` would be passed to `info` as a set of arguments, just like when it's used to execute a command and this is exactly what we want to see. (I've added a fixup commit for this)

@wawuwo Thank you very much.

Yes, I'm agree with your decision!

I've approved it. Could you append fixup' commit to the according commit in the PR? And I will merge the PR after that

Could you also prepare next PR from #2847

wawuwo:
@anton-bondarev 
>Could you append fixup' commit to the according commit in the PR? And I will merge the PR after that
I'm not quite sure about the workflow here… Should I rebase with `--autosquash` and then force push the branch?

anton-bondarev:
> Should I rebase with `--autosquash` and then force push the branch?

Yes, you should rebase under upstream/master

```
git fetch upstream
git rebase upstream/master
```
with `--autosquash` or I prefer use interactive mode for manual manipulate with git history
```
git rebase -i upstream/master
```

then make force push your branch

anton-bondarev:
@wawuwo Great! Thanks a lot!
I've merged this PR. Let's next part:)

PR #2865, opened 2023-07-29
templates: x86/test/net mods.conf-vfork depr mods

timtim-ww:
Clean out x86/test/net template of old modules hindering vfork functioning. Because of them Dropbear in multi-process mode didn't function properly


anton-bondarev (approved):
:shipit:

PR #2866, opened 2023-07-29
third-party: DRPBR: pipe stub-write instead of close

timtim-ww:
Currently in Embox in pipeline layer on close() event reading end of pipe isn't notified. Case needs investigation. Dropbear uses close to notify main process of successful authentication. This is the stub to notify by using write instead, until the case is solved.
Additionally building of client and additional applications from the package is removed, since Embox doesn't use them anyway


anton-bondarev (approved):
LGTM

PR #2867, opened 2023-07-29
test: continuous testing-multiple ssh logins test

timtim-ww:
Now that dropbear accepts multiple connections and correctly interprets authentication success this test should pass

#1720


anton-bondarev (approved):
Let's try! :)

PR #2868, opened 2023-07-31
test: pipe() close should notify read-end

timtim-ww:
Although POSIX doesn't seem to pose it directly, we find that third-party programs might be relying on it. As it happens in case of dropbear. Linux man page on select() do mention, that on end-of-file readfs descriptors in select() should be ready to read. So it might be a good idea to stick to it.
This test should pass in case select()'s readfs descriptors are notified on close() event


anton-bondarev (approved):
:shipit:

PR #2872, opened 2023-08-01
test: pipe close() test more correct interpretation

timtim-ww:
More correct result interpretation and a minor naming typo correction


anton-bondarev (approved):
:shipit:

PR #2873, opened 2023-08-01
idesc_pipe: added read-end wakeup on write-end close

### LINKED ISSUE #2869: In pipelines when writing end is closed, reading end should be notified of this event
Although POSIX doesn't seem to have direct mention of this issue, but we find that programs might be relying on it. As it happens in case of dropbear. Linux man page on select() do mention, that on end-of-file readfs descriptors in select() should be ready to read. So it might be a good idea to stick to it.
Here is a small test that first tries to write a test message to a pipeline and then closes. Both events are being watched.
First part passes smoothly, but then on close we have a timeout in select().
#2868

timtim-ww:
Edited pipe status logic, so that now the reading end of the pipe is always ready to read when write-end is closed. The read buffer will be zero-length. (copying Linux behaviour, since, as stated earlier, POSIX doesn't seem to have strict directive for that)
Some additions to the test for that situation

fixed #2869

### DIFF
--- a/src/compat/posix/idx/idesc_pipe.c
+++ b/src/compat/posix/idx/idesc_pipe.c
@@ -55,6 +55,7 @@ struct pipe {
 
 	struct idesc_pipe read_desc;    /**< Reading end of pipe */
 	struct idesc_pipe write_desc;   /**< Writing end of pipe */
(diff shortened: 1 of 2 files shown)


anton-bondarev (approved):
LGTM

PR #2877, opened 2023-08-02
mk/flags.mk: fix gcc version detection

Blarse:
Not all gcc builds match "^gcc" regex, but gcc --version always prints version on the first line, so no need to use grep in the first place


anton-bondarev (approved):
:shipit:

anton-bondarev:
@Blarse Thanks! I've mereged this!

Issue #2835, opened 2023-07-07
Fix shell scripts according to ShellCheck's recommendations 


wawuwo:
Hi! Some of the scripts in `scripts` folder do not pass [ShellCheck](https://github.com/koalaman/shellcheck) analysis. For example:
```
$ shellcheck gdb_load_stm32.sh 

In gdb_load_stm32.sh line 3:
grep 'QSPI)' ./conf/lds.conf > /dev/null && echo "This script doesn't work with QSPI sections!" && exit -1
                                                                                                        ^-- SC2242 (error): Can only exit with status 0-255. Other data should be written to stdout/stderr.


In gdb_load_stm32.sh line 6:
GDB_CMD_FILE=$(mktemp --tmpdir=$PWD embox_gdb_stm32f7_qspi.XXXX)
                               ^--^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
GDB_CMD_FILE=$(mktemp --tmpdir="$PWD" embox_gdb_stm32f7_qspi.XXXX)


In gdb_load_stm32.sh line 8:
EMBOX_BIN=$(mktemp --tmpdir=$PWD embox_XXXX.bin)
^-------^ SC2034 (warning): EMBOX_BIN appears unused. Verify use (or export if used externally).
                            ^--^ SC2086 (info): Double quote to prevent globbing and word splitting.

…
```

I would like to make a contribution and I am willing to fix such issues for all shell scripts from said folder. Hence I wanted to ask if this kind of contribution will be of any value for the project?

anton-bondarev:
@wawuwo Hi

> I would like to make a contribution and I am willing to fix such issues for all shell scripts from said folder. Hence I wanted to ask if this kind of contribution will be of any value for the project?

Yes, sure.

wawuwo:
Sorry for the late reply, I saw your answer and since then I've been fixing the scripts in a forked repository.

I wanted to ask if there is a way to check that I haven't broken something accidentally? If there are typical usage scenarios I could check them myself. (I am striving to be careful but you never know until you check)

anton-bondarev:
> Sorry for the late reply

No problem!

> I wanted to ask if there is a way to check that I haven't broken something accidentally? If there are typical usage scenarios I could check them myself. (I am striving to be careful but you never know until you check)

Unfortunataly for full checking you need apropriate boards. I have a couple of such boards. so let me check them when you create PR. I'll try to do it next week.



[...]

PR #2879, opened 2023-08-04
Refactor autotest/run.sh

wawuwo:
Hi @anton-bondarev! This PR continues the series of PRs related to #2835. It introduces a lot of changes in a single file:

1. Adds comments here and there
2. Rewrites listing of testsuites and testcases. Original approach used global variables to initialize and store lists of testsuite and testcase names which in turn were printed out by special functions. Thus these functions depended heavily on global state. I've tried to reduce this dependence and moved all listing logic into the body of this functions. So this
```diff
-TESTSUITE_LIST=$(ls $TESTSUITE_DIR)
```
has moved here
```diff
 show_available_testsuites() {
        echo "Available testsuites:"
-       echo $TESTSUITE_LIST
+       ls "$TESTSUITE_DIR"
 }
```
And this
```diff
-TESTCASE_DIR=
-TESTCASE_LIST=
…
-TESTCASE_DIR=$BASEDIR/testsuite/$TESTSUITE
-TESTCASE_LIST=$(for i in $(ls -d $TESTCASE_DIR/*/); do basename $i; done)
```
has moved there
```diff
 show_available_testcases() {
-       echo "Availables testcase for testsuite $TESTSUITE:"
-       echo $TESTCASE_LIST
+       local TESTSUITE=$1
+       echo "Availables testcase for testsuite '$TESTSUITE':"
+
+       # Print testcase names on one line separated with spaces.
+       # '-n' omits trailing newline
+       for testcase in "$TESTSUITE_DIR/$TESTSUITE"/*/; do
+           echo -n "$(basename "$testcase") "
+       done
+       echo
 }
```
3. Testcase names now stored as an array instead of a string
```diff
-TESTCASES=$@
+TESTCASES=( "$@" )
```
And its usage is now explicitly states that it is a group of values. No implicit conversions like "split string on spaces" etc.
```diff
-if [ -z "$TESTCASES" ]; then
+if [[ "${#TESTCASES[@]}" -eq 0 ]]; then
…
-for testcase in $TESTCASES; do
+for testcase in "${TESTCASES[@]}"; do
```
4. Other changes are mostly quoting.

I've performed a simple regression test to make sure I haven't broken something: I ran every available testsuite with original script and then with refactored one. Test runs had the same outcome: the test which failed with original version failed with refactored too, which passed with original version passed with refactored too (Embox used in the test runs was started as `./scripts/qemu/qemu_with_disk.sh`)

### DIFF
--- a/scripts/autotest/run.sh
+++ b/scripts/autotest/run.sh
@@ -1,49 +1,115 @@
 #!/bin/bash
 
+# - BASH_SOURCE is an array variable whose members are source filenames.
+# - '%/*' — is a parameter expansion feature which uses everything after
+#   %-sign as a pattern and removes the shortest match to this pattern
+#   from the parameter.
+# This magic spell gives a path to a dir where current source file
+# is located.
 BASEDIR=$(realpath ${BASH_SOURCE%/*})
 TESTSUITE_DIR=$BASEDIR/testsuite
-TESTSUITE_LIST=$(ls $TESTSUITE_DIR)
 
-TESTCASE_DIR=
-TESTCASE_LIST=
 
+print_usage() {
+	echo "Usage: $0 <testsuite> <test case 1> <test case 2> …"
+	echo "  You can 'export TEST_PRINT_ALL=1' to make tests echo to console"
+
+}
+
+
+# Prints available testsuites
+# Usage:
+#    show_available_testsuites
+#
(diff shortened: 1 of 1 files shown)


anton-bondarev (approved):
LGTM

anton-bondarev:
@wawuwo Thanks a lot. I've merged it

PR #2881, opened 2023-08-09
makefile: typo in build system .builded->.built

timtim-ww:
Really small and insignificant, but feels better =)


anton-bondarev (approved):
:)
:shipit:

PR #2882, opened 2023-08-09
Fix shell scripts in scripts/docker according to ShellCheck

wawuwo:
Hi @anton-bondarev!

As usual, most of the changes is wrapping variables containing paths in double quotes and also

1. Checking the exit-code of a command directly
```diff
-       mount | grep -q $PROJECT_LOC
-       if [ $? != 0 ]; then
+       if ! mount | grep -q "$PROJECT_LOC"; then
```

2. Removing obsolete trick (https://www.shellcheck.net/wiki/SC2268)
```diff
-if [ x0 = x$started ]; then
+if [ 0 = $started ]; then
```

3. Removing useless command substitution. There is no need to catch `pwd`'s output and pass it as argument to `echo`. We can use `pwd`'s output as is.
```diff
-       (cd "$1" && echo "$(pwd -P)")
+       (cd "$1" && pwd -P)
```

### DIFF
--- a/scripts/docker/container/bin/gdbwrapper2
+++ b/scripts/docker/container/bin/gdbwrapper2
@@ -1,18 +1,17 @@
 #!/bin/sh
 
-. $(dirname $0)/../gdbwrapper_lib.sh
+. "$(dirname "$0")/../gdbwrapper_lib.sh"
 
 # make alias from host path to container path. After this no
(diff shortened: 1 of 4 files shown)


anton-bondarev (approved):
LGTM

anton-bondarev:
@wawuwo Thanks a lot!

PR #2884, opened 2023-08-09
Fix gdb* scripts according to ShellCheck

wawuwo:
Hi @anton-bondarev! Another portion of the #2835. This PR is about `gdb*.sh` scripts.

Changes are of two types:
1. Wrapping in double quotes the variables containing paths 
2. Replacing patterns like this:
```bash
echo "…" >> some-file
echo "…" >> some-file
echo "…" >> some-file
```
with group commands:
```bash
{
    echo "…"
    echo "…"
    echo "…" 
} >> some-file
```
It gives the same result but looks cleaner (an example of group command's output redirect and pipelining: https://onlinegdb.com/7dQabeEcoF).

These changes don't modify code semantics, so I think there is no need to check the scripts with real hardware.

### DIFF
--- a/scripts/gdb_erase_stm32h7.sh
+++ b/scripts/gdb_erase_stm32h7.sh
@@ -1,24 +1,27 @@
 #!/usr/bin/env bash
 
 GDB=arm-none-eabi-gdb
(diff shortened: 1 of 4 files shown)


anton-bondarev (approved):
LGTM

anton-bondarev:
@wawuwo Thanks. a lot! I've merged it.

PR #2888, opened 2023-08-11
Fix app_diff.sh according to ShellCheck guidelines

wawuwo:
Hi @anton-bondarev! Another PR in the #2835 series.

I believe this PR is the last one and #2835 may be closed after merging it.

This PR:
1. Wraps variables containing paths in double quotes to prevent them from globbing and word splitting.
2. Replaces `! -z` with `-n` in condition check (`-z` — is empty?, `-n` — is not empty?) — new condition does the same, but looks cleaner.
3. Replaces pushd/popd with `cd` in subshell. Workdir is changed in subshell and parent shell's workdir remains untouched.
4. Replaces implicit array-of-strings-to-string conversion with explicit array expansion.

### DIFF
--- a/scripts/app_diff.sh
+++ b/scripts/app_diff.sh
@@ -25,40 +25,40 @@ print_usage() {
 	echo -e "  export PATCHES_APPLY_BEFORE=third_party/fuse/patch1.txt:third_party/fuse/patch2.txt"
 }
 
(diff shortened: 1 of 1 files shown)


anton-bondarev (approved):
LGTM

anton-bondarev:
> I believe this PR is the last one and #2835 may be closed after merging it.

@wawuwo Great job! Thank you very much!

PR #2899, opened 2023-08-18
timer: display work of the system timer

timtim-ww:
Very simple way to see if system timer displays reasonable data


anton-bondarev (approved):
LGTM

PR #2910, opened 2023-08-24
netstat: show tcp sockets grouped by hierarchy

timtim-ww:
-0 option for netstat to show tcp sockets grouped by parent-child hierarchy
some minor output edit so that everything is nicely lined up


anton-bondarev (approved):
LGTM

PR #2923, opened 2023-09-01
net: TCP dynamic alloc of sockets for backlog queue

timtim-ww:
Dynamic allocation of sockets in backlog queue. Still needs further testing, but at the moment seems to work more or less fine.
Minor addition - in tcp.c layer added TCP_PRINT_STATE_INFO define, just in case one would like to see how states are changed in tcp sockets, for debugging purposes, decided to keep it, might be handy in the future.


anton-bondarev (approved):
LGTM

PR #2930, opened 2023-09-04
cmd: service: chk that service hasn't exited since start

timtim-ww:
Service utility didn't take any action when service unexpectedly stopped after the last start.
Necessary checks added in this PR


anton-bondarev (approved):
LGTM

PR #2931, opened 2023-09-04
cmd: service utility check follow up

timtim-ww:
Turned out check if service is running should be a bit more complex
This follow up should fix that


anton-bondarev (approved):
LGTM

PR #2934, opened 2023-09-05
tests: Minor changes to net tests

timtim-ww:
On of the sock_stream tests hung an unclosed socket, close() added
In modbus test client output formatting slightly changed


anton-bondarev (approved):
:shipit:

PR #2935, opened 2023-09-05
net: TCP dynamic backlog alloc critical addition

timtim-ww:
There was no check for having new socket successfully allocated. When the pool is empty leads to crash


anton-bondarev (approved):
:shipit:

PR #2946, opened 2023-09-14
net: tcp FINWAIT2 state timeout added

timtim-ww:
Socket may be stuck in FINWAIT_2 state waiting for fin from the remote host forever, so although it violates specification, seems reasonable to have timeout for it
Anyway it is possible to set timeout option in Mybuild to 0, so that it will wait for fin forever


anton-bondarev (approved):
LGTM

PR #2953, opened 2023-09-26
fix stm32/nucleo_f410rb template

ASASSd:
**platform/stm32/f4/nucleo_f410rb:** `(template)` Fix non-working out-of-box config template for `nucleo_f410rb`


anton-bondarev (approved):
:shipit:

PR #2956, opened 2023-09-30
Httpd rework

krivospitsky:
add use_index_as_404 option to httpd (useful for JS SPA)
httpd_cgi: move CGI_PREFIX to configurable option (/cgi-bin/ path can be change to somethinh like /api/)
move all CGI logic to httpd_cgi. httpd now contains only file logic
unify logic for abstract modules. request processing execute via one method httpd_try_process. cgi use this method to process scriptn in /cgi-bin/ folder and real command.
add documentation with example of using httpd_try_process function in own module.


anton-bondarev (approved):
LGTM

anton-bondarev:
@krivospitsky Thanks a lot. especially for readme!

PR #2957, opened 2023-09-30
get latest version cJSON from github

krivospitsky:


anton-bondarev:
@krivospitsky Thanks a lot.
However I've updated cJSON in different way in #2999

Issue #2963, opened 2023-10-01
Add tests for sin() functions


anton-bondarev:
Look at https://github.com/embox/embox/blob/master/src/compat/libc/math/tests/sqrt_tests.c

AlinaStudyPM:
Hello, I want to work on this issue

anton-bondarev:
Hello @AlinaStudyPM I've assigned it with you

anton-bondarev:
Look at https://github.com/embox/embox/blob/master/src/compat/libc/math/tests/sqrt_tests.c

anton-bondarev:
Closed by #3050

PR #2977, opened 2023-10-03
Added tests for sin()

AlinaStudyPM:


AlinaStudyPM:
#2963

src/compat/libc/math/tests/Mybuild
```diff
@@ -7,3 +7,11 @@ module sqrt_tests {
 module fmod_tests {
 	source "fmod_tests.c"
 }
+
+module log10_test {
```
anton-bondarev, line 11: It seems wrong test name

PR #2978, opened 2023-10-03
Improve stm32/nucleo_f410rb template - add ticker command and heap

ASASSd:
Add ticker command to stm32/nucleo_f410rb template, also add some compat.libc stuff and static heap


anton-bondarev (approved):
:shipit:

PR #2982, opened 2023-10-04
tcp: reset established but not accept()ed connections

timtim-ww:
When listening socket gets closed established connections on accept queue never get closed, instead being released straight away. Commit addresses this issue, resetting these connections


anton-bondarev (approved):
:shipit:

PR #2998, opened 2023-10-07
fix Time.my depends

ASASSd:
Fix time.my depends on `embox.compat.libc.stdlib.system` because time.c uses `system()`


anton-bondarev (approved):
:shipit:

PR #3011, opened 2023-10-11
Fix NULL pointer dereference in fflush()

KirillSmirnov:
fflush(0) is absolutely valid invocation and it means "flush all output streams". Since embox does not track output streams right now, it takes some effort to implement it properly. But at least it should not dereference NULL.

fflush(0) is heavily used in iperf third-party tool.


anton-bondarev:
@KirillSmirnov  you are definitely  right about fflush(NULL).  Thanks a lot.

PR #3024, opened 2023-10-23
tty: add workaround fix in tty_ioctl()

timtim-ww:
workaround for tty device not cause shutdown if it tries to setup several times


anton-bondarev (approved):
LGTM

PR #3036, opened 2023-10-30
net: fix inet_pton string to binary address conversion

timtim-ww:
inet_pton didn't behave correctly, for example it affected pjsip project, when trying to resolve server ip
this commit should fix inet_pton for ip v4 and v6


anton-bondarev (approved):
LGTM

PR #3040, opened 2023-10-31
utils: route CIDR netmask notation support

timtim-ww:
packetdrill in default configuration executes route utility with hostname containing CIDR nemask notation. Previous version of ip string address to binary conversion ignored whatever followed the address. With introduction of more accurate address conversion functions route became unable to correctly parse such a string, which lead for packetdrill and potentially others to fail.
This commit addresses CIDR notation handling in route utility


anton-bondarev (approved):
LGTM

PR #3042, opened 2023-10-31
template: CI test/fs template cifs fix

timtim-ww:
This pull-request should fix failure of the test/fs template in CI when testing cifs file system, that appeared after improving the behavior of inet_*ton functions. The assumption that failure is caused by erroneous colon in mount command for cifs file system in scripts/continuous/run.sh


anton-bondarev (approved):
LGTM

PR #3045, opened 2023-11-01
Fixed mkramdisk cmd error handling

rai044der:
Fixed error handling in mkramdisk command
If system has no available memory for ramdisk, ramdisk_create function will return an error code, which will be used as address of ram disk, which is incorrect.
This patch fixes this problem, now error code is being extracted and checked, before any actions made with ram disk.


anton-bondarev (approved):
LGTM

anton-bondarev:
@rai044der Thanks!

PR #3051, opened 2023-11-06
Arpa inet improve

timtim-ww:
In addition to previous inet_pton edits this commit brings inet_aton in accordance with standard
Additional tests for inet_aton added
ADD: inet_ntop improvements goes here as well


anton-bondarev (approved):
LGTM

PR #3053, opened 2023-11-08
Added message on incorrect input for export cmd

rai044der:
Now if user tries to add a variable without name,
system will print an error message


anton-bondarev (approved):
:shipit:

anton-bondarev:
@rai044der Thanks!

PR #3093, opened 2023-11-28
clock_source_read() fix

### LINKED ISSUE #2902: TSC clock source seems to be working incorrectly
If Embox chooses to use TSC as clock source, it seems to give incorrect data to the module, since kernel system timer loops in a loop of negative(!) 3 to positive 7 nominal seconds, instead of showing somewhat reasonable uptime.
#2899 gives a perspective on that if ran with TSC clock source, otherwise it shows reasonable data
The configuration used is x86/qemu

InfernalKn1ght:
Fix #2902


anton-bondarev (approved):
LGTM

anton-bondarev:
@InfernalKn1ght Thanks/ I've merged it

PR #3104, opened 2023-12-06
Fix build templates/x86/test/units on gcc (GCC) 13.2.1 20230801

DmitryPilyuk:
Modules that cause compilation errors are commented out in mods.conf


anton-bondarev (approved):
:shipit:

PR #3105, opened 2023-12-09
src/lib/crypt/des: fix des_test.c

DmitryPilyuk:


anton-bondarev:
@DmitryPilyuk Thanks. I've merged it.

However it will be better if you separate a PR in two commits (code and template)  in the future

PR #3133, opened 2023-12-29
append options to libiec61850. Modify project/server61850

ivm2008:
I'm not sure if basic_io should be run as a service. Otherwise, it will be impossible to start tls_server


anton-bondarev (approved):
LGTM

PR #3142, opened 2024-01-13
stm32_rng: Fix name

Deryugin:
There is a typo in a file path. Module name is fine, so no template change required


anton-bondarev (approved):
:shipit:

PR #3157, opened 2024-01-26
append platform/templates/bpi_p2_zero. Use u-boot

ivm2008:
generally speaking, this is not so much a new platform as the experience of using u-boot


anton-bondarev (approved):
:shipit:

PR #3175, opened 2024-02-17
Fs improve

anton-bondarev:


src/fs/driver/dfs/dfs.c
```diff
@@ -316,7 +316,7 @@ static int dfs_iterate(struct inode *next, char *name_buf,
 	/* End of directory */
 	return -1;
 }
-
+#if 0
```
Deryugin, line 319: Why do you want to remove this handler?

anton-bondarev: I suddenly found out that ino_pathname() is not used now.
I am rearchitecturing file subsystem so I am trying to simplify diferrent parts of FS. Maybe ino_pathname() will be returned in the future.
By the way what's target of this routine?

PR #3202, opened 2024-03-27
updated des test

Khokholeva:
Added new test case to des_test.c
Added embox.lib.libcrypt.des dependency to des_test module


anton-bondarev (approved):
LGTM

PR #3204, opened 2024-03-28
Tests for crypt module algorithms

Khokholeva:
1) Added des to LibCrypt module
2) Fixed b64.c
3) Added tests for b64 and md5


anton-bondarev (approved):
LGTM

PR #3221, opened 2024-04-16
Adding LCD support for STM32L152C-Discovery

bursian:


bursian:
It is not finished now. LCD support will be in special module in src/drivers

anton-bondarev (approved):
LGTM

Issue #3266, opened 2024-05-18
 Add tests for abs() from math.h 


anton-bondarev:
Look at https://github.com/embox/embox/blob/master/src/compat/libc/math/tests/sqrt_tests.c

kapilkarnani14:
Hi, @anton-bondarev, Can I work on this one ?

anton-bondarev:
Closed #3271

PR #3271, opened 2024-05-18
test: add unit test for abs in math.h

chenguokai:
This solves issue #3266 and enable two math.h tests on aarch64 platform.


anton-bondarev (approved):
LGTM

PR #3369, opened 2024-08-08
Fix: PMP Module (Update PMP Register Management and API Declaration)

SurajSonawane2415:
1. Removed the unnecessary code for manually clearing PMP registers, as it disrupted existing configurations and was found to be redundant.
2. Deleted the pmp_init function declaration from the header file, since it is initialized using EMBOX_UNIT_INIT(pmp_init) and does not need to be exposed in the header.
3. Changed the functions definition of write_pmpcfg and write_pmpaddr from static inline to public to enable its use outside the .c file by including the .h file.

### DIFF
--- a/src/arch/riscv/pmp/pmp.c
+++ b/src/arch/riscv/pmp/pmp.c
@@ -17,7 +17,7 @@
  * @param reg The index of the PMP configuration register to write to.
  * @param value The value to write to the register.
  */
-static inline void write_pmpcfg(uint32_t reg, uint32_t value) {
+void write_pmpcfg(uint32_t reg, uint32_t value) {
     REG32_STORE(PMP_CFG_BASE + reg * 4, value);
 }
 
@@ -27,7 +27,7 @@ static inline void write_pmpcfg(uint32_t reg, uint32_t value) {
  * @param reg The index of the PMP address register to write to.
  * @param value The address value to write to the register.
  */
-static inline void write_pmpaddr(uint32_t reg, uint32_t value) {
+void write_pmpaddr(uint32_t reg, uint32_t value) {
     REG32_STORE(PMP_ADDR_BASE + reg * 4, value);
 }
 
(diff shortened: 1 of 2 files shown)


src/arch/riscv/pmp/pmp.h
```diff
@@ -60,14 +53,14 @@ void set_pmp_entry(unsigned int *index, unsigned int flags, uintptr_t base, size
  * @param reg The index of the PMP configuration register to write to.
  * @param value The value to write to the register.
  */
-static inline void write_pmpcfg(uint32_t reg, uint32_t value);
+void write_pmpcfg(uint32_t reg, uint32_t value);
```
anton-bondarev, line 56: please use 'exten' for declarations

src/arch/riscv/pmp/pmp.c
```diff
@@ -17,7 +17,7 @@
  * @param reg The index of the PMP configuration register to write to.
  * @param value The value to write to the register.
  */
-static inline void write_pmpcfg(uint32_t reg, uint32_t value) {
+void write_pmpcfg(uint32_t reg, uint32_t value) {
```
anton-bondarev, line 20: Do you want to make these routines as global (global view)

anton-bondarev (approved):
LGTM

src/arch/riscv/pmp/pmp.h
SurajSonawane2415: Okay, I will use extern and commit the changes in a few minutes.

src/arch/riscv/pmp/pmp.c
SurajSonawane2415: Yes, I do want to make these routines global. I’ll update the declarations to use extern for the global view and ensure the definitions are appropriately handled. I’ll commit these changes shortly.

SurajSonawane2415:
@anton-bondarev I've made the changes. Could you please review?

anton-bondarev:
> @anton-bondarev I've made the changes. Could you please review?

Thanks. It's OK now.
I've merged the PR

SurajSonawane2415:
Thank you for reviewing and merging this PR.

PR #3376, opened 2024-08-13
mk: Replace the definition method of CXX_CMD macro

sycamoremoon:
macOS 14.5 seems not support the order version definition.


anton-bondarev (approved):
:shipit:

PR #3378, opened 2024-08-17
Prepare Multiprocessor on x86 arch with testing modules

sycamoremoon:
- Currently SMP on x86 arch can set affinity to specify which CPU to execute. Two may running simultaneously and individually or one is in `idle` and another is running thread. The latter may occur when set all threads' affinities to one CPU.
- Each CPU can handle interrupt individually, which can be set in `startup_ap()`. Only clock interrupt is be set in this PR. 

Comments on any commit are welcome!

### DIFF
--- a/src/arch/x86/kernel/irq_handler.c
+++ b/src/arch/x86/kernel/irq_handler.c
@@ -14,11 +14,12 @@
 #include <kernel/irq.h>
 
 fastcall void irq_handler(pt_regs_t *regs) {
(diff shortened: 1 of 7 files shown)


anton-bondarev (approved):
LGTM

anton-bondarev:
@sycamoremoon Thanks
I've approved it.
Could you rebase it under embox/master

sycamoremoon:
No problem

PR #3383, opened 2024-08-17
kernel:lthread: Add static lthread schedee type

sycamoremoon:
if not set type explicitly, the default type is SCHEDEE_THREAD
In that case, it always attemp to context_switch


anton-bondarev (approved):
:shipit:

PR #3384, opened 2024-08-17
driver:interrupt: Change apic to logical dest mode

sycamoremoon:
Previous physical destination mode is not good in SMP system, because physical mode could only delivey inerrupt to one CPU once a time.


anton-bondarev (approved):
LGTM

anton-bondarev:
@sycamoremoon Thanks. I've merged it

PR #3385, opened 2024-08-17
kernel: define size of BSP and AP stack size separately

sycamoremoon:
This PR should be update with latest embox/master to change every place.
If you decide to add this option, please contact me


anton-bondarev (approved):
LGTM

PR #3392, opened 2024-08-22
riscv:Fix PMP_Test suite

SurajSonawane2415:
This PR addresses an issue in the PMP test case where checks involving registers (REG32_LOAD) were causing errors. Since the set_pmp_entry function only computes values for pmp_addr and pmp_cfg without setting the registers, the checks for register values are redundant and unnecessary.

The existing assertions (test_assert_equal(pmp_addr[0], base >> 2); and test_assert_equal(pmp_cfg[0], flags);) are sufficient for validating the function's behavior. This PR removes the redundant checks to prevent errors.

Changes:

    Removed unnecessary REG32_LOAD checks in the PMP test case.
    Re-added test suite into riscv/qemu template 
    
Changes have been tested using the riscv/qemu template to ensure correctness:
Screenshot from 2024-08-20 20-54-22


anton-bondarev (approved):
LGTM

PR #3395, opened 2024-08-23
drivers: Describes irqctrl APIs in more detail

sycamoremoon:
With smp support, the meaning of irq APIs is somewhat unclear. Please discuss further and come up with a more reasonable API.
And do we really need the `irqctrl_eoi_in_cpu` API. I think we can include the self CPUID information when calling the `irqctrl_eoi` function. SInce each CPU only needs to send its own eoi signal. There seems to be no situation where CPU 1 needs to send the eoi signal instead of CPU 2.


anton-bondarev (approved):
:shipit:

anton-bondarev:
@sycamoremoon Thanks a lot!

PR #3397, opened 2024-08-23
riscv_clint: update module to add SMP support

SurajSonawane2415:


src/drivers/clock/riscv_clk/riscv_clk.c
```diff
@@ -38,7 +38,7 @@ static int clock_handler(unsigned int irq_nr, void *dev_id) {
 	(void)a6;
 	asm volatile("ecall");
 #else
-	clint_set_mtimecmp(clint_get_mtime() + COUNT_OFFSET);
+	clint_set_mtimecmp(clint_get_mtime() + COUNT_OFFSET, 0);
```
sycamoremoon, line 41: I recommend that use `cpu_get_id()` instead of  0. Since `cpu_get_id()`  will just return 0. But this will increase compatibility

anton-bondarev (approved):
LGTM

src/drivers/clock/riscv_clk/riscv_clk.c
SurajSonawane2415: @sycamoremoon thanks for suggestion, done with changes.

PR #3399, opened 2024-08-24
kernel:sched: Give each CPU a sched_tick_timer

sycamoremoon:
In SMP case, we use affinity to specify which cpu that a thread belongs to. When we put all threads to one CPU, the rest of it do nothing, IDLE_state cpu will delete the sched_tick_timer. So it's necessary to give each cpu a timer in order to not disturb each other

### DIFF
--- a/src/kernel/thread/thread_switch.c
+++ b/src/kernel/thread/thread_switch.c
@@ -28,7 +28,7 @@ void thread_ack_switched(void) {
 	log_debug("first switch on cpu :%d",cpu_get_id());
 	ipl_enable();
 	/* we add first timer the same behavious as sched_ticker_update*/
-	sched_ticker_add();
+	sched_ticker_add(cpu_get_id());
 	sched_timing_start(&thread_self()->schedee);	
 	sched_unlock();
 }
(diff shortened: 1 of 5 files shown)


src/kernel/thread/thread_switch.c
```diff
@@ -28,7 +28,7 @@ void thread_ack_switched(void) {
 	log_debug("first switch on cpu :%d",cpu_get_id());
 	ipl_enable();
 	/* we add first timer the same behavious as sched_ticker_update*/
-	sched_ticker_add();
+	sched_ticker_add(cpu_get_id());
```
anton-bondarev, line 31: Maybe now we should use 
```
#ifdef SMP
 sched_ticker_add(cpu_get_id());
 #else
 sched_ticker_add(0);
 #endif
 ```

anton-bondarev (approved):
LGTM

src/kernel/thread/thread_switch.c
sycamoremoon: Ok, I will rewrite it totally

PR #3400, opened 2024-08-24
kernel: Add case that runq_get_next returns 0

sycamoremoon:
In some archs such as RISCV, dereferencing NULL will crash. But in some case, x86 will not.
When `runq_get_next()` returns NULL,  which means there is no other threads in `rq` queue. So we can just delete the `sched_timer` for the only thread to run.

### DIFF
--- a/src/kernel/sched/sched.c
+++ b/src/kernel/sched/sched.c
@@ -416,6 +416,14 @@ static void sched_ticker_update(void) {
 	cur = schedee_get_current();
 
 	next = runq_get_next(&rq.queue);
+	/**
+	 * If runq_get_next() returns NULL which means the rq is empty at all.
+	 * There is no need for thread switching, just delete the sched_tick
+	 */
+	if(next == NULL) {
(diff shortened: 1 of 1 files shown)


src/kernel/sched/sched.c
```diff
@@ -416,6 +416,7 @@ static void sched_ticker_update(void) {
 	cur = schedee_get_current();
 
 	next = runq_get_next(&rq.queue);
+	if(next == NULL) goto empty_thread_queue;
```
anton-bondarev, line 419: Could you write this part more carefuly?
I mean use 
```
if (next == NULL ) {

}
```
Execept this, it's great if you add a comment

sycamoremoon: Ok, No Problem

anton-bondarev (approved):
LGTM

anton-bondarev:
@sycamoremoon Thanks a lot!

PR #3403, opened 2024-08-25
arch:riscv: Fix context_switch bug

sycamoremoon:
When thread is created, original context contains a sp which points to next thread. The addtional modification to sp register will cause memory leaking when context switching, which will change the data belongs to next thread.

### DIFF
--- a/src/arch/riscv/kernel/context_switch.S
+++ b/src/arch/riscv/kernel/context_switch.S
@@ -17,7 +17,6 @@
 
 .align 2
 context_switch:
-	addi    sp, sp, -CTX_SIZE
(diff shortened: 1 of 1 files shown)


anton-bondarev:
@sycamoremoon Are you sure that it does not need to allocate stack frame here?

sycamoremoon:
yes, at least for the memory model of pool allocation.  But it can also work in heap allocation. In riscv, context switching has nothing to do with stack

anton-bondarev (approved):
LGTM

anton-bondarev:
@sycamoremoon Thank you very much. It seems you are rigth!

PR #3404, opened 2024-08-25
Create riscv smp tempate and improve smp case

sycamoremoon:
This is the basic version of riscv smp, which allow threads scheduling cross CPUS( two for now).


anton-bondarev (approved):
LGTM

sycamoremoon:
After this updating, riscv CPUS can normally execute threads crossing CPUS, which can be specified by affinities.
Different from x86, each riscv CPUS has a completely independent clock. It's still a question for how to handler every clock interrupt from all CPUs in a good way.  It's rather a critital thing.
I separated clock interrupt handler of APs prepare for further changes. Currently I just let go of all CPUs' clock interrupts, This will cause jiffies to increase faster(depends on how many CPUs the system has). Maybe something bad would happen as more CPUs added, but it seems working in two CPUS for now.

anton-bondarev (approved):
LGTM

anton-bondarev:
> After this updating, riscv CPUS can normally execute threads crossing CPUS, which can be specified by affinities. Different from x86, each riscv CPUS has a completely independent clock. It's still a question for how to handler every clock interrupt from all CPUs in a good way. It's rather a critital thing. I separated clock interrupt handler of APs prepare for further changes. Currently I just let go of all CPUs' clock interrupts, This will cause jiffies to increase faster(depends on how many CPUs the system has). Maybe something bad would happen as more CPUs added, but it seems working in two CPUS for now.

Thanks a lot for the explanation!

sycamoremoon:
#3416 PR seems to mess up the commit history of 
 `src/arch/riscv/kernel/smp/Mybuild` unexpectedly .  In order to reserve commit history, I update this commit to PR. https://github.com/embox/embox/pull/3404/commits/fd4cddaf68d4c23419e30aa0366b91d15880caf5

PR #3410, opened 2024-08-27
Improve system timer strategy for smp

sycamoremoon:
In the previous situation, each CPU would interfere with each other during timer scheduling. Timer sharing was introduced in this PR to solve this problem.

### DIFF
--- a/src/include/kernel/sched.h
+++ b/src/include/kernel/sched.h
@@ -114,6 +114,10 @@ extern void sched_set_current(struct schedee *schedee);
 
 extern void sched_ticker_add(void);
 extern void sched_ticker_del(void);
+extern void* sched_ticker_get_timer(void);
+
+extern void sched_ticker_set_shared(void);
+extern void sched_ticker_set_private(void);
 
 extern int sched_active(struct schedee *s);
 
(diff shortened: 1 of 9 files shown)


src/include/kernel/sched.h
```diff
@@ -114,6 +114,10 @@ extern void sched_set_current(struct schedee *schedee);
 
 extern void sched_ticker_add(void);
 extern void sched_ticker_del(void);
+extern void* get_sched_tick_timer(void);
```
anton-bondarev, line 117: It's better to use sched_ticker_get_time()

sycamoremoon:
I have renamed the `get_sche_tick_timer` to `sched_ticker_get_timer`.

src/include/kernel/time/timer.h
```diff
@@ -53,6 +64,8 @@ struct sys_timer {
 	void *param;
 	unsigned int flags;
 	uint32_t state; /**< do we use timer_set or timer_init_start? */
+
+	struct sys_timer_sharing *timer_sharing;
```
anton-bondarev, line 68: Is this only for SMP?
if yes maybe use SMP macro?

sycamoremoon: I think maybe it's useful for NOSMP case and just a pointer. so I did't put this file into SMP.  But happy to make changes

anton-bondarev: Ok, thanks!

anton-bondarev (approved):
LGTM

PR #3411, opened 2024-08-28
arch:x86: change the smp_send_resched function

sycamoremoon:
The smp_send_resched function is mainly used in sched_tick handler which gives up current thread by force and find a new thread to run. But when one cpu want to give up current thread by itself doesn't means other cpus also want to give up current thread that runs on other cpus

### DIFF
--- a/src/arch/x86/kernel/smp/smp.c
+++ b/src/arch/x86/kernel/smp/smp.c
@@ -116,10 +116,12 @@ void smp_send_resched(int cpu_id) {
 
 void resched(void) {
 	extern void sched_post_switch(void);
+	extern void sched_post_switch_noyield(void);
 
 	lapic_send_eoi();
 
-	sched_post_switch();
+	//sched_post_switch();
+	sched_post_switch_noyield();
 }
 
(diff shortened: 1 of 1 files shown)


src/arch/x86/kernel/smp/smp.c
```diff
@@ -116,10 +116,12 @@ void smp_send_resched(int cpu_id) {
 
 void resched(void) {
 	extern void sched_post_switch(void);
+	extern void sched_post_switch_noyield(void);
 
 	lapic_send_eoi();
 
-	sched_post_switch();
+	//sched_post_switch();
+	sched_post_switch_noyield();
```
anton-bondarev, line 124: Maybe bow use SMP macro?
Or you are sure that it is for all type of systems

sycamoremoon: I think it's no need to do. Because this function is only called in [sched_tick](https://github.com/embox/embox/blob/a6d2f9b713dc40f89c2a5a59e15dacd03fb96f7a/src/kernel/sched/sched_ticker.c#L31), which already used SMP macro .

PR #3412, opened 2024-08-28
sched: Add API of get schedee with no affinity

sycamoremoon:
In smp system, it's useful to know whether is really empty in rq or just because of affinity to get an NULL result


anton-bondarev (approved):
LGTM

anton-bondarev (approved):
LGTM

PR #3414, opened 2024-08-29
improve x86 in smp case one step further

sycamoremoon:
1. Allow more than two cpus
2. Allow each cpu handles all events independently after calling `irq_enable_attached()`


src/arch/x86/kernel/smp/ap_trampoline.S
```diff
@@ -34,7 +33,7 @@ C_ENTRY(__ap_trampoline):
 	.align 0x4
 
 C_LABEL(__ap_sp):
-	.space 4 * (NCPU - 1)
```
anton-bondarev, line 37: If I right understand we will use the only one __ap_sp for all CPU

sycamoremoon:
I updated and kept the `__ap_sp` array for clarity, even though it wastes a small amount of memory.

anton-bondarev (approved):
LGTM

PR #3415, opened 2024-08-29
kernel:irq: Add function of enable attached IRQs

sycamoremoon:
This is one way to enable all registered interrput source in AP. I'm not sure if the method is good, if someone  come up better  way, please give me a suggestion


anton-bondarev (approved):
LGTM

PR #3379, opened 2024-08-17
arch:x86: remove max_supported_cores option

sycamoremoon:
there is no need to consider AMP currently. Additional option maybe make people confusing, NCPU is enough for now.

PR #3416, opened 2024-08-31
kernel:time: Add SMP jiffies tick increasing rule

sycamoremoon:
In some SMP cases, each CPU has a complete independent clock. In order to prevent scheduling frequency changes as the number of CPUs increase.


sycamoremoon:
The previously forgotten `max_supported_cors` options which should be removed in #3379 PR have also been removed in this PR

anton-bondarev (approved):
Thanks.
Let's try!

PR #3418, opened 2024-08-31
riscv:Improve RISC-V FPU module

SurajSonawane2415:
- **Added riscv_fpu_context_init Function:** Initializes the FPU context by clearing all registers. This function ensures that the FPU context is properly set up before usage.
- **Added riscv_fpu_disable Function:** Provides a mechanism to disable the FPU by clearing the FS (Floating-point Status) field in the mstatus register. This is useful for contexts where FPU functionality is not required

### DIFF
--- a/src/arch/riscv/kernel/fpu.c
+++ b/src/arch/riscv/kernel/fpu.c
@@ -1,17 +1,46 @@
 /**
  * @file
- * @brief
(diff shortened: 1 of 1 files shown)


anton-bondarev (approved):
LGTM

SurajSonawane2415:
Thank you @anton-bondarev !

PR #3424, opened 2024-09-11
template: riscv: Uncomment test_pmp in riscv/qemu template

SurajSonawane2415:
- Removed `test_pmp` from the `riscv/qemu` template, which I had previously fixed in a prior PR [PR #3392](https://github.com/embox/embox/pull/3392) but forgot to remove.
- Tested and confirmed that the template working perfectly.


anton-bondarev (approved):
LGTM

PR #3487, opened 2024-12-05
mips: correct mips_cache and add Ralink SoC platform

123serge123:
1. Module *mips_cache* was corrected to use predefined values for data and instruction cache line size.
2. Added ralink diag port driver and new platform **ralink**

### DIFF
--- a/platform/ralink/templates/rt5350/build.conf
+++ b/platform/ralink/templates/rt5350/build.conf
@@ -1,0 +1,17 @@
+
+TARGET = embox
+
+ARCH = mips
+
+PLATFORM = ralink
+
+CROSS_COMPILE = mips-mti-elf-
+// CROSS_COMPILE = mips-elf-
+// CROSS_COMPILE = mipsel-unknown-linux-gnu-
+
+CFLAGS += -O1 -g3
+CFLAGS += -G0 -march=mips32r2 -mabi=32 -EL -DSOC_RT5350 -DRALINK
+
+CFLAGS += -msoft-float
+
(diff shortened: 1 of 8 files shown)


anton-bondarev (approved):
LGTM

anton-bondarev:
@123serge123 Thanks a lot.
I've approved the PR.
The only question refer with src/arch/mips/mm/mips_cache.c.
It is used in baikal_t1 and you've removed `include <drivers/mips/global_control_block.h>`  form this file so baikal_t1 template can be broken.
Could you please extract this commit to a new PR. And we're reviewing it a bit later?

123serge123:
> The only question refer with src/arch/mips/mm/mips_cache.c.
It is used in baikal_t1 and you've removed include <drivers/mips/global_control_block.h> form this file so baikal_t1 template can be broken.

I've analized this file. It was included only for reference to function **mips_cm_l2_line_size**. I've added this ref for *USE_L2_CACHE* is true case too.

```
#if USE_L2_CACHE
extern uint32_t mips_cm_l2_line_size(void);

static uint32_t l2_line_size = CONFIG_SYS_SCACHE_LINE_SIZE;
#endif
```
 baikal-t template was compiled without problem.

> Could you please extract this commit to a new PR. And we're reviewing it a bit later?

It's simpler to remove this PR and make new one only for *mips_cache* module. Without this changes compilattion of ralink platform is stopped with error (do not find header file).

anton-bondarev:
@123serge123 thanks!
It's enough (baikal-t1 compilation) for me 
I've merged!

PR #3537, opened 2025-03-17
Improve tish

AlekseyZhmulin:


src/cmds/shell/tish.c
```diff
@@ -190,17 +198,17 @@ static int process_external(struct cmd_data *cdata) {
 
 	cdata->started = 0;
 
-	pid = new_task(cdata->argv[0], run_cmd, cdata);
+	pid = new_task("", run_cmd, cdata);
```
anton-bondarev, line 201: Now new_task() doesn't use cmd name (the first argument)?

anton-bondarev: I found a new task_set_name routine. Yes, it looks better, but maybe remove the first argument at all?
It's only a suggestion for the future

anton-bondarev (approved):
LGTM

PR #3573, opened 2025-04-29
drivers: clock: make riscv_clk an abstract module

sycamoremoon:
Some vendors get riscv clocksource from CSR TIME, better to make an abstraction.


anton-bondarev (approved):
LGTM

PR #3576, opened 2025-05-01
Tun: fix memory leak and race condition

kamendov-maxim:
fix the memory leak in tun_dev_read by adding skb_free and race condition in tun_xmit by adding lock to the code that accesses queue


anton-bondarev (approved):
LGTM

anton-bondarev:
@kamendov-maxim Thanks!
I've merged it

PR #3578, opened 2025-05-01
Ping: implement -p option and complete TODO

kamendov-maxim:
add -p flag feature implementation - fill packet with specified pattern up to 16 bytes
(there was a switch case statement but this option did not work)

also complete /* TODO: get unique id */ by making id the same as pid of ping


anton-bondarev (approved):
LGTM

anton-bondarev:
@kamendov-maxim thanks a lot 
I've merged the PR

PR #3596, opened 2025-05-09
Rename clean up riscv clint and timer

sycamoremoon:
- Commit 1: 
> riscv: clint: Rename clint APIs to be generic
> 
> Reason below:
> 1. clint also used in S-mode, so it's better rename "mtime" to
> "time", etc.
> 2. some platforms do not have mmio in clint
> 
> And remove redundant comments, clint code is simple and short 
- Commit 2:
> Move different condition cases to riscv_clint, provide currrent
> time should be the reponse of clint module. This would be prettier


anton-bondarev (approved):
LGTM

anton-bondarev:
Thank you! expectantly for the good explanation! I've approved the PR

anton-bondarev:
@sycamoremoon I've merged it. Thanks a lot!

PR #3598, opened 2025-05-09
Add 64bit smp templates

sycamoremoon:
Add templates

### DIFF
--- a/platform/licheepi/templates/smp/README.md
+++ b/platform/licheepi/templates/smp/README.md
@@ -1,0 +1,21 @@
+Refer to platform/licheepi/templates/licheepi_4a/README.md.
+
+In previous single core case, we use `go` command for booting up embox image. But in this smp case,
+we cannot use that unfotunately. The reason is that `go` command only works for the BSP, APs still
+not boot up.
+
+For u-boot normal case, the APs spinning in the `secondary_hart_loop` and  waiting for `bootx`
+cmd, BUT th1520 vender uboot does some $hity modification. we cannot boot APs even use `bootx`
+cmd. And the mainline uboot is under development, so we cannot use that either.
+
+So I did the patch for the vendor uboot to make it work temporarily.
+The vendor code can work and pretty messy, so the patch uses an ugly way.
+
+Here is the steps:
+1. Download vendor u-boot source [1].
+2. Apply the patch using `git am smp-uboot.patch` command under vender u-boot dir.
+3. Build and burn the u-boot following [2]
+4. Use command `tftp 0x20000000 embox;` and `bootelf 0x20000000` to boot up embox. (assume you set up the ip address and have tftp server running on your host)
+
+Link: https://github.com/revyos/thead-u-boot.git [1]
+Link: https://wiki.sipeed.com/hardware/en/lichee/th1520/lpi4a/4_burn_image.html#Linux [2]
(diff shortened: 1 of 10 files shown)


anton-bondarev (commented):
It's very COOL.
However I don't fully understand why does it require the pathc (platform/licheepi/templates/smp/smp-uboot.patch)

sycamoremoon:
I’ll try to explain the situation, Suggestions are welcome

Support for custom (non-Linux) OSes in U-Boot is relatively rare. I spent some time digging into how booting is handled in the vendor-provided U-Boot, and it turned out to be far messier than I expected.

In this U-Boot, only one hart is running; the other harts are either in wfi or spinning. Normally, U-Boot uses IPI to notify the application processors (APs) to jump into the Linux kernel. However, in this case, the vendor U-Boot uses a special, non-standard mechanism to wake up the APs.

1. I tried using the uboot-uimage script to wrap Embox into the uImage format, but it failed. The reason is that the vendor U-Boot doesn’t use the standard Linux boot process. For example, instead of jumping to the kernel at bootm 0x20000, it jumps to a completely different address where OpenSBI is loaded. As a result, it’s impossible to use the original U-Boot to boot my OS without modifying its source code. However, modifying the messy bootm code is even harder than modifying the bootelf command—and doing so would likely break the existing Linux boot process.
2. Since modifying the code is necessary anyway, why not change the go command instead of bootelf? I found a [mail thread](https://lore.kernel.org/u-boot/D9K3I0TAUJ5J.14MKHV437D997@ti.com/) discussing this. In short, the go command is semantically just “go”—it doesn’t imply booting an image. Therefore, it shouldn’t flush the instruction cache (even though some architectures do it, it’s not technically correct). So I decided to modify the bootelf command instead.

If the vendor uboot is standard, it's totally possible to use wrapped embox uImage to boot.
---
At last, this is the affect of go command. APs have gone out of control, of couse the BSP was stucked :(
Screenshot 2025-05-09 at 11 02 53 PM

Use bootelf at patched code, the APs was reseted to the entry address of embox elf.
Screenshot 2025-05-09 at 11 13 26 PM

anton-bondarev (approved):
LGTM

anton-bondarev:
@sycamoremoon Thank you for the explanation! I understand the trouble now.
I don't fully like the way with patch, but I don't know better way. So I suggest to merge the PR as is, and when we guess a better way how we can do it we'll do it

I've approved the PR. Could you please rebase it?

eyestorefivemstore:
Very clean code with good test coverage. This is a great example of best practices.

sycamoremoon:
> @sycamoremoon Thank you for the explanation! I understand the trouble now. I don't fully like the way with patch, but I don't know better way. So I suggest to merge the PR as is, and when we guess a better way how we can do it we'll do it
> 
> I've approved the PR. Could you please rebase it?

Thanks, I don't like the way either. No need to merge right away. The mainline uboot is under developing. maybe we can merge after that
https://lore.kernel.org/all/20250426170059.35571-4-ziyao@disroot.org

eyestoreoye:
This is a well-structured codebase. I appreciate the clean separation of concerns.

sycamoremoon:
I see, I rebased it :-)

anton-bondarev:
@sycamoremoon Thank you very much!
I've merged the PR

Issue #3591, opened 2025-05-09
Add tests for tgamma() from math.h


anton-bondarev:
Look at https://github.com/embox/embox/blob/master/src/compat/libc/math/tests/sqrt_tests.c

PR #3616, opened 2025-05-19
Add tests for tgamma() from math.h

Agonisia:
Hi, this is my first PR, please advise me if there are any deficiencies. 

Completed testing of tgamma() based on x86. 

I've noticed that the openlibm currently used by embox seems to be a bit undefined.

``` 
Errors are reported as specified in math_errhandling.
For IEEE-compatible type double, overflow happens if 0 < x < 1/[DBL_MAX] or if x > 171.7.  
  
If a range error due to overflow occurs, [±HUGE_VAL], ±HUGE_VALF, or ±HUGE_VALL is returned.
POSIX requires that a pole error occurs if the argument is zero, but a domain error occurs when the argument is a negative integer
```
However, errno is not set to ERANGE in the case of an overflow, nor is it set to EDOM when the parameter is a negative integer.

Considering RTOS might trim some standard library features for performance, I commented out these two tests. 

Thank you for your reviewing:)

### DIFF
--- a/src/compat/libc/math/tests/Mybuild
+++ b/src/compat/libc/math/tests/Mybuild
@@ -176,6 +176,11 @@ module frexp_test {
 }
 
 module remainder_test {
-       @Cflags("-fno-builtin")
-        source "remainder_test.c"
+  @Cflags("-fno-builtin")
+  source "remainder_test.c"
+}
+
+module tgamma_test {
+	@Cflags("-fno-builtin")
+	source "tgamma_test.c"
 }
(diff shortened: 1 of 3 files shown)


anton-bondarev (approved):
LGTM

anton-bondarev:
@Agonisia Thanks a lot. Especially your comment about openlibm implementation. Could you create an issue about this?

Also please add (with a separate commit) your test into 'x86/test/unit' template
I've approved the PR. Please add a commit with template modification and prepare to merge (rebase unde embox/master)

anton-bondarev:
This is for #3591

Agonisia:
Have added tgamma_test into the x86 template, rebased based on embox/master. 
I guess that's all? I'm gonna open an issue following your suggestion, thank you. 
Also, I can do the test for ldexp().

anton-bondarev:
@Agonisia thanks a lot. I've merged it

PR #3622, opened 2025-05-21
Add initial riscv S-mode support

sycamoremoon:
Adding basic support for riscv smode with mainline openSBI, using sbi call for timer


src/arch/riscv/kernel/Mybuild
```diff
@@ -2,8 +2,10 @@ package embox.arch.riscv.kernel
 
 module conf {
 	option boolean smode = false
+	depends sbi
```
anton-bondarev, line 5: Maybe we should use direct including for SBI module in mods.conf if it needs?I mean if we have a small MCU it can be not supported or we don't want to insert it in the final image

sycamoremoon: Yes, that's the point I dont like, but the question is the way we select smode: **a option**. I don't know how to tell whether a module is included based on an option, or is there a way to do that?

anton-bondarev: @sycamoremoon I see. Let's try to use abstract module something like
```
@DefaultImpl(sbi_empty)
abstract module conf {
     option boolean smode = false

}

module sbi_full extends  conf {
        option boolean smode = true
	source "sbi.c"
	source "sbi_ecall.c"
}


module sbi_empty extends  conf {
        option boolean smode = false

}
```

sycamoremoon: Good method, I will apply these changes to it, Thanks

anton-bondarev (approved):
LGTM

anton-bondarev:
@sycamoremoon Thanks a lot. I've merged this PR

PR #3639, opened 2025-05-31
updated readme preparing environment command

shauryarane05:
Python 2 has reached end-of-life and is no longer supported. Most modern tools and scripts require Python 3 for compatibility and security. Updating to python3 avoids issues with unsupported dependencies and keeps the development environment up-to-date.


anton-bondarev:
@shauryarane05 Thanks!

PR #3642, opened 2025-06-03
add check qemu-version  for RUNAS flag (-runas (removed in 10.0))

pechenovadasha:
There is problem with running script auto_qemu because the flag -runas for qemu-system was removed in 10.0.
In the script It was added check for version of qemu for RUNAS instruction and if it is lower 10 version than used  flag "-runas" else "-run-with user".


anton-bondarev:
@pechenovadasha Thanks. I've merged it

PR #3651, opened 2025-06-09
docs: add debian dependency python-is-python3 in README

fabiomurer:
When building on debian an error occur because the compilation script wants to use `python` but under debian `python` is not present. 
To fix this i added the `python-is-python3` dependency that creates a symlink to point the `/usr/bin/python` interpreter at the current default `python3`.
with this the compilation is successful.

Another way to fix this is to change the compilation scripts but is more invasive.


anton-bondarev (approved):
:shipit:

PR #3675, opened 2025-07-01
Add flag for making lvgl library 

pechenovadasha:
There was a problem(arm-none-eabi-gcc: error: unrecognized command-line option '-arch'; did you mean '-march='?)  when making the lvgl library on a Mac M1.
This problem was solved by adding the '-D CMAKE_SYSTEM_NAME=Generic' flag to the Makefile.


anton-bondarev (approved):
:shipit:

PR #3755, opened 2025-09-29
Add Device Tree support for RISC-V (BeagleV-Fire)

Shankari02:
This PR introduces Device Tree (DT) support for the RISC-V BeagleV-Fire platform in Embox.
It comes after the ARM PR, since the initial `devicetree_init` logic was added there.
For RISC-V, a separate `devicetree.c` implementation is provided because the ARM version was placed under `arch/arm/arm-m/kernel/`.

### Additions:
New template: `templates/riscv/beaglevfire-devicetree/`
Minimal DTS (`beaglevfire.dts`) describing SoC, UART, and GPIO nodes
Build integration to generate `beaglevfire.dtb` and embed it in the image
`devicetree_init` module for early DT initialization and parsing


anton-bondarev (approved):
LGTM

PR #3760, opened 2025-10-06
Update TUN/TAP driver

kamendov-maxim:
This PR significantly enhances the TUN/TAP driver in Embox by adding support for multiple virtual interfaces, TAP mode, and dynamic configuration via ioctl

The current TUN driver implementation has several critical limitations:
- Only one pre-created static interface supported
- No TAP mode (L2/Ethernet frames) support
- No dynamic interface configuration
- Incompatible with standard Linux TUN/TAP API

This PR adds:
1. Multiple devices Support
   - Introduced tun_devices[MAX_TUNTAP_DEVICES] array for managing up to 16 virtual devices
   - Transformed /dev/tun0 into a cloning device for creating new interfaces
   - Each interface gets its own isolated struct idesc with unique operations

2. TAP Mode Implementation
   - Added full L2 (Ethernet frame) support via IFF_TAP flag

3. Dynamic Configuration via ioctl
   - Implemented TUNSETIFF for setting interface name and mode
   - Implemented TUNGETIFF for retrieving current interface parameters

4. Resource Management
   - Proper cleanup on interface close
   - Queue flushing and route removal
   - Memory deallocation for closed devices

Testing:
Two demonstration applications have been added:
- tun_example - demonstrates TUN mode (L3/IP packets)
- tap_example - demonstrates TAP mode (L2/Ethernet frames)

These applications demonstrate the key difference between TUN and TAP modes by implementing the same functionality (ICMP echo response) at different network layers

How to Test

```bash
# Build with example configuration
make confload-x86/qemu_tuntap_example

# Run Embox in QEMU
sudo ./scripts/qemu/auto_qemu

# In Embox console:
tun_example &
tap_example &

# Test TUN interface
ping 10.0.3.2

# Test TAP interface  
ping 10.0.4.2
```

### DIFF
--- a/src/drivers/net/tuntap/tuntap.c
+++ b/src/drivers/net/tuntap/tuntap.c
@@ -43,7 +43,18 @@ struct tun_dev {
 	bool is_registered;
 };
 
-struct tun_dev *tun_devices[MAX_TUNTAP_DEVICES];
+struct idesc_tuntap {
+	struct idesc idesc;
+	void *priv;
+};
+
+struct idesc_tun_map {
+	struct idesc *idesc;
+	struct tun_dev *tun;
+} tun_devices[MAX_TUNTAP_DEVICES];
+
+static struct mutex tun_devices_mtx;
+
 static const struct char_dev_ops tun_dev_ops;
 
 EMBOX_UNIT_INIT(tun_dev_init);
@@ -71,6 +82,19 @@ static int tun_open(struct net_device *dev) {
 static int tun_set_mac(struct net_device *dev, const void *addr) {
 	memcpy(dev->dev_addr, addr, ETH_ALEN);
 	return ENOERR;
+}
+
+static struct tun_dev *find_tun_by_idesc(struct idesc *idesc) {
+	struct tun_dev *tun = NULL;
+	mutex_lock(&tun_devices_mtx);
+	for (size_t i = 0; i < MAX_TUNTAP_DEVICES; ++i) {
+		if (tun_devices[i].idesc == idesc) {
+			tun = tun_devices[i].tun;
+			break;
+		}
+	}
+	mutex_unlock(&tun_devices_mtx);
+	return tun;
 }
 
 static int tun_xmit(struct net_device *dev, struct sk_buff *skb) {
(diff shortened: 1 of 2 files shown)


anton-bondarev (commented):
Thanks a lot!
One significant issue. You modify 'struct idesc'
i think it's better if you use an inherited structure from 
I mean 
```
struct idesc_tuntab {
struct idesc idesc;
void *priv;
};
```

kamendov-maxim:
I wasn't sure how to fix this using the inherited structure approach: I would need to replace the pointer to struct idesc in the file system with a new one in tun_dev_open(), but I only get a pointer to struct itself, so I can't do anything with the pointer. But I think this approach also solves the problem.

anton-bondarev (approved):
LGTM

anton-bondarev:
> I would need to replace the pointer to struct idesc in the file system with a new one in tun_dev_open(), but I only get a pointer to struct itself as an argument, so I can't do anything with the pointer. But I think this approach also solves the problem.

Yes, I think it's a better way
I mean you can do someting like idesc_serial 
[https://github.com/embox/embox/blob/master/src/drivers/tty/serial/idesc_serial.c](https://github.com/embox/embox/blob/master/src/drivers/tty/serial/idesc_serial.c)

anton-bondarev (changes requested):
please use idesc_tuntab instead of idesc modifications

kamendov-maxim:
I'm sorry, but it seems I don't quite understand your solution. To use struct idesc_tuntap I need to allocate it in tun_dev_open(), copy the old idesc into it, and replace the old struct idesc pointer with the new one. The problem here is that I only have a pointer to the structure itself, so I can't change the pointer in file system, so I will always get a pointer to the old idesc and will not be able to get the void *priv field
If I just cast idesc to struct tuntap_idesc*, I will use the memory immediately after it for priv that was not allocated for it, which is unsafe
My current solution with searching up tun device by idesc also doesn't need to change the idesc structure, I removed void *priv field that I added earlier from struct idesc
I understand why your approach might be better, but currently I don't really understand how to do it

PR #3829, opened 2025-12-17
Platform/esp32c3: Add an interrupt controller

Urtix:
Added interrupt controller and timer stub.
Matrixint renamed to esp32c3_intc.


anton-bondarev (approved):
LGTM

PR #3961, opened 2026-04-17
Add a bmp280 sensor

Urtix:
1) Reading raw data from registers responsible for temperature has been implemented.
2) The bmp280 command outputs the raw temperature value.


anton-bondarev (approved):
LGTM

PR #3962, opened 2026-04-17
Ncnn: cmake and MobileNetV3 example

krivenkodm:
## Summary

This PR refactors the NCNN integration for Embox and validates the new layout with two working model demos:

- YOLOv8n object detection
- MobileNetV3-Small image classification

The main goal was to replace the previous ad hoc integration with a cleaner CMake-based flow, reduce reliance on generated checked-in artifacts, and make the project structure ready for multiple models.

### 1. Reworked NCNN integration in `third-party/lib/ncnn`
- switched NCNN runtime integration to a CMake-based build
- added an Embox-specific toolchain file
- replaced manual header overrides with a cleaner integration scheme:
  - `platform.h` is now generated through a patched upstream CMake flow
  - `features.h` and `cwchar` are provided through Embox overlay tree
- cleaned up obsolete compatibility files and old manual copies
- kept only the compatibility pieces that are still required for Embox:
  - `ncnn_new_delete.cpp`
  - `ncnn_math_trig_shim.cpp`

### 2. Split project code into reusable modules
Introduced a clearer structure under `project/ncnn`:

- `common/image`
- `common/postprocess`
- `common/types`
- `common/labels`
- `common/output`
- `models/yolov8n`
- `models/mobilenetv3_small`

This makes the codebase easier to extend with additional models without mixing runtime, model logic, and demo-specific code.

### 3. Refactored YOLOv8n into a standalone model module
- moved YOLOv8n-specific logic out of the demo command
- kept detection postprocessing reusable
- simplified the demo command so it acts as a thin wrapper around the model API

### 4. Added a real MobileNetV3-Small classifier demo
- added a second model type to validate multi-model support
- integrated a working MobileNetV3-Small classifier
- uses embedded `id.h + mem.h` assets, similar to the binary-memory loading approach
- added ImageNet labels for readable top-k output

Example output now looks like:

- `cls=263 prob=0.916 name=Pembroke`
- `cls=264 prob=0.053 name=Cardigan`

## Why

Before this change:
- NCNN integration relied on more manual patching/copying
- model-specific code was too tightly coupled to demos
- the layout was not well suited for supporting more than one model

After this change:
- NCNN integration is cleaner and more maintainable
- the project is structured around reusable common code and isolated model wrappers
- the architecture is now validated on both detection and classification workloads

## Testing

Tested on Embox ARM QEMU template.

Verified:
- successful NCNN runtime build through CMake
- successful build of YOLOv8n demo
- successful build of MobileNetV3-Small classifier demo
- successful runtime inference for classifier demo on a dog image
- top-k classifier output is reasonable and human-readable via labels

## Notes

- classifier integration uses binary model loading (`id.h + mem.h`) to avoid the fragile text-param path
- this PR intentionally keeps the model demos thin and moves reusable logic into shared modules
- further models can now be added with much less integration work


anton-bondarev (approved):
LGTM

PR #3995, opened 2026-06-09
Fix broken download URLs and update build patch for Mesa3D

SecretPersona5:
Mesa and GLU archives moved to new locations, old URLs returned 404/FTP timeout. Updated both to archive.mesa3d.org.

The existing patch for mesa-18.2.5 referenced asm-generic/types32.h which no longer exists in Embox - fixed in three places (GL/gl.h, drm-uapi/drm.h, compiler.h). Also added __GNUC_PREREQ definition to Mesa's compiler.h under __EMBOX__ guard, since Mesa's include path shadowed Embox's own compiler.h.

Tested on arm/qemu (arm926ej-s, integratorcp), osdemo_fb


anton-bondarev (approved):
LGTM

PR #4024, opened 2026-06-22
Improve usb_dev on stm32f7

SecretPersona5:
Fixes the USB CDC-ACM gadget on NUCLEO-F429ZI/STM32F769I. It enumerated but failed to configure.

Problem
config index 0 descriptor too short (expected 67, got 9)
config 1 has 0 interfaces, different from the descriptor's value: 2
can't set config #1, error -71

Plus garbled manufacturer/product strings. The device showed up but Linux could never configure it.

 Root causes and fixes
- EP0 was stalling right after a successful multi-packet transfer, that is why the config descriptor came back truncated. Fixed in stm32f7_udc.c.
- Bulk endpoints were hardcoded to wMaxPacketSize=64.  Bumped to 512.
- acm_tty_init() queued the first RX request before the endpoints were enabled. Reordered.
- TX/RX were byte at a time with no real buffering, output got dropped under load. Rewrote around ring buffers.
- Added a second independent shell so the USB console does not replace the existing UART console, both run side by side now.

Testing
Verified on real hardware: clean enumeration, full config descriptor reads correctly, /dev/ttyACM* shows up, shell stays interactive through a full boot banner, UART and USB consoles work at the same time.


anton-bondarev (approved):
LGTM

PR #4028, opened 2026-06-24
Improve stm32f7: usb_dev: CDC-ECM gadget

SecretPersona5:
Fixed cdc_ECM functionality. Packets are now sent successfully.
* Initialization order corrected
* wMaxPacketSize 64 -> 512 for bulk-endpoint
* Fixed TX synchronization. Added completion-callback + busy-flag
* Synchronized MAC addresses
* Fixed a bug with ECM_RX_BUF_SZ = 1024; the Ethernet frame reassembly buffer from 512-byte USB bulk chunks was too small for a full-size frame with an MTU of 1500.
* Removed annoying printk for each USB IRQ

Reproduced and tested on the stm32f769i-discovery


anton-bondarev (approved):
LGTM

PR #4061, opened 2026-08-19
src/drivers/video: QEMU ramfb, bcm2711_fb; RPi4 template

SBKarr:
aarch64 display path for the QEMU virt machine and real Pi 4.

### drivers/video: ramfb driver for the QEMU virt machine

`-device ramfb` as a pseudo video device: the guest allocates a buffer and
writes its address, format and geometry into the fw_cfg file `etc/ramfb`,
and the host scans out of that memory from then on. The driver looks the
file up in the fw_cfg directory over the DMA interface, points it at a
static buffer and publishes it through `fb_create()` as `/dev/fb0`.

### drivers/video: VideoCore mailbox framebuffer for BCM2711 (Pi 4)

On the Pi the framebuffer is owned by the VideoCore firmware and is set up
over a property-channel mailbox request. `embox.driver.video.raspi_video`
already does this for the older SoCs, but it goes through the
`bcm2835_mailbox` driver, sends the tags one at a time and defaults to a
BCM2835 mailbox base — none of which works on a Pi 4. Rather than reworking
raspi_video and risking the boards it currently serves, this is a separate
driver.

It sends one bulk request (SET PHYS/VIRT SIZE, VIRT OFFSET, DEPTH, PIXEL
ORDER, ALLOCATE, GET PITCH) — the sequence the Pi 4 firmware expects —
converts the returned VideoCore bus address to an ARM physical address,
maps it uncached into the kernel context and publishes it as `/dev/fb0` in
BGRA8888. Display size comes from the firmware (EDID), with a configurable
fallback so a board booted with no monitor attached still gets a buffer.
The mailbox base is an option, so the same code covers the earlier SoCs'
base addresses.

The serial console is untouched: HDMI is scanout only.

### templates/aarch64: Raspberry Pi 4B template

There is currently no Pi 4 template in the tree. This one boots from the SD
card as `kernel8.img` with the console on the 40-pin header:

  * GIC-400 (GICv2) at `0xff841000` / `0xff842000` — low peripheral mode,;
  * PL011 UART0 at `0xFE201000`, 48 MHz core clock, IRQ 153 (SPI 121) —
    `dtoverlay=disable-bt` muxes it onto GPIO 14/15;
  * generic armv8 physical timer on IRQ 30, `core_freq` 1.5 GHz;
  * 4 KiB translation granule, matching the page size;
  * HDMI through `embox.driver.video.bcm2711_fb`, console stays on UART;
  * loopback only — the on-board GENET NIC has no driver yet.

### DIFF
--- a/src/drivers/video/bcm2711_fb/Mybuild
+++ b/src/drivers/video/bcm2711_fb/Mybuild
@@ -1,0 +1,24 @@
+package embox.driver.video
+
+module bcm2711_fb {
+	/* VideoCore mailbox. BCM2711 (Pi 4) is at 0xFE00B880, BCM2837 (Pi 3)
+	 * at 0x3F00B880, BCM2835 (Pi 1) at 0x2000B880. */
+	option number base_addr = 0xFE00B880
+
+	/* Used when the firmware reports no display size (HDMI unplugged,
+	 * no EDID) - the mailbox is still able to allocate a buffer. */
+	option number fb_xres = 1920
+	option number fb_yres = 1080
+
+	option number fb_bpp = 32
+
+	source "bcm2711_fb.c"
+
+	depends embox.driver.periph_memory
+	@NoRuntime depends fb
+	depends fb_devfs
+	@NoRuntime depends embox.compat.libc.all
+	depends embox.mem.mmap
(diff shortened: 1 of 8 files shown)


anton-bondarev (approved):
LGTM

PR #4070, opened 2026-09-09
Fixed and improved FAT and oldfs for SMP concurrency support

SBKarr:
## Motivation

Implementing SMP in Embox requires a stable base of operations that won't break when running in multi-core mode. This is the first of three stages in our plan (others: kernel serialization and actual SMP on aarch64). Its goal is to stabilize file system functions sufficiently so that SMP tests actually test SMP, rather than issues with other modules.

Another goal of this patch set is to enable OTA update support in the future, which requires guarantees that fs functions won't corrupt the file system on their own when writing large files from a known-unstable network.

At the same time, the FAT module has been enhanced so that no write will break fsck, even with relatively innocuous warnings like an extra empty sector in a chain.

Only FAT and oldfs are affected.

## Overview

This collection of 18 patches brings upstream embox's filesystem and VFS layer
to a state suitable for SMP operation. The changes address three categories of
problems:

1. **Concurrency safety** — locking for shared data structures (mount table,
   inode tree, FAT driver's global scratch buffer, test emitter)
2. **Correctness bugs** — off-by-one errors, wrong cluster allocation timing,
   missing validation, use-after-free on unmount
3. **Inode lifetime management** — reference counting and generation numbers
   to prevent stale descriptor use after unmount/remount

## Patch Table

| # | Commit | Subject | Brief |
|---|--------|---------|-------|
| 1 | 6651aac2e9 | util: atomic RMW with fallback for non-lock-free targets | Provides atomic read-modify-write operations that degrade to barrier-protected plain operations on targets without lock-free support (e.g., 80386) |
| 2 | a086fe1931 | fat: serialise driver with global lock | Adds a recursive mutex around all FAT driver operations to protect the shared `fat_sector_buff` global scratch buffer |
| 3 | 8ba2fa4fd5 | fat: size FAT table for volume when formatting | Computes correct FAT table size during formatting instead of hardcoded 1 sector, preventing writes past table boundary |
| 4 | b7d1ba3cbd | fat: optimize free cluster search with hint | Adds per-volume free-cluster search hint to avoid rescanning already-allocated clusters; also invalidates FAT32 FSInfo free count |
| 5 | b63630b683 | fat: bound long name offset from directory entry | Validates long filename sequence numbers before using them as buffer offsets, preventing stack overflow from corrupted entries |
| 6 | 7869ccb506 | fat: handle FAT12/16 root directory as fixed run of entries | Walks FAT12/16 root directory linearly and stops at its end, instead of following cluster chain past root |
| 7 | 20a5d03b19 | fat: fix reserved FAT entries and root directory | Clears root directory on format (no "ROOT DIR" entry) and writes correct reserved FAT entries 0 and 1 |
| 8 | 2fc667f1d7 | fat: fix cluster allocation timing in write | Allocates new cluster when starting a new cluster boundary, not after ending one; prevents extra unused trailing cluster |
| 9 | 05ef3960c6 | fat: validate volume geometry before use | Validates bytepersec/secperclus before dividing by them; prevents division-by-zero and buffer overflows |
| 10 | d6ee0ace9f | fat: fix cluster chain walk and directory entry handling | Adds `>= 2` guard to cluster chain walks; uses firstcluster (not cursor) for directory entry matching; fixes long name cleanup |
| 11 | b819198a34 | fat: clear new directory cluster before writing entries | Zeroes entire new directory cluster before writing . and .. entries, preventing garbage entries from leftover data |
| 12 | 05ac471adf | oldfs: inode reference counting with generation numbers | Adds i_ref/i_dying reference counting and i_gen generation numbers to inodes; prevents use-after-free and stale descriptor use |
| 13 | 71466694b1 | oldfs: kmount undoes what it built when it fails | Frees superblock and unwalks created inodes on kmount failure; prevents resource leaks and orphaned subtrees |
| 14 | d0ba5cbdfd | oldfs: kumount refuses volume with open files | Validates mount point, seals superblock to prevent new opens, refuses unmount with open descriptors |
| 15 | f90c16330e | oldfs: mount table takes a lock | Adds spinlock to mount table operations; serialises path lookups against mount/unmount |
| 16 | e53e159297 | oldfs: kread clamps against one reading of the file size | Reads i_size and pos once, clamps read size against that; prevents TOCTOU race where clamp raises size |
| 17 | ab045e1b7e | framework: test emitter takes a lock | Replaces interrupt masking with spinlock for test output buffer; safe on SMP |
| 18 | bad3d4dc62 | tests: fs FAT operations, big files, inode lifetime, mount SMP | Adds comprehensive FS test suites: fat_ops, fat_big, inode_life, mount_smp |


KevinACoder:
very good, i also see some issue in oldfs when trying to mount blkdev

anton-bondarev (approved):
LGTM

anton-bondarev:
> very good, i also see some issue in oldfs when trying to mount blkdev

It's better to use dvfs instead of old fs. We are switching and improve dvfs now

```
/* for old fs comment dvfs part */
	@Runlevel(2) include embox.fs.node(fnode_quantity=1024)
	@Runlevel(2) include embox.fs.rootfs_oldfs
	@Runlevel(2) include embox.fs.driver.initfs
	@Runlevel(2) include embox.fs.driver.ramfs
	@Runlevel(2) include embox.fs.driver.ext2
	@Runlevel(2) include embox.fs.driver.fat
	include embox.fs.driver.devfs_old
	include embox.compat.posix.file_system_oldfs

/* for dvfs comment old fs part */
/*
	@Runlevel(2) include embox.fs.dvfs.core
	@Runlevel(2) include embox.fs.driver.fat_dvfs
	@Runlevel(2) include embox.fs.driver.initfs_dvfs
	@Runlevel(2) include embox.fs.rootfs_dvfs
	include embox.compat.posix.file_system_dvfs
	include embox.fs.driver.devfs_dvfs
*/
```

anton-bondarev:
@SBKarr Cool. Thank you very much

Could you also do the same for dvfs? We try to use and develop it

SBKarr:
> @SBKarr Cool. Thank you very much
> 
> Could you also do the same for dvfs? We try to use and develop it

Y, I'll take a look

PR #4072, opened 2026-09-10
ESP32C3 improve

DmitryPilyuk:
Fixed an interrupt controller bug that could cause the system to hang. Also fixed a related UART issue.


anton-bondarev (approved):
:shipit:

PR #4075, opened 2026-09-14
kernel+mem+riscv: SMP support: serialization and locks

SBKarr:
## Motivation

This is the second set of patches (of three) for SMP support on Embox. The main idea is to implement kernel operation in concurrent multithreading. This required, first of all, changing the locking protocol to support preemption and scheduling as separate lock classes (0002). Then, classifying existing locks into these classes. This is not an optimization, but a necessity in conditions where two real cores compete for a single resource.

Problems were identified through lengthy test runs (approximately two hours per single test) in a concurrent environment. We likely haven't found all the serialization issues, but this will become clearer over time. The tests are included in the patch set.

The existing risc-v tests failed, so the patches for risc-v are also included, but we didn't have a physical device to test them.

Also included is a patch (0001), which fixes a compatibility issue with libstdc++ and C11 after out previous pull request.

## Overview

A set of 26 patches. They make the kernel correct on more than one core:
real spinlocks, a scheduler whose run queue is read under its own lock,
per-CPU timer lists, locking for the memory allocators, and the riscv SMP
pieces the two-hart suites need.

- **Spinlocks and critical sections** — a real acquire/release instead of a
  busy-wait, a BKL that owns its own word, and per-CPU data on its own cache
  line.
- **Scheduler** — the run queue is read under `rq.lock`, a schedee leaves the
  queue at the level it was filed under, nothing touches a schedee after its
  last context switch, and wakeups go only to cores that can run the schedee.
- **riscv SMP** — one `mtimecmp` per hart (as CLINT has) and the boot hart
  unmasking its own software interrupt.
- **Memory** — locking for the pool and page allocators, a segment unlinked
  before it is freed, and a double-free that names its caller.
- **Tests** — SMP suites for sleep, join, the mutex shape, and the run queue,
  plus the heap under two tasks.

## Patches

Hashes are from the reference `kernel` branch, base `76f820b899`.

| # | Hash | Description |
|---|------|-------------|
| 0001 | `cb598516ff` | `util`: `atomic_rmw_`-prefix the RMW macros |
| 0002 | `9e2e5b897b` | `kernel/critical`: a spin region wants preemption off, not the BKL |
| 0003 | `91f04e950f` | `kernel/critical`: a per-CPU read has to name the CPU that is asking |
| 0004 | `c3a6b31537` | `kernel/spinlock`: a real acquire and a real release |
| 0005 | `59390ec116` | `kernel/cpu`: give each CPU block of per-CPU data a cache line of its own |
| 0006 | `afdd14f993` | `kernel/sched`: the run queue is read under its own lock |
| 0007 | `3f1707c02e` | `kernel/sched`: a schedee leaves the run queue at the level it was filed under |
| 0008 | `28b92adad7` | `kernel/sched`: nothing may touch a schedee after its last context switch |
| 0009 | `cd80be1310` | `kernel/sched`: wake the cores that can actually run a newly ready schedee |
| 0010 | `95ab526caa` | `kernel/sched`: the tick does not poke a CPU that has not been started |
| 0011 | `3e96431806` | `kernel/sched`: `__schedee_priority_set()` must not take `sched_lock()` under `rq.lock` |
| 0012 | `2950598d7b` | `kernel/sched`: the affinity accessors are one-line reads |
| 0013 | `663904ed2a` | `kernel/thread`: a thread pointer carries a magic, and a thread is deleted once |
| 0014 | `946c6714bf` | `kernel/task`: a task exits once, in order, and the parent reaps it last |
| 0015 | `4af30fb67b` | `kernel/task/idesc`: the descriptor table takes a lock |
| 0016 | `aa86938486` | `kernel/time`: one timer list per CPU, each behind its own lock |
| 0017 | `031970b8b2` | `kernel/time`: count the timer interrupts each core has seen |
| 0018 | `d572a8373a` | `drivers/clock/riscv_mtimer`: one `mtimecmp` per hart, as CLINT has |
| 0019 | `be3bc8a6d3` | `arch/riscv/smp`: the boot hart unmasks its own software interrupt |
| 0020 | `2c0f9b0c39` | `mem/page`: a request larger than the pool is refused, not answered wrongly |
| 0021 | `39b595a511` | `mem/pool`: `pool_alloc()` and `pool_free()` take a lock |
| 0022 | `5187a6690d` | `mem/heap`: unlink a segment before freeing it, and name who freed twice |
| 0023 | `9c45da7056` | `drivers/tty`: the tty rings have a thread side and an interrupt side |
| 0024 | `9d3cb9f124` | `drivers/mmc`: report I/O errors, wait for the card, and address ACMD6 correctly |
| 0025 | `b270d5ce1c` | `tests/kernel`: sleep, join, the mutex_test shape, and the run queue |
| 0026 | `9400c808b5` | `tests/mem`: the heap under two tasks |


anton-bondarev (approved):
CI

anton-bondarev:
@SBKarr Thank you very much!

PR #4077, opened 2026-09-16
libc/math: Add tests for fma()

cnichifor:
Add tests for fma() covering normal and special floating-point cases.

Tested with the x86/test/units configuration under QEMU.

Related to #3923


anton-bondarev (approved):
LGTM

anton-bondarev:
@cnichifor Thanks. I've merged it

PR #4078, opened 2026-09-18
aarch64: SMP, bringing a secondary cores up

SBKarr:
## Motivation

This is the last part of the three steps for running SMP on aarch64. Its purpose is to actually launch secondary cores and provide the ability to verify (via CI) that this will actually work in the future. The preceding work (on the file system and kernel) was necessary so that this part could be tested on real-world workloads.

This work has been tested on an RPi4 (and Radxa 5C; a template and drivers will be available later) using multithreaded software rendering and the graphic user application workload.

Not resolved here: the multithreaded performance issue may be limited by BKL synchronization, especially when using TLS emulation. This can be addressed at the application level by using a non-blocking TLS access scheme, using TPIDR_EL1 as the key. A possible OS-level solution might be futex-like synchronization and/or separation of protected data.

## Overview

Nine patches bring up the secondary cores of an aarch64 machine;

- **Interrupt controller** — a GIC has state that belongs to the machine and
  state that belongs to one CPU, and only the first was set up; plus the SGI
  one core raises on another.
- **Bringing a core up** — logical ids in `TPIDR_EL1`, `PSCI_CPU_ON` or a spin
  table to start a core, and the reset handler's secondary entry points.
- **Faults** — a permission fault has a level, `SPSR_ELn_M_MASK` was a typo,
  the SError syndrome is printed, and an abort on one core stops the others
  before they print over the report.
- **Translation and tables** — the MMU is on before the first spinlock, and a
  partial table walk does not leave the previous walk's pointers behind.
- **Tests** — `smp_threads`, `templates/aarch64/smp`, and the SMP suites added
  to `x86/smp`.

Patch 0009 is the test suite plus everything that runs it. `templates/aarch64/smp`
builds `src/arch/aarch64/kernel/smp/` for CI suite.

## Patches

| # | Hash | Description |
|---|------|-------------|
| 0001 | `503412c61a` | `drivers/interrupt/gic`: `irqctrl_init_cpu()` and `irqctrl_send_ipi()`, a redistributor frame per CPU, bounded register waits |
| 0002 | `98f590a322` | `aarch64`: the logical CPU id in `TPIDR_EL1`, the MPIDR behind it, and the PSCI calls that start a core |
| 0003 | `214396fd1f` | `aarch64`: a secondary is started, comes up on its own stack with its own tick, joins the scheduler — and is stopped again |
| 0004 | `e31e6832ff` | `aarch64`: all four permission fault codes, with the level; `SPSR_ELn_M_MASK` is binary |
| 0005 | `27de41a439` | `aarch64`: an abort or a failed assertion stops the other cores before printing, and reports where each was |
| 0006 | `a7ef1ce049` | `aarch64`: `ESR_EL1`, decoded, and `FAR_EL1` on an SError |
| 0007 | `b6fca58f1a` | `aarch64`: `arch_mmu_early_on()` maps 1 GiB blocks and turns translation on from `kernel_start()` |
| 0008 | `e6c1921f8f` | `mem/vmem`: a walk that stops clears the levels it did not reach, and its callers test them |
| 0009 | `7157fb4d6b` | `tests/smp`: the `smp_threads` suite, `templates/aarch64/smp`, and the SMP suites on `x86/smp` |


anton-bondarev (approved):
LGTM

anton-bondarev:
@SBKarr Thank you very much.

SBKarr:
Another known issue: when a core holding a BKL dies, the BKL remains locked and other cores may not receive an interrupt to stop gracefully if they were also waiting on the BKL. Since this is no longer a valid system state, it's not a big problem, but it is impossible to read the state of the hung cores.

